PR #2, opened 2015-03-11
Fixed prompt text no longer being copied to the prompts struct

willco007:
Regression in 1.5 from 1.4.4.


bagder:
Argh, fixed in commit fe3e23022b. Regression was introduced by me in 031566f9c.

Thanks!

PR #3, opened 2015-03-11
KNOWNHOST_CHECK_ define values

willco007:
Change LIBSSH2_KNOWNHOST_CHECK_* error values so their values don't collide with LIBSSH2_FX_* defined codes to make error matching easier.


willco007:
Ugh, sorry about the pull request mess. I really only wanted the known_hosts defines file.

bagder:
Ok, I sympathize with what you're trying to do here but this change breaks the ABI. Applications built against 1.5.0 will not work with a version compiled with this...

willco007:
Yeah, OK. Perhaps revisit with the next major release?

bagder:
Fair enough, added milestone for that to make it findable.

bagder:
Closed for now.

PR #4, opened 2015-03-11
mingw build: allow to pass custom CFLAGS

vszakats:
Allow to pass custom `CFLAGS` options via environment variable `LIBSSH2_CFLAG_EXTRAS`. Default and automatically added options of `GNUmakefile` have preference over custom ones. This addition is useful for passing f.e. custom CPU tuning or LTO optimization (`-flto -ffat-lto-objects`) options. The only current way to do this is to edit `GNUmakefile`. This patch makes it unnecessary.

This is a mirror of similar libcurl patch: https://github.com/bagder/curl/pull/136

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -110,7 +110,7 @@ endif
 -include $(OBJDIR)/version.inc
 
 # Global flags for all compilers
-CFLAGS	= $(OPT) -D$(DB) -DLIBSSH2_WIN32 # -DHAVE_CONFIG_H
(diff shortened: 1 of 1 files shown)


bagder:
Fair enough, simple and small. Merged as commit 8f00a7471d.

BTW, please wrap lines in the description at column 72 to make them nicer to read in ordinary "git log" displays etc. I made it like this before I pushed it.

Thanks!

vszakats:
Thanks! I'll make sure to add comments in the commit message and wrap it next time.

PR #5, opened 2015-03-11
Allow authentication keys to be passed in memory.

calavera:
This patch allows to pass ssh keys in memory, without reading files or getting them from an agent. It only works with OpenSSL.

As discussed in the mailing list a few months ago, it would be better to have additional support for libgcrypt before merging this PR, but I don't have the experience necessary to write a good implementation, so I'll leave it as exercise for the reader.

Shout out to Joe Turpin who implemented the patch originally, I just re-applied and cleaned it:
http://www.libssh2.org/mail/libssh2-devel-archive-2012-01/0015.shtml

* Adds `libssh2_userauth_publickey_frommemory` macro for `libssh2_userauth_publickey_frommemory_ex`.
* Use an unimplemented error for extracting keys from memory with libgcrypt.


bagder:
Please make a clear note in the docs about it only working for the OpenSSL backed version.

calavera:
Thanks for pointing that out @bagder. I've added a notice in both docs to make it clear.

alamaison:
The patch looks good to me.  But can you remove the #defined 'alternative' signature macro (and the _ex suffix on the real signature)?  I know the existing API has many of these but we've stopped doing that for new functions.

calavera:
Thanks for letting me know @alamaison. I've removed the macro and renamed the function already. I have a doubt though, should I remove the size parameters and check those values inside the function or I should leave it as it is now?

That way the function signature would still be something like:

```
libssh2_userauth_publickey_frommemory(session, username, publickey, privatekey, passphrase)
```

alamaison:
Thanks @calavera. Please leave the size parameters. The library shouldn't be calculating sizes on the callers behalf. The callers strings might not even be null terminated

bagder:
I agree with @alamaison on this and also that's how some of the other functions already work.

calavera:
cool, the lengths are still in :smile:

alamaison:
One last issue: the type of `user_len` parameter should be `size_t`, not `unsigned int`.

calavera:
Fixed, thanks again @alamaison!

alamaison:
Thanks.  Merged: https://github.com/libssh2/libssh2/commit/18cfec8336e88ab7df8a4427b803b9a177053674

mback2k:
This pull request / merge broke the WinCNG build, please take a look at the following autobuilds: https://libssh2build.uxnr.de/admin/waterfall

calavera:
Hey @mback2k sorry about that, but thanks for pointing me to the build. I'm not familiar with that code but I can open a new PR to add those functions. Do you think you can point me to some api documentation for LIBCRYPT32? I've not done much windows development but if it can parse the key strings we can have the full feature working there.

alamaison:
Fixed: https://github.com/libssh2/libssh2/commit/1de36eb5b4ae7328404b728927723705ad1f40ef
Testing ongoing: https://ci.appveyor.com/project/alamaison/libssh2

mback2k:
@calavera No problem. It should be possible to implement it using the existing functionality, like in _libssh2_wincng_pub_priv_keyfile and with _libssh2_wincng_asn_decode_bns. I will give it a try this weekend.
@alamaison Thanks for adding the stub functions. My builds are green again.

mback2k:
@calavera I just implemented this feature for WinCNG in 71d45d3df1579225b3b4a2e9d5df8cfbb13db4d9. Would you mind giving it a try?

PR #7, opened 2015-03-12
Must init ctx

willco007:
Must init ctx before using it or openssl will reuse the hmac which is not thread safe and causes a crash.


willco007:
Trying to cheery pick our changes for this issue; I think I have them all now.

alamaison:
Thanks.  Merged: https://github.com/libssh2/libssh2/commit/14d9ee01bc52ab55b44b1c402e810aa52e6a1364

PR #8, opened 2015-03-12
CMake build system.

alamaison:
Tested:
 - Windows:
    - Visual C++ 2005/2008/2010/2012/2013/MinGW-w64
    - static/shared
    - 32/64-bit
    - OpenSSL/WinCNG
    - Without zlib
 - Linux:
    - GCC 4.6.3/Clang 3.4
    - static/shared
    - 32/64-bit
    - OpenSSL/Libgcrypt
    - With/Without zlib
 - MacOS X
    - AppleClang 6.0.0
    - static
    - 64-bit
    - OpenSSL
    - Without zlib


bagder:
+1 from me

bagder:
Unfortunately this broke the configure build... (and if I may ask, I think we should avoid merge commits and keep a linear history)

alamaison:
What error are you seeing?  It seems to be working: https://drone.io/github.com/alamaison/libssh2/51

alamaison:
I'll do pull requests from master in future.

alamaison:
I think I know the problem, testing a fix now.

alamaison:
Fixed: https://github.com/libssh2/libssh2/commit/042993b8eb65d9d0b391f29fec8b589fa1f3baad

The CMake build didn't actually break the Autotools build, it just seemed that way because it made the error less obvious.  You probably forgot to run buildconf, am I right?

Previously, you would have seen an error, during configure, saying something like 'could not find src/libssh2_config.h.in'.  But I had added a file at that location for CMake.  buildconf would overwrite it, which was why the CI builds didn't notice anything.

bagder:
Yeps, works fine now. One little nit remains though. After I run buildconf and configure (running in the source tree), a file in the git repo is changed: example/libssh2_config.h.in.

I think we need to come up with a way to make this not happen.

alamaison:
Same cause:  Fixed now:
https://github.com/libssh2/libssh2/commit/bbbdf946a725e6d349a4e47d4c9cdd997295e25c

bagder:
Lovely, thanks. Confirmed to work for me now!

JxLiu2016:
@alamaison Hi, I build libssh2 with x86 Windows. Is it possible to build libssh2 with 64 windows?
I wish to use libssh2.lib with "Release 64" in Visual Studio 2015.

alamaison:
Yes, you just choose the 64 bit generator when you configure the build with
CMake

PR #9, opened 2015-03-15
Fix channel write loop in direct_tcpip example

jakob:
This pull request fixes the channel write loop in the direct_tcpip sample.

https://trac.libssh2.org/ticket/281
https://trac.libssh2.org/ticket/293


bagder:
Thanks a lot, merged in commit a1e744bb5eb2

PR #15, opened 2015-04-28
libssh2_channel_open: more detailed error message

jakob:
The error message returned by libssh2_channel_open in case of a server side channel open failure is now more detailed and includes the four standard error conditions in RFC 4254.


alamaison:
Thanks.  Merged: 4383a39d83e6212cb33323932a4c04032be8a84d

PR #20, opened 2015-06-12
two updates

tim-le:
1)added a I/O callback for select

2)support concurrent multiple opening channels, just tested for "direct-tcpip"


bagder:
Please do not do two different logical changes in the same pull request. Make one for each, it makes it easier to review and handle them independently of each other.

PR #22, opened 2015-06-12
support concurrent multiple opening channels

tim-le:
in non-blocking mode ,  now It's able to concurrently open many "direct-tcpip" channels.

### DIFF
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -654,6 +654,12 @@ libssh2_channel_direct_tcpip_ex(LIBSSH2_SESSION *session, const char *host,
 #define libssh2_channel_direct_tcpip(session, host, port) \
   libssh2_channel_direct_tcpip_ex((session), (host), (port), "127.0.0.1", 22)
 
+LIBSSH2_API LIBSSH2_CHANNEL *
+    libssh2_channel_direct_tcpip_alloc(LIBSSH2_SESSION *session, const char *host,
+    int port, const char *shost, int sport);
+LIBSSH2_API int
+    libssh2_channel_direct_tcpip_open(LIBSSH2_CHANNEL *channel);
+
 LIBSSH2_API LIBSSH2_LISTENER *
 libssh2_channel_forward_listen_ex(LIBSSH2_SESSION *session, const char *host,
                                   int port, int *bound_port, int queue_maxsize);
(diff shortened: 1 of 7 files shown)


bagder:
Please explain what didn't work and how this fixes those problems.

Also, we use the `libssh2_` prefix for exported symbols. Functions and symbols that are library-private should be prefixed with `_libssh2_"`.

This seems to provide two new exported functions. Is that on purpose and if so please explain why we want them and how they work!

tim-le:
yes, need two exported symbols, since I am new to libssh2 , I try as possible as I can to keep the existing logic code，so any ugly or redundant code just fix it if you think necessary.
Now let's say the problem about channel_direct_tcpip, imagine that a socks5 server accepts client and provides proxy service through one ssh session with many channels. 
I use channel_direct_tcpip for this purpose.
channel_direct_tcpip internally calls _libssh2_channel_open,  then a LIBSSH2_CHANNEL Object is created and saved at session->open_channel( note all the open_xxx variables ), then usually a NULL is returned with I/O ERROR_EAGAIN. 
Before the state of this channel is finally confirmed,  if socks5 server have got a new client with a new proxy target, and then try to open a second channel, _libssh2_channel_open couldn't deal with as expected. I thought _libssh2_channel_open should be able to continue sending a new open_packet for new channel. however _libssh2_channel_open keeps checking out all the session->open_xxx variables for the pending channel. well, the new proxy target stuck.

so my thoughts: Do not rely on session->open_xxx variables， I think they(open_xxx) should move into LIBSSH2_CHANNEL, but without changing the existing API logic, I copied them  into LIBSSH2_CHANNEL and created libssh2_channel_alloc and     
libssh2_channel_open2.

Finally, I'm happy to say, this is my first time developing with opensource project, a great opensource project.

tim-le:
I missed out to mention the LIBSSH2_CALLBACK_SELECT, 
we only have LIBSSH2_CALLBACK_SEND & LIBSSH2_CALLBACK_RECV, 
when I set the callbacks to provide a custom I/O underlying, maybe not a socket-based layer, so I found the internal select couldn't handle it. I need my CALLBACK_SELECT.
BTW,  int libssh2_session_handshake(LIBSSH2_SESSION *session, libssh2_socket_t socket),
it didn't handle well with none-socket.
but it's not a problem, I pass socket=0 to trick with it.

bagder:
I still don't understand.

Can you please write the documentation for your suggested new functions and I'll read that and then I might understand and/or have further comments. You could also consider taking this to the mailing list instead for better discussion and design of new functions.

bagder:
Out of date, closing.

PR #23, opened 2015-06-12
GNUmakefile: add -m64 CFLAGS when targeting mingw64, -m32/-m64 to LDFLAGS

vszakats:
`libssh2` mirror of `curl` patch https://github.com/bagder/curl/commit/d21b66835f2af781a3c2a685abc92ef9f0cd86be

This allows to build for the non-default target when using a multi-target mingw distro.
Also bump default OpenSSL dependency path to 1.0.2d.

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -14,7 +14,7 @@ endif
 
 # Edit the path below to point to the base of your OpenSSL package.
 ifndef OPENSSL_PATH
-OPENSSL_PATH = ../../openssl-0.9.8zc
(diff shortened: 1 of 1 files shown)


vszakats:
Updated OpenSSL version to 1.0.2d.

vszakats:
I'm using these settings to build libssh2 successfully in multi-target mingw scenario and it's working well.

Is there anything of concern to be updated here before it can be merged?

vszakats:
Albeit it's a valid patch and the same was accepted for libcurl, I'm closing this to not leave it lingering.

PR #40, opened 2015-09-06
Add support for HMAC-SHA-256 and HMAC-SHA-512.

bk2204:
Implement support for these algorithms and wire them up to the libgcrypt
and OpenSSL backends.  Increase the maximum MAC buffer size to 64 bytes
to prevent buffer overflows.  Prefer HMAC-SHA-256 over HMAC-SHA-512, and
that over HMAC-SHA-1, as OpenSSH does.

I did not wire these up to the Windows backend, as I have no Windows boxes to test on.  There should be sufficient stubs to make this compile on Windows, however, until someone can wire it up. By modifying the test-ssh2_fixture.sh file to limit the MACs option to a specific value, I can confirm interoperability with OpenSSH for both algorithms.

### DIFF
--- a/src/libgcrypt.h
+++ b/src/libgcrypt.h
@@ -42,6 +42,8 @@
 #define LIBSSH2_MD5 1
 
 #define LIBSSH2_HMAC_RIPEMD 1
+#define LIBSSH2_HMAC_SHA256 1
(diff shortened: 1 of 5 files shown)


bagder:
Looks like a fine and clean implementation. I'll merge soon if nobody else does!

PR #44, opened 2015-09-17
WinCNG: support for SHA256/512 HMAC

sbredahl:
Since SHA256/512 is now available via openssl/libcrypto and libgrypt backends, this commit should enable hmac-sha2-256 and hmac-sha2-512 being used with WinCNG as well. I have tested both against my openssh 6.6 server running on Ubuntu 14.04.

### DIFF
--- a/src/wincng.c
+++ b/src/wincng.c
@@ -93,6 +93,14 @@
 #define BCRYPT_SHA1_ALGORITHM L"SHA1"
 #endif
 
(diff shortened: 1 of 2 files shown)


bagder:
Lovely! Can I only ask you to set your name/email properly in the patch?

sbredahl:
I've recommited the files using proper username :-)

PR #51, opened 2015-10-05
kex.c: removed dupe entry from libssh2_kex_methods[]

sbredahl:
This removes a duplicate entry in the libssh2_kex_methods[] array, introduced in https://github.com/libssh2/libssh2/commit/fc4a969a0512e226de9b821496d20b9ddf53b741.


bagder:
Thanks!

Issue #50, opened 2015-10-02
libssh2_sftp_read() with maxlen of 8192 causes memory corruption


Corillian:
When calling `libssh2_sftp_read()` with a buffer size of 4096 my application works as expected. If I do nothing other than change the buffer size to 8192 the application begins crashing due to memory corruption within libssh2. I am using a blocking socket with libssh2 set to blocking mode. Problems start occurring when libssh2 arbitrarily skips ahead within the file. Sometimes it does it with the first read and sometimes it doesn't do it until after multiple successful reads:
```
NOTE: Requested size is always 8192

GetBuffer() [New] About to read from requested file position 0 and actual file position 0
GetBuffer() [New] Read completed: bytes read = 2768, totalByteRead = 2768, new actual file position = 10960
GetBuffer() [New] About to read from requested file position 0 and actual file position 10960
GetBuffer() [New] Read completed: bytes read = 5424, totalByteRead = 8192, new actual file position = 16384
```
If you look at the log output above you can see that the libssh2 file pointer position started at 0. 8192 bytes were requested, 2768 were returned (it's always this number once problems begin), and the new file  pointer after read has somehow jumped to **10960**.

I compiled libssh2 and all of its dependencies as 64-bit DLL's using Visual Studio 2013. It's linking to OpenSSL 1.0.2d and zlib 1.2.8. All of my testing was done with Windows 7 x64. My sequence of actions is:
1.  I set function callbacks with OpenSSL for creating and destroying locks
2. Initialize winsock and libssh2
3. Create a (blocking) socket
4. Do the ssh handshake using a private key in memory
5. Create the sftp session and open a file. When the sftp session is created I set function pointers to a custom set of memory allocation functions that just call `malloc()`, `realloc()`, and `free()`
6. Begin reading from the file. Every call to `libssh2_sftp_read()` is preceded by `libssh2_sftp_seek64()` to explicitly set the file pointer

I can't provide a callstack because the memory corruption appears to be nuking it. All I get is an access violation at `msvcr120.dll!000007fedea9c3f9() Unknown`. If there's any additional information I can provide please let me know.

sbredahl:
I can confirm that some weird things happen if the buffer maxlen is much larger than 4k (7k works for me also, but 8k is too large). I end up in an infinite loop where libssh2_sftp_read() keeps returning 0 (in blocking mode). 

I don't get any access violation however (also Win7 x64).

sbredahl:
[...]

PR #53, opened 2015-10-14
sftp.c: max_read_ahead should not exceed MAX_SFTP_READ_SIZE

sbredahl:
As discussed in https://github.com/libssh2/libssh2/issues/50, the recent changes in the sftp_read() implementation causes data corruption if the buffer_maxlen parameter used in calls to libssh2_sftp_read() is "too large", say 8192 bytes.

It appears that the max_read_ahead value computed in sftp.c must not exceed MAX_SFTP_READ_SIZE (which currently happens if buffer_maxlen > 7500) and this commit limits max_read_ahead in that situation. 

(Also, I'm pretty sure that this is not "the right solution" but it is better than not having it. )

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -1313,7 +1313,7 @@ static ssize_t sftp_read(LIBSSH2_SFTP_HANDLE * handle, char *buffer,
             /* Number of bytes asked for that haven't been acked yet */
             size_t already = (size_t)(filep->offset_sent - filep->offset);
 
-            size_t max_read_ahead = buffer_size*4;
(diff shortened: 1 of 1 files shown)


bagder:
Without understanding _why_ the size causes a problem this feels like nothing more than covering over the actual problem and hence I am against applying it.

sbredahl:
I don't disagree, but you should probably remove the changes from commit https://github.com/libssh2/libssh2/commit/d754fee2f24565d8c89dc37a7f05ba1222a7bfc4 then.

PR #70, opened 2015-12-25
add OpenSSL 1.1.0 compatibility

vszakats:


bagder:
Looks good!

vszakats:
Thanks! Squashed into a single commit.

Because caller code doesn't (except for a few cases) check the result of `libssh2_{sha1,sha256,md5,hmac_ctx}_init()` functions &mdash; in a further update it may be useful to prepare related hash update/deinit functions to catch `ctx == NULL` and gracefully fail, instead of forwarding that `NULL` to OpenSSL which would happily [dereference it](https://github.com/openssl/openssl/blob/master/crypto/evp/digest.c#L257) without verification.

Not sure what a graceful failure mode could be, or if it's better to crash in such case. (Similar question likely applies to existing SSL engines as well.)

vszakats:
Updated for 1.1.0-pre2.

bagder:
Thanks!

PR #73, opened 2016-01-07
add placeholder for library name depending on platform

minodada:


bagder:
Why would we want this?

dadaleal:
Hello,
 
currently the linker create the ssh2.lib but libgit2 expected libssh2.lib. This leads to Errors.
Modification: If MSVC compiler is used, create libssh2.lib instead ssh2.lib.
 
Best regards
 
Date: Mon, 18 Jan 2016 04:46:13 -0800

alamaison:
I don't see how this PR achieves that goal.  It changes the flags that are inserted in libssh2.pc on builds with MSVC, but that's all.  It doesn't change the name of the output library.

minodada:
I did not so clearly explained above.
"It changes the flags that are inserted in libssh2.pc " : yes that's right.
The line in libss2.pc without modification: Libs: -L${libdir} -lssh2 -lbcrypt -lcrypt32
With modification: Libs: -L${libdir} -llibssh2 -lbcrypt -lcrypt32. So the linker try to link libssh2.lib and not ssh2.lib when builds libgit2. (MSVC compiler create libssh2.lib instead ssh2.lib)

alamaison:
I still don't really understand.  Is libgit2 using libssh2.pc to find the flags for MSVC.  I'd be surprised if that worked, regardless of the name.

If you mean that libgit2 is built with something like mingw, but wants to link against an MSVS-built libssh2.lib by using the .pc file flags, the I would have thought mingw would want -lssh2 and would add the prefix itself, no?

PR #75, opened 2016-01-13
Fix libssh2_sftp_read issues

jakob:
I have reviewed sftp_read() and discovered 3 separate bugs that cause data corruption.

Two of them were introduced by https://github.com/libssh2/libssh2/commit/d754fee2f24565d8c89dc37a7f05ba1222a7bfc4

I have tested this code with example-sftp and various read buffer sizes. I believe this fixes the data skipping issues reported by @sbredahl in #50.

However, I don't think the bugs that I have fixed could cause memory corruption as initially reported by @Corillian, so we probably have to keep digging.

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -1367,7 +1367,9 @@ static ssize_t sftp_read(LIBSSH2_SFTP_HANDLE * handle, char *buffer,
 
         while(count > 0) {
             unsigned char *s;
-            uint32_t size = MIN(MAX_SFTP_READ_SIZE, count);
+            uint32_t size = count;
+            if (size < buffer_size) size = buffer_size;
+            if (size > MAX_SFTP_READ_SIZE) size = MAX_SFTP_READ_SIZE;
 
             /* 25 = packet_len(4) + packet_type(1) + request_id(4) +
                handle_len(4) + offset(8) + count(4) */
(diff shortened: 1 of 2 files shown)


sbredahl:
@jakob I applied your 3 patches "by hand" and they do not seem to change the situation here. I still need to keep the buffer size <= SFTP_MAX_READ_SIZE / 4 to avoid data corruption. 

Using example-sftp with a 8k buffer and 10 byte remote file, my downloaded file is 0 bytes. Can you replicate this? 

(I'm using ubuntu x64 client and server here).

jakob:
@sbredahl I couldn't replicate your problem, but I have found and fixed another bug (4973df3).

I have now added a bunch of sanity checks to sftp_read() to make it more likely that errors are detected. Could you try if it works (or at least reports an error) with these new commits?

sbredahl:
@jakob I cannot reproduce any of the issues I had before with your latest set of patches! I think you nailed it this time - thanks a bunch!

jakob:
@DerekLuer is there a problem with line 1588?

bagder:
Thanks a lot for your work. I also appreciate that we have gotten some testing on this patch set, as it makes me feel safer with merging it.

There are however a few nits on the style I think:
- there are a few trailing white spaces
- you've indented with tabs instead of spaces in a few places
- we write source code lines that wrap before column 80
- "if (size < buffer_size) size = buffer_size;" should be written in two lines
- we don't use // comments as we want to comply with C89 standard (and really old compilers)

jakob:
Thanks for your comments @bagder 

I have now fixed the coding style in the parts that I touched.

bagder:
I still get this:

> sftp.c:1378:13: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]

And please squash/fixup your code style fixes so that this patch set is only a series of logical fixes that fix existing problems, not problems within the patch series itself.

jakob:
@bagder thanks for your comments. It took me a while to fix the commits, but I managed to do so. Please see #79 for a cleaned up version of this pull request.

PR #80, opened 2016-02-10
GNUmakefile: list system libs after user libs

vszakats:
Otherwise some referenced WinSock functions will fail to 
resolve when linking against LibreSSL 2.3.x static libraries 
with mingw.


vszakats:
This may possibly be required not only for LibreSSL, but other variants as well. The canonical way to list Windows system libraries is to put them after any custom ones (but at least _after_ the objects dependent on them).

bagder:
thanks!

PR #81, opened 2016-02-10
fix support for LibreSSL after having added OpenSSL 1.1.0-pre1/2 support

vszakats:
Test log: https://ci.appveyor.com/project/vsz/harbour-deps/build/1.0.478

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -177,7 +177,7 @@ _libssh2_cipher_init(_libssh2_cipher_ctx * h,
                      _libssh2_cipher_type(algo),
                      unsigned char *iv, unsigned char *secret, int encrypt)
 {
-#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+#ifdef HAVE_OPAQUE_STRUCTS
     *h = EVP_CIPHER_CTX_new();
     return !EVP_CipherInit(*h, algo(), secret, iv, encrypt);
 #else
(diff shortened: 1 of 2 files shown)


vszakats:
Test log: https://ci.appveyor.com/project/vsz/harbour-deps/build/1.0.478

asavah:
Indeed this fixes building libssh2 against libressl-portable git tree.

bagder:
So how's the status of this when I've just merged the openssl 1.1.0 pre3 PR ?

vszakats:
This patch remains required after applying the other one.

bagder:
It does add a lot of `#ifdef` lines with lots of details. What would you say about making the code like this instead:

At the top of the file:
~~~
#if OPENSSL_VERSION_NUMBER >= 0x10100000L && \
    !defined(LIBRESSL_VERSION_NUMBER)
#define HAVE_OPAQUE_PKEY 1
#endif
~~~

... and then further down we only do `#ifdef HAVE_OPAQUE_PKEY` when we have that difference. I think it'll make the code gentler on the eye.

vszakats:
I was "too fast" with a previous set of patches, where applying OpenSSL 1.1.0-pre1/2 updates. They consequently broke comparibility with LibreSSL (which apparently didn't follow the struct opacity changes of OpenSSL as of yet). This patch disables those updates using the same #if checks as used in curl sources, when compiled against LibreSSL.

vszakats:
I agree. Will rework this later today accordingly, if that's fine with you.

bagder:
Thanks!

vszakats:
I've updated the patch. Chosen a more generic name for the macro, no change otherwise. [Test build](https://ci.appveyor.com/project/vsz/harbour-deps/build/1.0.589) went fine.

vszakats:
Rebased on master and updated to use the new macro for the previously applied patch 298f056fd6931e5e8bedc4fce5142a46c8f9f4b9 as well.

bagder:
thanks!

PR #82, opened 2016-02-13
OpenSSL 1.1.0-pre3 compatibility

vszakats:


vszakats:
Let me restart this one.

Issue #11, opened 2015-03-24
OpenSSL HMAC_cleanup, and HMAC_Init are stated as deprecated


bagder:
According to OpenSSL commit 00a5a74bbc854972c4 (Feb 10 2015). https://git.openssl.org/?p=openssl.git;a=commit;h=00a5a74bbc854972c4484a0b204bc9437d2757d9

Thus, building libssh2 with an OpenSSL from then or later causes link errors.

vszakats:
Unless I'm misreading something obvious, this issue should be resolved after 73930e657709a571f4ea0d9089dbc16ba167aeaa.

I can also confirm successful builds using latest OpenSSL 1.1.0 master/HEAD alongside the latest revision of libssh2.

bagder:
Ok, let me double-check my build setup so that I didn't mess something up...

bagder:
Should be moot now, especially after 1.1.0 support was merged in commit 64ebfd8182.

PR #83, opened 2016-02-13
add OpenSSL 1.1.0-pre3-dev compatibility

vszakats:
by using API instead of accessing an internal structure.


bagder:
thanks!

bagder:
just a FYI that #11 seems to still plague us with recent openssl

vszakats:
Left a comment there. I think the issue should be resolved by now.

PR #84, opened 2016-02-19
Resolve improper handling of remote eof responses

vnsavage:
1) If we return non-null pointer, we need to clear the session recv
state, otherwise subsequent requests on the session are going to
segfault.
2) The libssh2_channel_eof usage for checking EOF here is
incorrect, because it can mask a real EOF.


salva:
#104 would make handling (2) easier.

salva:
I think the fix for (1) is not correct.

If the channel is not at EOF, the read operation should be retried, handled as an EAGAIN error.

salva:
I have added a patch into #104 that fixes this issue properly.

willco007:
This is issue closable? It appears it's been moved to #145.

PR #88, opened 2016-02-24
Fix DH kex issues

yumkam:
Passes `make check`, but not deeply tested.

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -115,10 +115,13 @@ static int diffie_hellman_sha1(LIBSSH2_SESSION *session,
                                kmdhgGPshakex_state_t *exchange_state)
 {
     int ret = 0;
+    int pbits;
     int rc;
     libssh2_sha1_ctx exchange_hash_ctx;
 
     if (exchange_state->state == libssh2_NB_state_idle) {
+        int p_bits;
+
         /* Setup initial values */
         exchange_state->e_packet = NULL;
         exchange_state->s_packet = NULL;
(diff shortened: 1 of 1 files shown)


yumkam:
PS those "check packet boundaries" issues are all over codebase. `grep ntoh *.c` and cry.

src/kex.c
```diff
@@ -133,7 +135,9 @@ static int diffie_hellman_sha1(LIBSSH2_SESSION *session,
         memset(&exchange_state->req_state, 0, sizeof(packet_require_state_t));
 
         /* Generate x and e */
-        _libssh2_bn_rand(exchange_state->x, group_order, 0, -1);
+        /* FIXME: too expensive, should not need more than 2*symmetric_bits */
```
adrienthebo: I believe that this property only holds where p is a safe prime. This should be fine for kex_method_diffie_helman_group14_sha1 and kex_method_diffie_helman_group1_sha1 but in kex_method_diffie_helman_group_exchange_sha1 and kex_method_diffie_helman_group_exchange_sha256 we're negotiating a "random(ish)" prime. Given that generating a safe prime is pretty time intensive there's a very good chance that this won't use safe primes often, so using short exponents with a p that isn't a Sophie Germaine prime seems dangerous.

I'm not a cryptographer by any stretch of the definition but http://link.springer.com/chapter/10.1007%2F3-540-68339-9_29 does make a case that this is risky.

yumkam:
@adrienthebo 
However, ssh generates `p` values **offline**, so it can freely verify that they are safe primes, see [rfc4419](https://tools.ietf.org/html/rfc4419#section-3). And it does (see `man 5 moduli`).
And that's why [rfc4419](https://tools.ietf.org/html/rfc4419#section-6.2) that defines ssh's diffie-hellman-group-exchange protocol (not only fixed ones!) permit use of smaller exponents.
Anyway, this commit have not performed any optimization (safe or not), it only reduced private exponent size to maximum permitted by rfc4419.

adrienthebo:
@yumkam fair enough - I would rather speak up and be demonstrated incorrect rather than stay silent and let a real issue slide. Thanks for pointing that out. :)

PR #93, opened 2016-02-27
docs: correctly describe channel_wait_eof

aidanhs:
channel_wait_eof waits for channel->remote.eof, which is set on
receiving a `SSH_MSG_CHANNEL_EOF` message. This message is sent
when a party has no more data to send on a channel.

See https://github.com/alexcrichton/ssh2-rs/issues/34#issuecomment-184850856 if you want a complete run-through of me coming to this conclusion


salva:
The change is correct.

PR #94, opened 2016-02-27
Fix MSVC 14 compilation errors

Zenju:
For _MSC_VER == 1900 these macros are not needed and create problems:

1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\stdio.h(1925): warning C4005: 'snprintf': macro redefinition (compiling source file libssh2-files\src\mac.c)
1> \win32\libssh2_config.h(27): note: see previous definition of 'snprintf' (compiling source file libssh2-files\src\mac.c)
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\stdio.h(1927): fatal error C1189: #error: Macro definition of snprintf conflicts with Standard Library function declaration (compiling source file libssh2-files\src\mac.c)

### DIFF
--- a/win32/libssh2_config.h
+++ b/win32/libssh2_config.h
@@ -24,6 +24,7 @@
 #define HAVE_SELECT
 
 #ifdef _MSC_VER
+#if _MSC_VER < 1900
 #define snprintf _snprintf
 #if _MSC_VER < 1500
 #define vsnprintf _vsnprintf
@@ -31,6 +32,7 @@
 #define strdup _strdup
 #define strncasecmp _strnicmp
 #define strcasecmp _stricmp
+#endif
 #else
 #ifndef __MINGW32__
 #define strncasecmp strnicmp


alamaison:
I've merged your fix, thanks.  However, these old Windows build files might get removed entirely in the future.  They are unmaintained and can never support the full range of toolchain combinations out there.  Instead you should use CMake to generate build files for your toolchain combination of choice.

Zenju:
Thanks, but I'm not using the provided Windows build files, but created my own VS 2015 project for libssh2. The requirement for me is to build libssh2 with the most recent VS release and by setting up the VS project myself I can fullfill all the specific additional requirements that are needed in my scenario. So personally I don't need officially maintained VS project files, but I hope you keep your libssh2 build setup as simple as it is now so that it's easy to create custom projects.

alamaison:
I'm not opposed to leaving `win32/libssh2_config.h` in-place if you find it useful.  But bear in mind that the 'simplicity' is an illusion created by hard-coding arbitrary stuff.  That stuff is constantly out-of-date.

I would encourage you to look at whether you can satisfy your additional requirements using CMake.  After all, VS2015 is only the latest version now.

Issue #24, opened 2015-06-16
Get rid of warnings in 64-bit


Zenju:
Hi,

it would be great if I could compile libssh2 without warnings on 64-bit:
1. Currently I see lots of:
   warning C4267: 'argument' : conversion from 'size_t' to 'unsigned int', possible loss of data

Which is caused by the macros using "strlen" returning size_t but the _ex functions taking "unsigned int", e.g.
# define libssh2_sftp_rmdir(sftp, path) \

```
libssh2_sftp_rmdir_ex((sftp), (path), strlen(path))
```

libssh2_sftp_rmdir_ex(LIBSSH2_SFTP *sftp, const char *path,
                      unsigned int path_len)

In general it seems that the "unsigned int" should be replaced by size_t since this is also what std::string::length() returns.
1. Here's another warning that's more serious, although we still have 23 years to go:
   warning C4244: '=' : conversion from 'int64_t' to 'unsigned long', possible loss of data
   
   ```
   std::int64_t mTime = ...;
   LIBSSH2_SFTP_ATTRIBUTES attribNew = {};
   attribNew.mtime = mTime; //32-bit target! loss of data!
   ```
2. There are more warnings in the implementation files that are shown when compiling libssh2 with Visual Studio 2013 in 64 bit.

bagder:
I agree, no warnings would be ideal. But remember that we cannot change any types in the public API/ABI so where `unsigned int` is used today we cannot freely change it to `size_t`.

In the (1) case we need a typecast in that macro. The (2) thing is part of the protocol, isn't it?

Feel free to submit pull requests to fix warnings.

Zenju:
> remember that we cannot change any types in the public API/ABI so where unsigned int is used today we cannot freely change it to size_t.

I'm primarily interested in a practical solution to my current problem, which is that I see lots of compiler warnings, all of which are extremely useful to find 64-bit bugs in general. So I don't want to disable them, not even for a single translation unit.

On the other hand this means I cannot use the libssh2 function macros since they are not type-correct. One mitigation might be to add type casts to the macros.

This leaves the problem to the _ex functions only, but at least I can then type-cast on my own code, which is still wordy and ugly since I would prefer an API without the "len" parameters, which seem about CPU-perf optimization only.

> The (2) thing is part of the protocol, isn't it?
Yes, maybe it's time to upgrade to a newer version of the spec?
https://filezilla-project.org/specs/draft-ietf-secsh-filexfer-13.txt
[...]

PR #96, opened 2016-02-28
Get rid of 64-bit build warnings

Zenju:
Get rid of all of those:
2>fs\sftp.cpp(1467): warning C4267: 'argument': conversion from 'size_t' to 'unsigned int', possible loss of data
In reference to:
https://github.com/libssh2/libssh2/issues/24

### DIFF
--- a/include/libssh2_sftp.h
+++ b/include/libssh2_sftp.h
@@ -227,10 +227,10 @@ LIBSSH2_API LIBSSH2_SFTP_HANDLE *libssh2_sftp_open_ex(LIBSSH2_SFTP *sftp,
                                                       unsigned long flags,
                                                       long mode, int open_type);
 #define libssh2_sftp_open(sftp, filename, flags, mode) \
-    libssh2_sftp_open_ex((sftp), (filename), strlen(filename), (flags), \
+    libssh2_sftp_open_ex((sftp), (filename), (unsigned int)strlen(filename), (flags), \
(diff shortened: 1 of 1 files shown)


alamaison:
This just silences the warning but doesn't solve the underlying problem, which is that the API should use size_t.  But we can't change it immediately because of backward-compatibility.

Zenju:
Is there really a problem? Andrei Alexandrescu says "use 64-bit code but 32-bit data", and I agree with him. I don't think there really is a valid case of libssh2 having to accept input sizes larger than 4 billion in its API (at least in the above cases).

alamaison:
Alexandrescu was talking in the context of squeezing all possible performance out of the CPU.  It's not a general rule.

The problem is not having to accept input sizes > 4GB but, rather that the C functions use size_t to represent lengths.  The caller should be able to pass the result of such a call to our API without having to cast.  That they can't is the cause of the warnings you are seeing.

Zenju:
The current situation is that the warnings originate from libssh2 code and there is nothing a 64-bit libssh2 user can do about it. If this will be solved in the future by supporting size_t that's fine, but for those users who don't want to hold their breath, please silence these warnings. PS: Just ignoring these warnings on a per-project/per-file/per-header level might hide other more important warnings, so it's no solution.

alamaison:
The plan is to add a #define that activates a non-backward-compatible ABI.  I'm motivated to fix this because the warnings drive me crazy too, but can't promise when

PR #104, opened 2016-04-22
libssh2_channel_read_ex: 0 means EOF

salva:
Until now, libssh2_channel_read_ex was returning 0 for two cases:

1) At EOF

2) After the data transport layer had produced data, but not for the
current channel.

And it was not possible to discern between the two conditions in any
way (note that libssh2_channel_eof would not return true when data is
still queued for the the other channel stream - !ext).

This patch, changes that behavior such that now, only on the first
case (at EOF) a zero is returned.

And the second case is converted to an EAGAIN.

### DIFF
--- a/src/channel.c
+++ b/src/channel.c
@@ -1889,11 +1889,17 @@ ssize_t _libssh2_channel_read(LIBSSH2_CHANNEL *channel, int stream_id,
            return code. */
         if(channel->remote.eof || channel->remote.close)
             return 0;
-        else if(rc != LIBSSH2_ERROR_EAGAIN)
-            return 0;
 
-        /* if the transport layer said EAGAIN then we say so as well */
-        return _libssh2_error(session, rc, "would block");
+        /* If the transport layer said EAGAIN then we say so as well.
+           If there wasn't any error, it means the transport layer got
+           new data but it went to some other place and so we return
+           EAGAIN too. */
+        if((rc == LIBSSH2_ERROR_EAGAIN) || (rc == LIBSSH2_ERROR_NONE))
+            return _libssh2_error(session, LIBSSH2_ERROR_EAGAIN, "would block");
+
+        /* should never happen */
+        _libssh2_debug(session, LIBSSH2_TRACE_CONN, "Internal error: bad code path reached (%d)", rc);
+        return rc;
     }
 
     channel->read_avail -= bytes_read;
--- a/docs/libssh2_channel_read_ex.3
+++ b/docs/libssh2_channel_read_ex.3
@@ -34,8 +34,8 @@ Actual number of bytes read or negative on failure. It returns
 LIBSSH2_ERROR_EAGAIN when it would otherwise block. While
 LIBSSH2_ERROR_EAGAIN is a negative number, it isn't really a failure per se.
 
-Note that a return value of zero (0) can in fact be a legitimate value and
-only signals that no payload data was read. It is not an error.
+A return value of zero (0) indicates EOF.
+
 .SH ERRORS
 \fILIBSSH2_ERROR_SOCKET_SEND\fP - Unable to send data on socket.
 
--- a/src/scp.c
+++ b/src/scp.c
@@ -738,11 +738,9 @@ scp_recv(LIBSSH2_SESSION * session, const char *path, libssh2_struct_stat * sb)
   scp_recv_empty_channel:
     /* the code only jumps here as a result of a zero read from channel_read()
        so we check EOF status to avoid getting stuck in a loop */
-    if(libssh2_channel_eof(session->scpRecv_channel))
-        _libssh2_error(session, LIBSSH2_ERROR_SCP_PROTOCOL,
-                       "Unexpected channel close");
-    else
-        return session->scpRecv_channel;
+    _libssh2_error(session, LIBSSH2_ERROR_SCP_PROTOCOL,
+                   "Unexpected channel close");
+
     /* fall-through */
   scp_recv_error:
     tmp_err_code = session->err_code;


salva:
Note that this change doesn't break backward compatibility. It just gives a precise and **useful** meaning to the zero return value which it hadn't.

src/channel.c
```diff
@@ -1889,11 +1889,17 @@ ssize_t _libssh2_channel_read(LIBSSH2_CHANNEL *channel, int stream_id,
...
-        else if(rc != LIBSSH2_ERROR_EAGAIN)
-            return 0;
 
-        /* if the transport layer said EAGAIN then we say so as well */
-        return _libssh2_error(session, rc, "would block");
+        /* If the transport layer said EAGAIN then we say so as well.
+           If there wasn't any error, it means the transport layer got
+           new data but it went to some other place and so we return
+           EAGAIN too. */
+        if((rc == LIBSSH2_ERROR_EAGAIN) || (rc == LIBSSH2_ERROR_NONE))
```
alamaison: `_libssh2_transport_read` returns `LIBSSH2_ERROR_NONE` if `session->socket_state == LIBSSH2_SOCKET_DISCONNECTED`.  Is retrying with EAGAIN the correct response to this?  Or should `_libssh2_transport_read` be changed to return `LIBSSH2_ERROR_SOCKET_DISCONNECT` in that case?

I'm not sure i understand 'it means the transport layer got new data but it went to some other place'.  What other place? If I'm reading the code right, `rc == LIBSSH2_ERROR_NONE` can happen if the first byte of the payload is 0 (see fullpacket).  What does that signify?

salva: After so long, I can not remember the details.

I think that I was getting spurious LIBSSH2_ERROR_NONE return codes from libssh2_channel_read, and I did supposed it was caused by data being eat at some upper layer.

But now, looking at the code (`libssh2_transport_read`), the only ways in which a LIBSSH2_ERROR_NONE can be returned are the two pointed by you and which should not happen under normal circumstances (or are servers sending packets of type 0?). So, maybe I was just trying to fix a bug that was not there at all.

Anyway, I still think that reserving 0 for signaling an EOF condition is a must. 

And the LIBSSH2_SOCKET_DISCONNECTED condition should also be handled properly, returning that error or 0 for EOF.

I would try to get back over this and recheck it in the next days.

alamaison: Thanks @salva.  Apologies for the delay

salva:
I have been looking at the places where `_libssh2_transport_read` is being called, and it seems to me that in most (if not all) of then the two possible conditions that cause a zero rc (socket disconnected and receiving a packet with type zero) are not handled correctly.

I think the right way to fix that would be to make `_libssh2_transport_read` return the LIBSSH2_ERROR_SOCKET_DISCONNECT error code and change the calling code to handle that correctly.

Digging into the git history, there doesn't seem to be a particular reason to have `_libssh2_transport_read` return 0 when the socket is disconnected and it seems to have been that way since forever (9d55db6501aa4e21f0858cf36cdc2ddc11b96e83).

These are the places where `_libssh2_transport_read` is called:

* [`channel_forward_accept`](https://github.com/libssh2/libssh2/blob/master/src/channel.c#L727): both type-0-packet and socket-disconnected result in a LIBSSH2_ERROR_CHANNEL_UNKNOWN error

* [`_libssh2_channel_read`](https://github.com/libssh2/libssh2/blob/master/src/channel.c#L1803): both conditions result in a zero return code.

* [`_libssh2_channel_write`](https://github.com/libssh2/libssh2/blob/master/src/channel.c#L2041): a type-0-packet stops the loop that reads packets before the write starts, that could result in a dead lock in some quite unlikely edge cases.

* [`channel_wait_eof`](https://github.com/libssh2/libssh2/blob/master/src/channel.c#L2265): A socket-disconnected condition results in an infinite loop.

* [`_libssh2_channel_close`](https://github.com/libssh2/libssh2/blob/master/src/channel.c#L2357): It seems to me that here, `_libssh2_transport_read` is wrapped in a loop which only pourpose is discarding spurious 0 return codes not caused by a LIBSSH2_SOCKET_DISCONNECTED condition.

* [`channel_wait_closed`](https://github.com/libssh2/libssh2/blob/master/src/channel.c#L425): The zero return code from _libssh2_transport_read results in the function succeeding (wrongly, I believe).

* [`_libssh2_packet_require`](https://github.com/libssh2/libssh2/blob/master/src/packet.c#L1117): I don't really understand what's going on here. There is a explicit check for a zero return value that when happens is used to check for timeouts, but as the library is non-blocking now at that layer, that may be the remainder of some pre non-blocking code.

* [`_libssh2_packet_burn`](https://github.com/libssh2/libssh2/blob/master/src/packet.c#L1183): It says...
```
        if (ret == 0) {
            /* FIXME: this might busyloop */
            continue;
        }
```

* [`_libssh2_packet_requirev`](https://github.com/libssh2/libssh2/blob/master/src/packet.c#L1212): The call is wrapped in a loop that skippes 0 return codes not caused by a disconnected socket condition.

* [`libssh2_poll`](https://github.com/libssh2/libssh2/blob/master/src/session.c#L11677): This is a quite long an deprecated function and I have run out of steam!



I would really appreciate if somebody else could take a look at my analysis.

alamaison:
@salva Thanks for your detailed analysis.

`_libssh2_transport_read` is documented:
> Returns packet type added to input queue (0 if nothing added), or a negative error number.

I would argue that '0 if nothing added' should be incorrect.  If we handle socket-disconnection properly by returning LIBSSH2_ERROR_SOCKET_DISCONNECTED then we are only left with packet-type 0 causing a zero return type.  Packet-type 0 isn't a valid SSH message number but i don't think it's the transport layer's job to read anything into that.  It should return 0 (as it does) and the clients should consider that the packet type, rather than 'nothing added'.  Various loops such as 
```
    do {
        rc = _libssh2_transport_read(listener->session);
    } while (rc > 0);
```
should change to
```
    do {
        rc = _libssh2_transport_read(listener->session);
    } while (rc >= 0);
```
and it's up to the channel layer to handle the invalid packet type.  Each of the `*_channel_*` functions you listed could check for zero (knowing now that it's not socket disconnected) and return a generic error.

> _libssh2_packet_require: I don't really understand what's going on here. There is a explicit check for a zero return value that when happens is used to check for timeouts, but as the library is non-blocking now at that layer, that may be the remainder of some pre non-blocking code.

The code looks useful to me.  It's implementing a timeout - just in a non-blocking way.  I think the condition `else if (ret == 0)` should change to just `else` (i.e. `ret != packet_type`).  That means that non-matching packets will also be subject to the timeout, which seems to me like the correct behaviour.

Do you concur?

I wonder is resolving this will fix some of the random hanging I sometimes see.

salva:
> I would argue that '0 if nothing added' should be incorrect. If we handle socket-disconnection ...
> ... could check for zero (knowing now that it's not socket disconnected) and return a generic error.

That's what I have done in #145, could you check it?

> The code looks useful to me. It's implementing a timeout - just in a non-blocking way...

But the thing is that this code is not reached most of the time, because `_libssh2_transport_read` would return an EAGAIN error when no data is available from the network. Compare it with the code in `_libssh2_packet_requirev` where the timeout test runs when `_libssh2_transport_read` returns EAGAIN (well, or 0!).

BTW, couldn't `_libssh2_packet_require` be written on top of `_libssh2_packet_requirev` in order to eliminate code duplication?

In any case, it seems to me that timeout handling needs also to be revised all along the code base, but I would push that out of the scope of this ticket.

alamaison:
#145 looks good to me.

> But the thing is that this code is not reached most of the time ...

Good point.  See my comments in #145.

> BTW, couldn't _libssh2_packet_require be written on top of _libssh2_packet_requirev in order to eliminate code duplication?

Agreed

salva:
Conversation moved to #145.

PR #106, opened 2016-05-08
mbedTLS crypto backend

wildart:
This is mbedTLS crypto backend (libmbedcrypto) for libssh2 cryptography.


EricForgy:
Hi libssh2 :)

For what it's worth, this would be super helpful for Julia users with private GitHub repos as discussed in the issue @wildart referenced above. Cheers

bagder:
@EricForgy, is that an "independent" claim that the patch works as intended?

EricForgy:
Hi @bagder, not an independent claim (I'm not qualified). Just morale support :)

If it helps, I'll see if someone in that Julia issue can help with an independent verification.

@wildart, is there some simple thing I can do to help? It was a challenge for me just to get SSH working with GitHub in the first place, so I'm just hoping Julia gets fixed.

wildart:
I changed travis configuration to include to mbedtls support. I wasn't sure where to put build scripts, so I left them in the project root.

cmake/FindmbedTLS.cmake
```diff
@@ -0,0 +1,82 @@
...
+FIND_PATH(MBEDTLS_INCLUDE_DIR
+    NAMES mbedtls/ssl.h
+    PATHS
+        ${MBEDTLS_ROOT_DIR}/include
+)
+FIND_LIBRARY(mbedtls_lib
+    NAMES mbedtls libmbedtls libmbedx509
+    PATHS
+        ${MBEDTLS_ROOT_DIR}/library
+        ${MBEDTLS_ROOT_DIR}/build/library
```
tkelman: these should probably refer to an install tree, not a build tree (which certainly isn't guaranteed to be called "build")

Keno:
What's the status of this? What's needed to move this forward?

Keno:
Just another friendly ping. I'd be happy to help out with whatever is needed here.

bagder:
@Keno did you try it out?

mbedtls/buildmbedtls
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+set -ex
+
+MBEDTLS_SRC_ARCH=mbedtls-2.2.1
+
+# download mbedtls sources
+cd mbedtls
+wget https://github.com/ARMmbed/mbedtls/archive/$MBEDTLS_SRC_ARCH.tar.gz
+tar -xzf $MBEDTLS_SRC_ARCH.tar.gz
```
blakejohnson: I realize that this file mainly exists to set up testing on TravisCI, but this doesn't work on macOS because `wget` is not available by default. You could conditionally substitute `wget` with `curl -O -L`. Alternatively, you could replace this file with something in the top-level CMakeLists.txt with the `ExternalProject_Add` command. I'm winging this, but it would be something like:
```cmake
SET(MBEDTLS_SRC_ARCH mbedtls-2.2.1)
ExternalProject_Add(
  URL https://github.com/ARMmbed/mbedtls/archive/${MBEDTLS_SRC_ARC}.tar.gz
  PATCH patch -p0 < ../mbedtls.patch
  CMAKE_ARGS -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_SHARED_MBEDTLS_LIBRARY=ON
  BUILD_COMMAND make
  INSTALL_COMMAND make install
)
```

wildart: `ExternalProject` is executed during the build time, however mbedtls installation is needed before the configuration step.

Keno:
I finally got around to trying this out and it worked for me on all platforms that I tried (Linux, Mac, Windows).

tkelman:
I was also able to get this to pass the tests on x86_64 ubuntu 14.04 via
```sh
git clone https://github.com/ARMmbed/mbedtls -b mbedtls-2.2.1
git clone https://github.com/wildart/libssh2 -b mbedtls
mkdir -p mbedtls/build
mkdir -p libssh2/build
cd mbedtls
patch -p0 < ../libssh2/mbedtls/mbedtls.patch
cd build
cmake .. -DUSE_SHARED_MBEDTLS_LIBRARY=ON -DCMAKE_INSTALL_PREFIX=$PWD/../../usr
make -j`nproc` install
cd ../../libssh2/build
cmake .. -DCRYPTO_BACKEND=mbedTLS -DCMAKE_INSTALL_PREFIX=$PWD/../../usr
make -j`nproc` install
LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/../../usr/lib ctest .
```

src/mbedtls.c
```diff
@@ -0,0 +1,570 @@
...
+    {
+        ret = mbedtls_cipher_reset(ctx);
+
+        if(!ret)
+            ret = mbedtls_cipher_update(ctx, block, blocklen, output, &olen);
+
+        if(!ret)
+            ret = mbedtls_cipher_finish(ctx, output + olen, &finish_olen);
+
+        olen += finish_olen;
```
pjbakker: If `mbedtls_ciper_reset()` fails, `olen` and `finish_olen` will still be uninitialized. Adding them and using them in the `_libssh2_mbedtls_safe_free()` below will have unwanted behaviour.

src/mbedtls.c
```diff
@@ -0,0 +1,570 @@
...
+
+int
+_libssh2_mbedtls_hash(const unsigned char *data, unsigned long datalen,
+                      mbedtls_md_type_t mdtype, unsigned char *hash)
+{
+    const mbedtls_md_info_t *md_info;
+    int ret;
+
+    md_info = mbedtls_md_info_from_type(mdtype);
+    ret = mbedtls_md(md_info, data, datalen, hash);
```
pjbakker: Are you sure md_info will always be non-NULL?

src/mbedtls.c
```diff
@@ -0,0 +1,570 @@
...
+void
+_libssh2_mbedtls_bignum_free(_libssh2_bn *bn)
+{
+    if (bn)
+    {
+        mbedtls_mpi_free(bn);
+#ifdef LIBSSH2_CLEAR_MEMORY
+        memset(bn, 0, sizeof(_libssh2_bn));
+#endif
+    }
```
pjbakker: Calls like this to `memset()` might be optimized out by the compiler. Consider using something like the mbed zeroize functions (present in each module) or using `memset_s()` (depending on the C version you want to support)

pjbakker:
Did a quick pass over the PR.. As I'm not that familiar with the libssh2 internals, I cannot comment too much on correctness. I flagged things that seemed worth a second look.

src/mbedtls.c
wildart: if any of the `mbedtls_` calls fail then `memcpy` will not be executed with `olen` as parameter. 
BTW, `_libssh2_mbedtls_safe_free` is called with `osize` not `olen`.

src/mbedtls.c
wildart: ah, `mbedtls_mpi_free` already does it.

src/mbedtls.c
```diff
@@ -0,0 +1,603 @@
...
+{
+    unsigned char hash[SHA_DIGEST_LENGTH];
+    int ret;
+
+    ret = _libssh2_mbedtls_hash(m, m_len, MBEDTLS_MD_SHA1, hash);
+    if(ret)
+        return -1; /* failure */
+
+    ret = mbedtls_rsa_pkcs1_verify(rsa, NULL, NULL, MBEDTLS_RSA_PUBLIC,
+                                   MBEDTLS_MD_SHA1, sig_len, hash, sig);
```
Keno: The length parameter here is the length of the hash, not the length of the signature, which is assumed to be `rsa->len`. That assumption should be verified, and `SHA_DIGEST_LENGTH` should be passed here.

src/mbedtls.c
```diff
@@ -0,0 +1,603 @@
...
+#ifndef LIBSSH2_CLEAR_MEMORY
+    (void)len;
+#endif
+
+    if (!buf)
+        return;
+
+#ifdef LIBSSH2_CLEAR_MEMORY
+    if (len > 0)
+        memset(buf, 0, len);
```
Keno: Should this use SecureZeroMemory on windows?

wildart: windows binary compiled with its own cyptolib that uses SecureZeroMemory: https://github.com/libssh2/libssh2/blob/433c327ee2b7cdf83b38848b6e9473f93366ed6d/src/wincng.c#L310-L326

src/mbedtls.c
```diff
@@ -0,0 +1,603 @@
...
+    if (!bn || bits <= 0)
+        return -1;
+
+    len = (bits + 1) >> 3;
+    err = mbedtls_mpi_fill_random(bn, len, mbedtls_ctr_drbg_random, &_libssh2_mbedtls_ctr_drbg);
+    if (err)
+        return -1;
+
+    /* If `top` is -1, the most significant bit of the random number can be zero. */
+    err = mbedtls_mpi_set_bit(bn, 0, 0);
```
Keno: Are you sure the bits are arranged in this order? I would have expected 0 to be the least significant bit. Also, this is a `may` not a requirement, so I believe this call shouldn't be here. However, we should 0 out the the top bits if we created too many.

src/mbedtls.c
```diff
@@ -0,0 +1,603 @@
...
+_libssh2_mbedtls_bignum_random(_libssh2_bn *bn, int bits, int top, int bottom)
+{
+    size_t len;
+    int err;
+    int i;
+
+    if (!bn || bits <= 0)
+        return -1;
+
+    len = (bits + 1) >> 3;
```
Keno: I don't think this does what you want (round `bits` up to the nearest full byte).

src/mbedtls.c
Keno: It seems reasonable that somebody may want to compile the mbedtls backend on windows. Does libssh2 not have it's own function that zeros memory? In general you'd want to avoid memset since the compiler is free to remove it if it can prove the memory is not going to be accessed again.

Keno: Ok, looking at the rest of the library it just uses memset everywhere, so this seems fine though it may be a good idea to take a pass through the library globally and put in a non-optimizable memset for hardening purposes.

src/mbedtls.c
```diff
@@ -0,0 +1,603 @@
...
+    if( ret != 0 )
+    {
+        mbedtls_pk_free(&pkey);
+        mbedtls_rsa_free(*rsa);
+        LIBSSH2_FREE(session, *rsa);
+        *rsa = NULL;
+        return -1;
+    }
+
+    mbedtls_rsa_context *pk_rsa = mbedtls_pk_rsa(pkey);
```
Keno: Verify that this is actually an RSA context first?

src/mbedtls.c
wildart: `mbedtls` provides mechanism of redefining memory management functions, so `mbedtls_free` can be set to provide secure memory freeing if necessary.

src/mbedtls.c
```diff
@@ -397,7 +397,7 @@ _libssh2_mbedtls_rsa_sha1_verify(libssh2_rsa_ctx *rsa,
         return -1; /* failure */
 
     ret = mbedtls_rsa_pkcs1_verify(rsa, NULL, NULL, MBEDTLS_RSA_PUBLIC,
-                                   MBEDTLS_MD_SHA1, sig_len, hash, sig);
+                                   MBEDTLS_MD_SHA1, SHA_DIGEST_LENGTH, hash, sig);
```
Keno: Also need to verify that `siglen >= rsa->len`.

src/mbedtls.c
Keno: Yes, this is fine, I was more concerned about libssh2 in general, but that's a separate concern.

tkelman: worth a separate issue, if the maintainers share the concern

Keno: Filed as https://github.com/libssh2/libssh2/issues/120, so it doesn't get forgotten.

src/mbedtls.c
wildart: Why does it matter? SHA1 can be computed for any buffer length.

Keno: Not the buffer length, the signature length. If you try to verify an RSA-1024 signature with an RSA-2048 key, you get invalid memory reads otherwise.

src/mbedtls.c
```diff
@@ -0,0 +1,603 @@
...
+
+    osize = blocklen+mbedtls_cipher_get_block_size(ctx);
+
+    output = (unsigned char *)mbedtls_calloc(osize, sizeof(char));
+    if(output)
+    {
+        ret = mbedtls_cipher_reset(ctx);
+
+        if(!ret)
+            ret = mbedtls_cipher_update(ctx, block, blocklen, output, &olen);
```
wildart: well, it's upto caller of `_libssh2_cipher_crypt`

Keno: What I'm saying is you don't need to allocate an extra `output` buf. libssh2 should only ever need to pass full blocks.

src/mbedtls.c
```diff
@@ -208,34 +208,40 @@ _libssh2_mbedtls_bignum_random(_libssh2_bn *bn, int bits, int top, int bottom)
...
         return -1;
 
-    /* If `top` is -1, the most significant bit of the random number can be zero. */
-    err = mbedtls_mpi_set_bit(bn, 0, 0);
-    if (err)
-        return -1;
+    /* Zero unsued bits above the most significant bit and if `top` is -1,
+       the most significant bit of the random number can be zero. */
+    msb = bits;
+    if (top == -1)
```
Keno: I don't think this is correct. "Can be zero" just means that it's not forced to 1 as the other values of `top` do.

src/mbedtls.c
```diff
@@ -0,0 +1,606 @@
...
+            return -1;
+    }
+
+    /* If `top` is -1, the most significant bit of the random number can be zero.
+       If top is 0, the most significant bit of the random number is set to 1,
+       and if top is 1, the two most significant bits of the number will be set
+       to 1, so that the product of two such random numbers will always have 2*bits length.
+    */
+    for(i=0;i<=top;++i) {
+        err = mbedtls_mpi_set_bit(bn, bits-i, 1);
```
Keno: `bits - i - 1`? It seems to me that the most significant bit is `bits - 1`, no?

wildart: I thought `bits` is index of the most significant bit. because from `kex.c` `_libssh2_bn_rand` is called with following parameters
```
_libssh2_bn_rand(exchange_state->x, group_order * 8 - 1, 0, -1);
```
Why subtract 1,  other then adjusting to 0-based bit index.

Keno: Well, `bits` is definitely the number of bits of randomness, and we know bits are zero indexed, so say we want one bit of randomness, then `0` is the MSB which is `bits-1`, no?

src/mbedtls.c
wildart: It's necessary according to [mbedtls docs](https://github.com/ARMmbed/mbedtls/blob/98e28a74e33f32bcb855e16f8d5d2016b2102129/include/mbedtls/cipher.h#L608-L610).

Keno: Doesn't match the appropriate check in the code though: https://github.com/ARMmbed/mbedtls/blob/a54f0f53c59e3db884c7d31a299f8bb7b80356eb/library/cipher.c#L295-L299. @pjbakker Which is correct, the code or the comment?

src/mbedtls.c
wildart: Sure.

src/mbedtls.c
pjbakker: That check is not for all cipher modes. So practically, depending on the mode it might hold, but it's not guaranteed from an API perspective at this point.

Keno: Can you specify for which cipher modes it does not apply? As far as I can tell only ECB and GCM branch off before this and both of them also seem to allow `input == output`. Is there any theoretical reason that should ever not work?

tkelman:
@wildart is there remaining review to address here as far as you're concerned?

@Keno what are your thoughts? If needed, would you be able to take this over so we can finish upstreaming it?

Keno:
I believe @wildart has addressed all the review comments I had. This is good to go from my perspective.

tkelman:
Can any of the libssh2 maintainers review this please? Julia has been carrying this patch for a month or two and it's been working for our purposes. We want to get this PR upstreamed some day. What do you need us to do to get this moving in that direction, even slowly?

bagder:
Can you then please squash the patch properly? Reviewing a patch series with some of the patches fixing mistakes in some of the earlier patches isn't very effective. Squash those that fix mistakes so that you only have distinct commits that each makes a logical change.

tkelman:
@Keno do you have collaborator access to @wildart's fork to do that?

Can also look at https://github.com/libssh2/libssh2/pull/106/files which just shows the end result.

bagder:
If the end result is the only thing to review, then the PR should be made as a single commit.

Any considerations on fixing the autotools build script to support mbedTLS? Shouldn't we have some docs somewhere that describes this support and perhaps how you build libssh2 to use it?

tkelman:
Yes, docs are an obvious omission. @wildart has been mia for a while so we'll have to either find someone with commit access to the fork (@Keno? hopefully he'll respond, maybe during the week) or open a new revised copy of the pr.

mbedtls does not have an autotools build system itself. It has a simple makefile but I don't know if it's enough for our purposes.

Keno:
Yes, I should have access to wildart's fork. I'll squash this.

Keno:
Turns out I don't. Opening a new PR.

bagder:
continued in #132

PR #118, opened 2016-07-29
Make libssh2 work again on os400.

monnerat:
Hi,

The CVE-2016-0787 fixes broke libssh2 on os400.
This pull request makes it work again.
See comment of commit 0b482a378f23d82b4dca104373d341bb423e501a and HACKING.CRYPTO for more information.

I tested it successfully with all crypto backends except wincng (not tested).

Thanks for pulling,

Patrick

### DIFF
--- a/docs/HACKING.CRYPTO
+++ b/docs/HACKING.CRYPTO
@@ -305,10 +305,50 @@ TripleDES-CBC algorithm identifier initializer.
 #define with constant value of type _libssh2_cipher_type().
 
 
-5) Big numbers.
+5) Diffie-Hellman support.
+If the crypto-library supports opaque Diffie-Hellman computations, symbol
+`libssh2_dh_key_pair' should be #defined as described below andhe rest of this
+section applies.
+Else, the Diffie-Hellman context MUST be defined as `_libssh2_bn *' and
+the computation is emulated via calls to _libssh2_bn_rand() and
+_libssh2_bn_mod_exp(): all other symbols in this section are unused in this
+case.
+
+5.1) Diffie-Hellman context.
+_libssh2_dh_ctx
+Type of a Diffie-Hellman computation context.
+Must always be defined.
+
+5.2) Diffie-Hellman computation procedures.
+void libssh2_dh_init(_libssh2_dh_ctx *dhctx);
+Initializes the Diffie-Hellman context at `dhctx'. No effective context
+creation needed here.
+
+int libssh2_dh_key_pair(_libssh2_dh_ctx *dhctx, _libssh2_bn *public,
+                        _libssh2_bn *g, _libssh2_bn *p, int group_order,
+                        _libssh2_bn_ctx *bnctx);
+Generates a Diffie-Hellman key pair using base `g', prime `p' and the given
+`group_order'. Can use the give big number context `bnctx' if needed.
+The private key is stored as opaque in the Diffie-Hellman context `*dhctx' and
+the public key is returned in `public'.
+0 is returned upon success, else -1.
+If defined, this procedure MUST be implemented as a #define'd macro.
+
+int libssh2_dh_secret(_libssh2_dh_ctx *dhctx, _libssh2_bn *secret,
+                      _libssh2_bn *f, _libssh2_bn *p, _libssh2_bn_ctx * bnctx)
+Computes the Diffie-Hellman secret from the previouly created context `*dhctx',
+the public key `f' from the other party and the same prime `p' used at
+context creation. The result is stored in `secret'.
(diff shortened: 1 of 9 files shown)


docs/HACKING.CRYPTO
```diff
@@ -305,10 +305,50 @@ TripleDES-CBC algorithm identifier initializer.
 #define with constant value of type _libssh2_cipher_type().
 
 
-5) Big numbers.
+5) Diffie-Hellman support.
+If the crypto-library supports opaque Diffie-Hellman computations, symbol
+`libssh2_dh_key_pair' should be #defined as described below andhe rest of this
```
alamaison: Nit: typo

docs/HACKING.CRYPTO
```diff
@@ -305,10 +305,50 @@ TripleDES-CBC algorithm identifier initializer.
...
+5.2) Diffie-Hellman computation procedures.
+void libssh2_dh_init(_libssh2_dh_ctx *dhctx);
+Initializes the Diffie-Hellman context at `dhctx'. No effective context
+creation needed here.
+
+int libssh2_dh_key_pair(_libssh2_dh_ctx *dhctx, _libssh2_bn *public,
+                        _libssh2_bn *g, _libssh2_bn *p, int group_order,
+                        _libssh2_bn_ctx *bnctx);
+Generates a Diffie-Hellman key pair using base `g', prime `p' and the given
+`group_order'. Can use the give big number context `bnctx' if needed.
```
alamaison: give = given?

src/kex.c
```diff
@@ -98,6 +98,44 @@
             }                                                              \
     }
 
+/*
+ * Generic Diffie-Hellman computation support.
+ *
+ * DH context should be a _libssh2_bn *.
+ */
+
+#ifndef libssh2_dh_key_pair
```
alamaison: It's confusing using `libssh2_dh_key_pair` both as a function definition and as a `#define`d symbol.  The two things exist in different namespaces.  For example `libssh2_dh_key_pair` defined on line 113 a won't make `#ifndef libssh2_dh_key_pair` true.  But at first glance that looks like what you're trying to do.  I would prefer a separate symbol (something like `#ifndef LIBSSH2_DH_IMPLEMENTATION`).

Though, really, I think the correct solution is to break out these functions into two separate files, where the only one is chosen at config time (like we do for the crypto-backend implementations).  One file would be for for 0S400, the other for everything else.  They would then simply implement the same set of 4 functions, no preprocessor needed.

monnerat: Thanks for the review.

Would'nt it be then preferable to implement both phases of Diffie-Hellman computation + constructor/destructor as crypto backend specific procedures ? This would also release the need to "publish" _libssh2_bn_mod_exp() and _libssh2_bn_rand().

If you agree, I can produce another pull request for it in a few days...

alamaison: > Would'nt it be then preferable to implement both phases of Diffie-Hellman computation + constructor/destructor as crypto backend specific procedures

Yes, that makes sense to me.  Sorry for the long delay in responding

monnerat: Hi Alexander,
In the meantime, I've lost an access to an iSeries, thus os400. I don't know if I'll regain such an access in the future.
This means I can implement what I suggested on Aug 19, but the os400 implementation will be "blind": I can't test it anymore.
Go or NoGo? Your opinion matters.
Thanks for your reply.

alamaison: I would rather merge your current tested code, even if it's not perfect.  We can always clean it up in the future should you regain access to OS400.

Could you fix the typos and then i'll merge it as it is.

monnerat:
Travis failed: probably not a libssh2 problem, but a travis one.
How to resubmit without a commit ?

alamaison:
> Most of the rework involved by my Aug 19 comment concerns non-OS400 backends, so I can test Linux-based backends afterwards.

So would you like me to merge this PR now and you'll rework in a new one?

> How to resubmit without a commit

I've rerun the failed job.

monnerat:
Yes please: I'm sure the OS400 code is OK. I'll rework the other backends in a new PR.
Thanks a lot.

alamaison:
Done.  Thanks for the contribution

monnerat:
Thanks to you for your help :-)

alamaison:
@monerat The change broke the mbedTLS backend: https://travis-ci.org/libssh2/libssh2/jobs/175349805  Could you fix up in a new PR?

monnerat:
Sorry for that.
New PR at https://github.com/libssh2/libssh2/pull/146

PR #119, opened 2016-08-03
Proper error checking for bn_rand/bn_mod_exp

Keno:
On most backends these functions can fail, but they weren't checked, so the key exchange could proceed with invalid bignums. Since this is on the error path, it shouldn't affect security, but it would be good to have this for proper error reporting. Since I didn't think any of the existing error categories was applicable, I created a new one. There are likely other bn functions in the same situation, but before doing a massive pull request, I wanted to start with this to see if it was acceptable, and figure out if this is the proper way to handle this.

### DIFF
--- a/docs/HACKING.CRYPTO
+++ b/docs/HACKING.CRYPTO
@@ -362,19 +362,19 @@ Converts the absolute value of bn into big-endian form and store it at
 val. val must point to _libssh2_bn_bytes(bn) bytes of memory.
 Returns the length of the big-endian number.
 
-void _libssh2_bn_rand(_libssh2_bn *bn, int bits, int top, int bottom);
+int _libssh2_bn_rand(_libssh2_bn *bn, int bits, int top, int bottom);
 Generates a cryptographically strong pseudo-random number of bits in
 length and stores it in bn. If top is -1, the most significant bit of the
 random number can be zero. If top is 0, it is set to 1, and if top is 1, the
 two most significant bits of the number will be set to 1, so that the product
 of two such random numbers will always have 2*bits length. If bottom is true,
-the number will be odd.
+the number will be odd. Returns 1 on success, 0 otherwise.
 
-void _libssh2_bn_mod_exp(_libssh2_bn *r, _libssh2_bn *a,
+int _libssh2_bn_mod_exp(_libssh2_bn *r, _libssh2_bn *a,
 	                 _libssh2_bn *p, _libssh2_bn *m,
 	                 _libssh2_bn_ctx *ctx);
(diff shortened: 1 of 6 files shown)


docs/HACKING.CRYPTO
```diff
@@ -362,19 +362,19 @@ Converts the absolute value of bn into big-endian form and store it at
...
 
-void _libssh2_bn_rand(_libssh2_bn *bn, int bits, int top, int bottom);
+int _libssh2_bn_rand(_libssh2_bn *bn, int bits, int top, int bottom);
 Generates a cryptographically strong pseudo-random number of bits in
 length and stores it in bn. If top is -1, the most significant bit of the
 random number can be zero. If top is 0, it is set to 1, and if top is 1, the
 two most significant bits of the number will be set to 1, so that the product
 of two such random numbers will always have 2*bits length. If bottom is true,
-the number will be odd.
+the number will be odd. Returns 1 on success, 0 otherwise.
```
mback2k: As far as I can tell most internal backend functions use 0 on success and -1 otherwise/on error. Please make sure that your new return values match the existing scheme.

Keno: While that's true, there is a precent of other `_libssh2_bn` functions that use this scheme. I have therefore taken the opposite approach of adjusting the internal APIs, since before these were specified as returning void. Happy to adjust everything the other, but I would like it to be consistent.

alamaison: If it's consistent, I'm happy.  @mback2k: any objection to me merging this?

mback2k: I don't think it is at the moment, because at least most backend related functions use 0 on success and -1 on error. In the WinCNG backend there is not a single function that uses 1 on success and 0 on error.

alamaison: Understood.  @Keno could you invert the return codes and rebase your changes onto master?

mback2k: I just took a look at the OpenSSL and libgcrypt backends. OpenSSL uses 0 on success and -1 **or** 1 on error for non-static functions. I think this needs to be unified at some point. At least the libgcrypt backend is consistent in using 0 on success and -1 on error, just like WinCNG (probably because I based the WinCNG code on the libgcrypt backend).

mback2k:
@Keno @alamaison Do you think this change is still relevant since f7daf3185a4e889fcd5272858db97764a279da1e has been merged? I think yes, but now it needs to be reworked to fix _libssh2_dh_secret and _libssh2_dh_secret instead.

Issue #17, opened 2015-05-21
can't seem to link to nonstandard openssl


kotoroshinoto:
no matter what I do, it ends up linking to the system files:
libssl.so.10 => /usr/lib64/libssl.so.10 (0x00007f83a6248000)
libcrypto.so.10 => /usr/lib64/libcrypto.so.10 (0x00007f83a5e68000)

it should be linking to the ones in 
ls /UCHC/HPC/Gooch/biotoolmodules/openssl/1.0.2a/lib/
engines  libcrypto.a  libcrypto.so  libcrypto.so.1.0.0  libssl.a  libssl.so  libssl.so.1.0.0  pkgconfig

what do I need to tell the configure script to accomplish this?

bagder:
Is that what ldd shows? If so, you need to make sure it picks the right libs at run-time, so it depends on your /etc/ld.so.conf setup. And you can also specify dir with `LD_LIBRARY_PATH`

kotoroshinoto:
it wasn't a runtime link problem because that soname was the system version of openssl which was not binary compatible with the one I wanted it to use. When I added an additional line forcing it to give a directory path to the linker it then was able to find the correct library.

I'm still not quite sure why this was happening nor why I had to go out of my way to be explicit. I already HAD the runtime AND compile time environmental variables set, and it was defaulting to the system path library anyway.

foldedspace:
It looks like "configure" is not honouring the "--with-libssl-prefix" argument:

     ./configure --prefix="/opt/libssh2-1.6.0" --with-openssl --with-libssl-prefix=/opt/openssl-1.0.2d

Gives me:
     ....
      configure: error: No crypto library found!
      Try --with-libssl-prefix=PATH
      or --with-libgcrypt-prefix=PATH
      or --with-wincng on Windows

Despite:

     $ ls /opt/openssl-1.0.2d/{bin,lib}
     /opt/openssl-1.0.2d/bin:
     c_rehash*  openssl*

     /opt/openssl-1.0.2d/lib:
     engines/  lib64@  libcrypto.a  libcrypto.so@  libcrypto.so.1.0.0*  libssl.a  libssl.so@    libssl.so.1.0.0*  pkgconfig/

However, the following works:

[...]

PR #121, opened 2016-08-28
Added support for OpenSSL_1_1_0

tholberton:
Added getters and setters for opaque types `DSA` `DSA_SIG` and `RSA`.


bagder:
It might fix #17

vszakats:
Can this patch be extended/updated to enable the new `HAVE_` options inside the source whenever OpenSSL 1.1.0 is detected? This would make this build-system agnostic.

Adding these lines to `src/openssl.h` could work:
```c
#if OPENSSL_VERSION_NUMBER >= 0x10100000L
# define HAVE_DSA_GET0_KEY
# define HAVE_DSA_SET0_PQG
# define HAVE_DSA_SET0_KEY
# define HAVE_DSA_SIG_GET0
# define HAVE_DSA_SIG_SET0
# define HAVE_RSA_GET0_KEY
# define HAVE_RSA_SET0_KEY
# define HAVE_RSA_SET0_FACTORS
# define HAVE_RSA_SET0_CRT_PARAMS
#endif
```

tholberton:
@vszakats I could do that. But, it would make more sense to me to put it into `src/openssl.c` like this:

```
#ifdef OPENSSL_VERSION_MAJOR >= 0x101
```

vszakats:
@tholberton Even better that way.

vszakats:
I missed `HAVE_DSA_GET0_PQG` from the top of the list of my snippet. Can you add it as well?

```c
# ifndef HAVE_DSA_GET0_PQG
#  define HAVE_DSA_GET0_PQG 1
# endif
```

vszakats:
With the `HAVE_DSA_GET0_PQG` modification, libssh2 builds fine against OpenSSL 1.1.0 using this patch:
https://ci.appveyor.com/project/vsz/harbour-deps/build/1.0.890#L2501

Basic SSH/SFTP test also worked fine.

tholberton:
I was actually thinking that it would be better to do something like this:
```
#if OPENSSL_VERSION_NUMBER < 0x10100000L
static void DSA_get0_pqg(const DSA * dsa, const BIGNUM ** p, const BIGNUM ** q, const BIGNUM ** g){
    if (p != NULL){
        *p = dsa->p;
    }
    /* ... */
}
/* ... */
#endif
```
And remove the functions checks in the build system and the `ifdef` sections in the code that I original wrote. It would look a lot cleaner and completely avoid build system code. I can write it with a minimal diff and you can just squash the merge to get rid of the work in between.

vszakats:
Perhaps it'd be enough to use the existing `HAVE_OPAQUE_STRUCTS` macro to replace the multiple `HAVE_*` ones, while keeping the current source modifications (which I think it's nice and pure, and would be somewhat obfuscated by using new static functions). Since this macro is already defined, all special cmake logic can be dropped, along with their definitions on top of `openssl.c`.

willco007:
I agree with @vszakats; having defines for every function seems unnecessary as they are OpenSSL specific. I would piggy back off the HAVE_OPAQUE_STRUCTS defined in the the header; perhaps renaming it to HAVE_OPAQUE_OPENSSL or something similar.

vszakats:
`HAVE_OPAQUE_OPENSSL` is slightly misleading as `HAVE_OPAQUE_STRUCTS` already applies to OpenSSL, plus the former may apply to BoringSSL/LibreSSL as well in the future. For these reasons, I feel the name `HAVE_OPAQUE_RSADSA` (or similar) would better tell its purpose. It's just a minor concern though.

Test builds against OpenSSL 1.1.0 are successful:
https://ci.appveyor.com/project/vsz/harbour-deps/build/1.0.897

willco007:
There is a typo in this commit; line 188 should be:

DSA_set0_pqg(*dsactx, p_bn, q_bn, g_bn);

willco007:
@vszakats My thought was we could add HAVE_OPAQUE_LIBRE, etc if they applied these API changes as well. However, that may get cumbersome down the road. Perhaps a wait and see approach is best for now?

vszakats:
@willco007 Yes, having to maintain different (and possibly colliding) fork-specific feature-sets can easily become a problem. I'd expect the forks to keep it sane and either not endorse certain OpenSSL updates or apply them in a compatible fashion. (We shall see)

Taking a peek at sister project `libcurl`, there the macro name chosen for the purpose at hand was `HAVE_OPAQUE_RSA_DSA_DH`, so I suggest to use the same name here.

vszakats:
On a second (third?) look, the correct macro definition would be:
```
#if OPENSSL_VERSION_NUMBER >= 0x10100000L && \
    !defined(LIBRESSL_VERSION_NUMBER)
# define HAVE_OPAQUE_RSA_DSA_SH 1
#endif
```
This can be merged with the block defining `HAVE_OPAQUE_STRUCTS` in `openssl.h`.

willco007:
My thought about that this name is if Libre or Boring don't implement the exact same API we're back to multiple defines for opaque structures.  Having it named something OpenSSL specific would be a hint to the user these calls are OpenSSL specific which may eliminate confusion in the future.

vszakats:
To get around the naming issue altogether, existing `HAVE_OPAQUE_STRUCTS` macro can be used in this patch. It will make these modifications work with all existing LibreSSL and OpenSSL versions.

tholberton:
My thought on the macro name is that `HAVE_OPAQUE_STRUCTS` is exceeding broad. The names `HAVE_OPAQUE_DSA` and `HAVE_OPAQUE_RSA` make a lot of sense to me. The name `HAVE_OPAQUE_OPENSSL` seems like a decent compromise.

vszakats:
Merged, thank you @tholberton!

PR #122, opened 2016-08-29
Non-blocking examples don't handle all LIBSSH2_ERROR_EAGAIN

Zenju:
This is just an example bug-fix, but unless I'm missing something, all the libssh2 non-blocking examples fail to handle "LIBSSH2_ERROR_EAGAIN" when calling clean-up functions like "libssh2_sftp_close, libssh2_sftp_shutdown, libssh2_session_disconnect"

### DIFF
--- a/example/sftp_nonblock.c
+++ b/example/sftp_nonblock.c
@@ -270,7 +270,8 @@ int main(int argc, char *argv[])
 #endif
 
     libssh2_sftp_close(sftp_handle);
-    libssh2_sftp_shutdown(sftp_session);
+    while (libssh2_sftp_shutdown(sftp_session) == LIBSSH2_ERROR_EAGAIN)
+           ;
 
 shutdown:
 


bagder:
Correct, they do contain that omission all over. And just doing a busy-look for the EAGAIN state is probably only a little bit better since it'll then make people adopt that pattern...

Zenju:
Okay, so no surprise there. What I'm currently missing the most with the non-blocking samples is some code showing real interleaving, e.g.:

`	
	LIBSSH2_SFTP sftpSession1 = ::libssh2_sftp_init(sshSession);
	LIBSSH2_SFTP sftpSession2 = ::libssh2_sftp_init(sshSession);

	libssh2_session_set_blocking(sshSession, 0);

	LIBSSH2_SFTP_HANDLE* dh1 = nullptr;
	LIBSSH2_SFTP_HANDLE* dh2 = nullptr;
	do
	{
		if (!dh1)
		{
			dh1 = ::libssh2_sftp_opendir(sftpSession1, "/folder1")
			if (!dh1 && ::libssh2_session_last_errno(sshSession) != LIBSSH2_ERROR_EAGAIN)
				... handle error				
		}
		if (!dh2)
		{
			dh2 = ::libssh2_sftp_opendir(sftpSession2, "/folder2")
			if (!dh2 && ::libssh2_session_last_errno(sshSession) != LIBSSH2_ERROR_EAGAIN)
				... handle error				
		}
	}
	while (!dh1 || !dh2)
`

Reading source code comments, it seems this should be supported and working as expected with the sole precondition that each libssh2 call returning "LIBSSH2_ERROR_EAGAIN" must be finished before issueing a different one. Is this not an officially supported scenario? (it *appears* to be working correctly, though, and improves perf by 2x with the above code).

bagder:
That seems like totally irrelevant to this pull request though, but yes libssh2 supports multiple channels used simultaneously.

PR #123, opened 2016-08-30
libssh2_sftp_init hang: last error not set

Zenju:
The problem is that the original if statement simply returns NULL, but does not set the session last error code. The consequence is that libssh2_sftp_init() also returns NULL and libssh2_session_last_errno(sshSession) == LIBSSH2_ERROR_NONE.

In my test the LIBSSH2_ERROR_EAGAIN is coming from sftp.c row 337:
            if(4 != sftp->partial_size_len)
                /\* we got a short read for the length part */
                return LIBSSH2_ERROR_EAGAIN;

with "partial_size_len == 0". Not sure if this is expected.

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -836,8 +836,11 @@ static LIBSSH2_SFTP *sftp_init(LIBSSH2_SESSION *session)
 
     rc = sftp_packet_require(sftp_handle, SSH_FXP_VERSION,
                              0, &data, &data_len);
-    if (rc == LIBSSH2_ERROR_EAGAIN)
-        return NULL;
+    if (rc == LIBSSH2_ERROR_EAGAIN) {
+         _libssh2_error(session, LIBSSH2_ERROR_EAGAIN,
+                        "Would block receiving SSH_FXP_VERSION");
+         return NULL;
+    }
     else if (rc) {
         _libssh2_error(session, rc,
                        "Timeout waiting for response from SFTP subsystem");


Zenju:
> with "partial_size_len == 0". Not sure if this is expected.

It seems the session is corrupted from here on and will always return "rc = _libssh2_channel_read" with rc == 0. Maybe this should be considered an error state instead?

Zenju:
Indeed "_libssh2_channel_read" return 0 is an error in my case due to heavy load or server limits: e.g. I'm testing with 10 parallel ssh sessions, and when each session has about 8 SFTP sessions, i.e. about 80 SFTP sessions total, the next libssh2_sftp_init() will hang and "_libssh2_channel_read" will return 0 as seen above.
The same happens for 15 parallel ssh sessions when reaching about 5 SFTP sessions each. Obviously the server seems to just not respond when the total SFTP session count reaches ~80. When I try to connect to the server with a different SFTP tool at this time, it will fail immediately with "Premature EOF", although it's a completely different process.
Is this the server misbehaving, or should libssh2 somehow detect this situation and report a proper error instead? My test case tries to create as many SFTP sessions as possible for maximum parallel performance, but loses all of it once libssh2_sftp_init() starts to hang (I'd prefer to get a simple error code instead of LIBSSH2_ERROR_TIMEOUT).

alamaison:
I've merged you PR because the change looks legit.  But I didn't understand from you comments if this actually fixed you issues.  If you're having other problems, you should probably create an issue for them

PR #131, opened 2016-09-23
Fix type warnings when building on macOS

willco007:


alamaison:
This change doesn't fix the warnings, it silences them.  There are legitimate issues passing size_t values to the libssh2 APIs currently, but we can't change the ABI to properly fix them without forcing an so-bump.  I'm planning to add a flag that lets users opt in to a new incompatible ABI, but I don't have an ETA

willco007:
While this change is technically incorrect; there is no harm in casting the values until a ABI breaking change is introduced.

PR #132, opened 2016-09-26
Rebased/Squashed mbedTLS branch

Keno:
This rebases and squashes #106. I've also added support for mbedtls to the autotools build system and added some documentation on the new build system options.


bagder:
`automake: error: cannot open < src/../Makefile.mbedTLS.inc: No such file or directory`

Forgot to add that file I guess?

Keno:
yes, sorry!

Keno:
should be fixed.

bagder:
Yes, works now. Now I just ran into problems with the configure script. Not really your fault as you just used the same style already used, its just that it isn't very good...

Keno:
I suppose I'm gonna have to build mbedtls from source on travis. I had hoped that wasn't necessary anymore since we don't need to patch it now, but it looks like Travis' ubuntu is too old to have the package and I couldn't find a good ppa.

bagder:
Thanks for all the work. I merged what we have now. For follow up fixes and more, just create new issues or PRs, thanks!

tkelman:
Fantastic, thank you so much @bagder!

bagder:
Ah right, we need to fix the travis build. Do we revert the commit 3e6de50 to start with or do you have a fix pending?

tkelman:
I'll try a PR

PR #133, opened 2016-09-27
Build mbedtls from source on Travis

tkelman:
Similar to https://github.com/libssh2/libssh2/pull/132/commits/58f105617f5a4200eca7d044bc97c26ff16b61ab but doing more of the work inline in `.travis.yml`, ref #132


tkelman:
On my fork there's a test failure on the mbedtls entries
```
8/11 Test  #8: test_public_key_auth_succeeds_with_correct_dsa_key ..............***Failed    0.81 sec
libssh2_userauth_publickey_fromfile_ex failed (-19): Callback returned error
Command: docker build -t libssh2/openssh_server openssh_server
Command: docker run --detach -P libssh2/openssh_server
Command: docker inspect --format "{{ index (index (index .NetworkSettings.Ports \"22/tcp\") 0) \"HostIp\" }}" f27caa69784948775bcc4612c43e203dafb77305cf2fd06c332167c27782c835
Command: docker inspect --format "{{ index (index (index .NetworkSettings.Ports \"22/tcp\") 0) \"HostPort\" }}" f27caa69784948775bcc4612c43e203dafb77305cf2fd06c332167c27782c835
Command: docker stop f27caa69784948775bcc4612c43e203dafb77305cf2fd06c332167c27782c835
```

but that was also happening in #132. Not sure what's wrong. We could maybe mark the mbedtls builds as allowed-failures for now?

Keno:
I suspect that's because the mbedtls backend only has RSA support implemented. You can grep for MBEDTLS_PK_RSA to see where this assumption is made.

tkelman:
Is that a problem, or should the test just be skipped for mbedtls?

alamaison:
> Is that a problem, or should the test just be skipped for mbedtls?

It's not a problem of that's all it supports.  The tests assume all backends support RSA and DSA.  I think it's reasonable to skip this test for mbedtls.  Do you have access to `CRYPTO_BACKEND` in `tests/CMakeLists.txt`?  If so, you could use that to exclude the DSA test.

tkelman:
Looks like that's working so far, thanks

alamaison:
> guess I need to fix the mbedtls build to use the same toolchain as the libssh build

The tests are passing.  Did you fix what you needed to?

tkelman:
yep, just had to pass the toolchain flag to the cmake call for mbedtls

PR #145, opened 2016-11-09
Fix transport read returning zero

salva:
Fix `_libssh2_transport_read returning` zero when the socket is disconnected.

These patches fix the issues discussed in #104.

This is probably not ready for merging yet.


salva:
The documentation before _libssh2_transport_read needs to be fixed too.

src/packet.c
```diff
@@ -1233,35 +1220,29 @@ _libssh2_packet_requirev(LIBSSH2_SESSION *session,
...
-            }
+            return ret;
+        }
+        else if (ret < 0) {
+            state->start = 0;
+            return ret;
         }
 
-        if (strchr((char *) packet_types, ret)) {
+        if (ret && strchr((char *) packet_types, ret)) {
```
alamaison: Maybe add a comment here explaining that this will skip any packets with the invalid type 0

src/packet.c
```diff
@@ -1127,16 +1127,6 @@ _libssh2_packet_require(LIBSSH2_SESSION * session, unsigned char packet_type,
                                       match_ofs, match_buf, match_len);
             state->start = 0;
             return ret;
-        } else if (ret == 0) {
```
alamaison: I agree that this code was pointless because you could only get if if the socket was disconnected, in which case who cares about the timeout.  But the timeout should not be applied to the LIBSSH2_ERROR_EAGAIN case, as done in _packet_requirev.

As you say though, it should be possible to implement _packet_require in terms of _packet_requirev.

src/packet.c
```diff
@@ -1233,35 +1220,29 @@ _libssh2_packet_requirev(LIBSSH2_SESSION *session,
...
             /* Be lazy, let packet_ask pull it out of the brigade */
             return _libssh2_packet_askv(session, packet_types, data,
                                         data_len, match_ofs, match_buf,
                                         match_len);
         }
     }
-
-    /* Only reached if the socket died */
-    state->start = 0;
-    return LIBSSH2_ERROR_SOCKET_DISCONNECT;
```
alamaison: Can you double-check that removing this doesn't generate a compiler warning?

salva:
> Maybe add a comment here explaining that this will skip any packets with the invalid type 0

The check was there because `strchr` would match the zero at the end of the `packet_types` string. That was quite ugly and conceptually wrong as this is not the layer to handle packets of type 0 in a special manner,  so, instead of adding a comment I have change the function to get the `packet_types` vector length as an argument. 0 is not a special value anymore.

Also, I have removed the timeout handling code from `_libssh2_packet_requirev` as it was broken, using a hard code value and done at the wrong layer. Timeouts are already handled globally by `_libssh2_wait_socket`.

After that, I have been able to refactor several functions, removing quite a good amount of duplicated code.

salva:
Finally, those are the changes remaining in #104 (which I am going to close now).

salva:
For me, the more dubious changes are those related to the timeout handling code being removed.

I think it was useless code as timeouts are already handled in the `libssh2_wait_socket`, it used a fixed timeout and was not included in all the functions called from the upper layer. Also, removing it greatly simplifies the code and allows to eliminate several state variables from the session structure.

But I may be overlooking something.

jonhoo:
I think this was probably closed prematurely. For example,  #84 was closed pending this fix, but with the fix closed, that issue should probably be re-opened. Or the fix merged I suppose.

willco007:
This needs to be rebased and is rather old at the moment, but I'll reopen it in hopes someone fixes it up.

jonhoo:
It is, also, I believe, the resolution to https://github.com/libssh2/libssh2/pull/104, unless that has since been fixed elsewhere?

PR #148, opened 2016-11-16
Check for netinet/in.h in the tests CMake file

ararslan:
In tests/openssh_fixture.c, the system header netinet/in.h is included if it was determined to exist. However, no such check for existence was actually being performed, so `#ifdef HAVE_NETINET_IN_H` is always false in the tests. This was causing problems on FreeBSD, since it didn't know about `sockaddr_in` without that header.


ararslan:
Looks like the tests passed everywhere but one of the Travis builds stalled.

PR #149, opened 2016-11-17
Implement Diffie-Hellman computations in crypto backends.

monnerat:
This implements what has been suggested in https://github.com/libssh2/libssh2/pull/118 on Aug, 19.

Not all backends feature the low level API needed to compute a Diffie-Hellman
secret, but some of them directly implement Diffie-Hellman support with opaque
private data. The later approach is now generalized and backends are
responsible for all Diffie Hellman computations.
As a side effect, procedures/macros _libssh2_bn_rand and _libssh2_bn_mod_exp
are no longer needed outside the backends.

Tested locally with openssl, libgcrypt and mbedtls successfully. os400qc3 unaltered (was already compatible).
wincng carefully changed, but untested.

### DIFF
--- a/docs/HACKING.CRYPTO
+++ b/docs/HACKING.CRYPTO
@@ -333,13 +333,6 @@ TripleDES-CBC algorithm identifier initializer.
 
 
 5) Diffie-Hellman support.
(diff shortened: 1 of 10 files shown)


monnerat:
Thanks for merge.

Issue #130, opened 2016-09-14
Feature request: CMake setting to allow for libssh2.dll static VC++ runtime on Windows


jpluimers:
Apparently my `CMake` foo is way below par as the below does not work so I'm missing something in https://cmake.org/Wiki/CMake_FAQ#How_can_I_build_my_MSVC_application_with_a_static_runtime.3F

Aim is to get rid of the `VCRUNTIME140.DLL` dependency by switching from `vcruntime.lib` (dynamic linking) to `libvcruntime.lib` (static linking of the VC++ runtime).

```
    git clone https://github.com/libssh2/libssh2.git
    pushd libssh2

    :: Try to force Static libraries (so VCRUNTIME140.DLL is not needed; hopefully this links libvcruntime.lib instead of vcruntime.lib)
    pushd cmake
    > c_flag_overrides.cmake type NUL
    >> c_flag_overrides.cmake echo    if(MSVC)
    >> c_flag_overrides.cmake echo        set(CMAKE_C_FLAGS_DEBUG_INIT "/D_DEBUG /MTd /Zi /Ob0 /Od /RTC1")
    >> c_flag_overrides.cmake echo        set(CMAKE_C_FLAGS_MINSIZEREL_INIT     "/MT /O1 /Ob1 /D NDEBUG")
    >> c_flag_overrides.cmake echo        set(CMAKE_C_FLAGS_RELEASE_INIT        "/MT /O2 /Ob2 /D NDEBUG")
    >> c_flag_overrides.cmake echo        set(CMAKE_C_FLAGS_RELWITHDEBINFO_INIT "/MT /Zi /O2 /Ob1 /D NDEBUG")
    >> c_flag_overrides.cmake echo    endif()

    > cxx_flag_overrides.cmake type NUL
    >> cxx_flag_overrides.cmake echo    if(MSVC)
    >> cxx_flag_overrides.cmake echo        set(CMAKE_CXX_FLAGS_DEBUG_INIT "/D_DEBUG /MTd /Zi /Ob0 /Od /RTC1")
    >> cxx_flag_overrides.cmake echo        set(CMAKE_CXX_FLAGS_MINSIZEREL_INIT     "/MT /O1 /Ob1 /D NDEBUG")
    >> cxx_flag_overrides.cmake echo        set(CMAKE_CXX_FLAGS_RELEASE_INIT        "/MT /O2 /Ob2 /D NDEBUG")
[...]

PR #156, opened 2016-12-06
Add changes to CFLAGS for BUILD_SHARED_LIBS equal OFF

kadorken:
Finally (I think) figured out pull requests on github. Hopefully this change is clear as to its intent to allow static linking of libssh2 with static linked openssl.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ else()
 endif()
 
 option(BUILD_SHARED_LIBS "Build Shared Libraries" OFF)
+option(LIBSSH2_WIN32_USE_STATIC_CRT "Change CFLAGS to use static linking" OFF)
 
 # Parse version
 
--- a/docs/INSTALL_CMAKE
+++ b/docs/INSTALL_CMAKE
@@ -45,6 +45,11 @@ The following options are available:
     Determines whether libssh2 is built as a static library or as a
     shared library (.dll/.so).  Can be `ON` or `OFF`.
 
+ * `LIBSSH2_WIN32_USE_STATIC_CRT=ON`
+
+	Determines whether the WIN32 CFLAGS for linking is changed from /Md (dll) to /Mt (static).
+	Default is `OFF`
+
  * `CRYPTO_BACKEND=`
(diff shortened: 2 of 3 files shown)


kadorken:
This is for issue #130

alamaison:
The change in this PR causes the compiler on windows to link statically against the CRT if BUILD_SHARED_LIBS is OFF. This doesn't make libssh2 link statically against OpenSSL. Or did you just mean that this is a necessary prerequisite for that to be possible?

A couple of comments:

- linking statically against the CRT should not be tied to building libssh2 as a static library. It's valid and common to build libssh2 static but still use the CRT via DLL. Instead we should introduce a new flag, LIBSSH2_WIN32_USE_STATIC_CRT, that enables the functionality you've added.

- I'm surprised such a flag doesn't already exist at the CMake level. Messing with compiler flags manually is generally a last resort in CMake. Are you sure we can't do this without?

kadorken:
In my usage, I was intending openssl to be built for static linking in parallel with libssh2. I was under the impression that the Cmake 'find' procedure for openssl looks for the .LIB files for static, or the .DLL for non-static. I had problems with Cmake  finding OpenSSL libraries under strange places (like Strawberry Perl !) other than the one I was trying to build statically).

I looked for a flag in CMake to control the linking, but I always ended up at [https://cmake.org/Wiki/CMake_FAQ#How_can_I_build_my_MSVC_application_with_a_static_runtime.3F](https://cmake.org/Wiki/CMake_FAQ#How_can_I_build_my_MSVC_application_with_a_static_runtime.3F)

I don't object to having a separate flag for the linking change; I didn't want to introduce such a flag since I'm not familiar with the overall direction of libssh2. It is a simple change

kadorken:
I think I managed to make the change and added to the Pull request the separate option LIBSSH2_WIN32_USE_STATIC_CRT.

Added it to the docs/INSTALL_CMAKE file as well.

simon-p-r:
Can this PR please be merged?

illera88:
Eooooooo can this PR get merged?

Issue #141, opened 2016-10-26
Patch to fix buffer overflow


doublex:
``` diff
*** userauth.c-old      2016-10-25 08:43:39.000000000 +0200
--- userauth.c  2016-10-26 12:12:22.921837935 +0200
*************** memory_read_publickey(LIBSSH2_SESSION *
*** 487,493 ****

      sp1++;

!     if ((sp2 = memchr(sp1, ' ', pubkey_len - (sp1 - pubkey - 1))) == NULL) {
          /* Assume that the id string is missing, but that it's okay */
          sp2 = pubkey + pubkey_len;
      }
--- 487,493 ----

      sp1++;

!     if ((sp2 = memchr(sp1, ' ', pubkey_len - (sp1 - pubkey + 1))) == NULL) {
          /* Assume that the id string is missing, but that it's okay */
          sp2 = pubkey + pubkey_len;
      }
```

doublex:
Maybe it would be better to write:
````
!     if ((sp2 = memchr(sp1, ' ', pubkey_len - (sp1 - pubkey))) == NULL) {
````

alamaison:
Could you open a Pull Request with this patch?  It makes it hard to review like this and even harder to merge

willco007:
fixed with #159

PR #159, opened 2016-12-13
buffer overflow (valgrind)

doublex:
- (-1) -> one more byte - not one byte less


willco007:
This patch is for issue #141.

doublex:
The problem is this line:
https://github.com/libssh2/libssh2/blob/master/src/userauth.c#L494

The problem is the "double minus", which is plus

willco007:
Yup, there was a conflict with the existing pull request so I updated your pull request and am waiting for the build to succeed before merging.

PR #160, opened 2016-12-13
segfault, variable mismatch

doublex:
Wrong variable - if "signv()" fails


doublex:
@willco007 
Could you also merge this PR?

willco007:
Not sure why this didn't land when it was approved.  I will make the fix since the pull request is now out of date. Thanks for bringing this back up.

PR #163, opened 2016-12-20
Fix a segfault using ssh-agent on Windows

### LINKED ISSUE #162: Segfault using ssh-agent on Windows
Originally reported https://github.com/rust-lang/cargo/issues/3401 it looks like the segfault in that issue comes from libssh2. Unfortunately there's a lot of layers of indirection to how it's used. Cargo (the Rust language's package manager) uses git2-rs, bindings for libgit2 in Rust. Then libgit2 itself is using libssh2 under the hood. Right now we've only seen this issue on Windows, not on other platforms yet.

There's some stack traces in the [last comment](https://github.com/rust-lang/cargo/issues/3401#issuecomment-267794703) on that issue, and I've verified that the segfault reproduces at the current head commit (f7daf3185a4e889fcd5272858db97764a279da1e) of libssh2. It looks like the segfault [happens here](https://github.com/libssh2/libssh2/blob/f7daf3185a4e889fcd5272858db97764a279da1e/src/wincng.c#L1992) because `bn->bignum` is NULL. That function is in turn [called here](https://github.com/libssh2/libssh2/blob/f7daf3185a4e889fcd5272858db97764a279da1e/src/kex.c#L143) during `diffie_hellman_sha1`, [called here](https://github.com/libssh2/libssh2/blob/f7daf3185a4e889fcd5272858db97764a279da1e/src/kex.c#L1472) in `kex_method_diffie_hellman_group14_sha1_key_exchange`, which I believe itself originates [here](https://github.com/libssh2/libssh2/blob/f7daf3185a4e889fcd5272858db97764a279da1e/src/kex.c#L2550).

I do realize that this is unfortunately not a huge amount to go off, but I'm hoping that someone with more knowledge than I could recognize something here and perhaps spot the bug! If it helps the program is trying to connect to `github.com` over SSH with the username `git`, and the ssh-agent itself was set up with `eval $(ssh-agent)`.

If any more information is needed I can certainly try to help provide it!

alexcrichton:
This commit fixes a segfault seen dereferencing a null pointer on Windows when
using ssh-agent. The problem ended up being that errors weren't being
communicated all the way through, causing null pointers to be used when
functions should have bailed out sooner.

The `_libssh2_dh_key_pair` function for WinCNG was modified to propagate errors,
and then the two callsites in kex.c of `diffie_hellman_sha{1,256}` were updated
to recognize this error and bail out.

Closes #162

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -133,8 +133,10 @@ static int diffie_hellman_sha1(LIBSSH2_SESSION *session,
         memset(&exchange_state->req_state, 0, sizeof(packet_require_state_t));
 
         /* Generate x and e */
-        libssh2_dh_key_pair(&exchange_state->x, exchange_state->e, g, p,
-                            group_order, exchange_state->ctx);
+        rc = libssh2_dh_key_pair(&exchange_state->x, exchange_state->e, g, p,
+                                 group_order, exchange_state->ctx);
+        if (rc)
+            goto clean_exit;
 
         /* Send KEX init */
         /* packet_type(1) + String Length(4) + leading 0(1) */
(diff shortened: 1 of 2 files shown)


alexcrichton:
I've confirmed locally that this does indeed segfault without this PR and then with this PR applied our segfault disappears, so it at least fixes that bug!

bagder:
Lovely, thanks a lot! Merged now.

PR #164, opened 2016-12-22
Add support for SHA256 host key fingerprints

sbredahl:
This commit adds SHA256 host key fingerprint support. This has been the default since OpenSSH 6.8 and now that libssh2 has SHA256 as a crypto lib requirement, this commit should not cause too much trouble.

Because libssh2 computes the host key hash during the kex, we don't know (a priori) if SHA1 or SHA256 fingerprints will be used, if at all, so we need to compute both. This is also true for MD5 fingerprints, if enabled.

### DIFF
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -403,7 +403,8 @@ typedef struct _LIBSSH2_POLLFD {
 /* Hash Types */
 #define LIBSSH2_HOSTKEY_HASH_MD5                            1
 #define LIBSSH2_HOSTKEY_HASH_SHA1                           2
-
+#define LIBSSH2_HOSTKEY_HASH_SHA256                         3
+    
 /* Hostkey Types */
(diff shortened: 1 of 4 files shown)


willco007:
This looks good, my only suggestion is the debug output should log the base 64 string like ssh for easier comparison with ssh.  My personal fork uses this:

`
char *base64Fingerprint = NULL;
_libssh2_base64_encode(session, (const char*)session->server_hostkey_sha256, SHA256_DIGEST_LENGTH, &base64Fingerprint);
if ( base64Fingerprint != NULL ) {
	_libssh2_debug(session, LIBSSH2_TRACE_KEX,
				   "Server's SHA256 Fingerprint: %s", base64Fingerprint);
	LIBSSH2_FREE(session, base64Fingerprint);
}
`

sbredahl:
This PR has been replaced with #180

PR #165, opened 2017-01-08
Sane wait socket

### LINKED ISSUE #147: Memory leaks / heap corruption
Hello.

If:
1. Work with session in non-blocking mode.
2. Create channel with libssh2_channel_direct_tcpip (operation in blocking mode and switch back).
3. Close channel  libssh2_channel_free (operation in blocking mode and switch back).

Leads to memory leaks and sometimes to heap corruption.

Maybe I do something wrong?

salva:
In order to fix #147, this patch solves several issues in `_libssh2_wait_socket` and also adds a new function, `libssh2_set_socket_disconnected` to tell the library that no further interaction with the remote server should be attempted.

For instance, in order to try an ordered session cleanup and if that fails a forced dirty one, the following code can be used:

      if (libssh2_free_session(session) != LIBSSH2_ERROR_NONE) {
          /* force session cleanup */
          libssh2_set_socket_disconnected(session);
          libssh2_free_session(session);
      }

The documentation for `libssh2_set_socket_disconnected` is still missing. I would add it once this patch is reviewed, if it is considered good.

### DIFF
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -563,6 +563,8 @@ LIBSSH2_API int libssh2_session_disconnect_ex(LIBSSH2_SESSION *session,
 
 LIBSSH2_API int libssh2_session_free(LIBSSH2_SESSION *session);
 
+LIBSSH2_API int libssh2_session_set_socket_disconnected(LIBSSH2_SESSION *session);
+
 LIBSSH2_API const char *libssh2_hostkey_hash(LIBSSH2_SESSION *session,
                                              int hash_type);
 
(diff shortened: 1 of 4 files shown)


willco007:
The assert on line 404 of transport.c can be hit incorrectly if line 573 hits rc == LIBSSH2_ERROR_EAGAIN. The clearing of the payload variable needs to happen directly after the fullpacket() call, not after the early return in the EAGAIN case.

jonhoo:
This should probably not be marked as stale given that it fixes a real leak.

willco007:
This PR was never updated after my last comments blocking it from inclusion; that's why the stale bot closed the issue.

PR #169, opened 2017-01-21
ssh2_echo: channel IO can not really be interleaved, fixes #167

### LINKED ISSUE #167: 'example/ssh2_echo' fails
The example `ssh2_echo 'server-ip' 'username' 'password'` fails.
The error occurs in line 267.

`libssh2_session_last_error(session)` reports:
`read failed: Unable to send transfer-window adjustment packet, deferring`
Thanks!

salva:
When `libssh2_channel_read`|`write` returns `EAGAIN` the same function
must be called repeatedly with the same arguments until it returns
success or a different error but ssh2_echo was not honoring that
requirement.

This patch sets the session to blocking just before calling the
read/write methods in order to ensure it.

### DIFF
--- a/example/ssh2_echo.c
+++ b/example/ssh2_echo.c
@@ -256,7 +256,10 @@ int main(int argc, char *argv[])
                 continue;
 
             if (fds[0].revents & LIBSSH2_POLLFD_POLLIN) {
-                int n = libssh2_channel_read(channel, buffer, sizeof(buffer));
+                int n;
+                libssh2_session_set_blocking(session, 1);
+                n = libssh2_channel_read(channel, buffer, sizeof(buffer));
+                libssh2_session_set_blocking(session, 0);
                 act++;
 
                 if (n == LIBSSH2_ERROR_EAGAIN) {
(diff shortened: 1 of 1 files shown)


example/ssh2_echo.c
```diff
@@ -256,7 +256,10 @@ int main(int argc, char *argv[])
                 continue;
 
             if (fds[0].revents & LIBSSH2_POLLFD_POLLIN) {
-                int n = libssh2_channel_read(channel, buffer, sizeof(buffer));
+                int n;
+                libssh2_session_set_blocking(session, 1);
```
pkittenis: This uses blocking mode for reads and writes that would block, while `ssh2_echo` is claiming to `tell libssh2 we want it all done non-blocking`.

If only reads should happen on the same session while read returns EAGAIN, shouldn't the read/write loops be changed to do that instead?

I'm not sure that is a limitation though, have not come across it and have tried interleaving reads with writes in non-blocking mode. There is at least [one other example script](https://www.libssh2.org/examples/tcpip-forward.html) that does the same and works fine, as does `ssh2_echo` for me.

doublex:
+1

willco007:
Sorry for the slow merge on this; any chance we could get an updated PR, thanks!

PR #173, opened 2017-02-08
Added WinCNG support for AES CTR mode

madwizard-thomas:
I have added support for AES CTR mode (aes128-ctr, aes192-ctr, aes256-ctr) in the WinCNG backend. It is based on ECB mode similar to the OpenSSL implementation. With these modes supported the WinCNG backend will also be compatible again with the latest OpenSSH server settings so the test suite will run with this backand now as well.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -643,3 +643,30 @@ void *_libssh2_calloc(LIBSSH2_SESSION* session, size_t size)
     }
     return p;
 }
+
+/* XOR operation on buffers input1 and input2, result in output.
+   It is safe to use an input buffer as the same output buffer */
+void _libssh2_xor_data(unsigned char *output,
+                       const unsigned char *input1,
+                       const unsigned char *input2,
+                       size_t length)
+{
+    size_t i;
+
+    for (i = 0; i < length; i++)
+        *output++ = *input1++ ^ *input2++;
+}
+
+/* Increments an AES CTR buffer to prepare it for use with the next AES block */
+void _libssh2_aes_ctr_increment(unsigned char *ctr,
+                                size_t length)
+{
+    if (length == 0)
+        return;
+    size_t i = (length - 1);
+    while (ctr[i]++ == 0xFF) {
+        if (i == 0)
+            break;
+        i--;
+    }
+}
--- a/src/misc.c
+++ b/src/misc.c
@@ -661,12 +661,16 @@ void _libssh2_xor_data(unsigned char *output,
 void _libssh2_aes_ctr_increment(unsigned char *ctr,
                                 size_t length)
 {
-    if (length == 0)
-        return;
-    size_t i = (length - 1);
-    while (ctr[i]++ == 0xFF) {
-        if (i == 0)
-            break;
-        i--;
+    unsigned char *pc;
+    unsigned int val, carry;
+  
+    pc = ctr + length - 1;
+    carry = 1;
+
+    while(pc >= ctr)
+    {
+        val = (unsigned int)*pc + carry;
+        *pc-- = val & 0xFF;
+        carry = val >> 8;
     }
 }
--- a/src/wincng.c
+++ b/src/wincng.c
@@ -253,6 +253,17 @@ _libssh2_wincng_init(void)
         }
     }
 
+    ret = BCryptOpenAlgorithmProvider(&_libssh2_wincng.hAlgAES_ECB,
+        BCRYPT_AES_ALGORITHM, NULL, 0);
+    if (BCRYPT_SUCCESS(ret)) {
+        ret = BCryptSetProperty(_libssh2_wincng.hAlgAES_ECB, BCRYPT_CHAINING_MODE,
+            (PBYTE)BCRYPT_CHAIN_MODE_ECB,
+            sizeof(BCRYPT_CHAIN_MODE_ECB), 0);
+        if (!BCRYPT_SUCCESS(ret)) {
+            (void)BCryptCloseAlgorithmProvider(_libssh2_wincng.hAlgAES_ECB, 0);
+        }
+    }
+
     ret = BCryptOpenAlgorithmProvider(&_libssh2_wincng.hAlgRC4_NA,
                                       BCRYPT_RC4_ALGORITHM, NULL, 0);
     if (BCRYPT_SUCCESS(ret)) {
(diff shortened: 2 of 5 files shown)


mback2k:
Thanks. Do you think it might be possible to abstract the CTR-on-ECB-mode implementation and make the OpenSSL and WinCNG backends use the same abstraction layer? I would prefer to have just one CTR-on-ECB-mode implementation in libssh2. What do you think?

madwizard-thomas:
That might be a good idea, I'll see if I can look into that. I think libgcrypt has a native ctr implementation so it wouldn't be shared across all backends.
I've also looked into GCM mode but I think the abstract crypto layer would need some more changes to support additional authenticated data.

madwizard-thomas:
Hmm looking a bit further into it there's only a very small part of the code that is similar. The crypto functionality works very differently in the backends and I think it would be hard to make an abstraction and still fit the cipher type/context model that is used. The only code that would be shared is the xor and increment of the counter:
```c
 for (i = 0; i < 16; i++)
        *out++ = *in++ ^ b1[i];

    i = 15;
    while (c->ctr[i]++ == 0xFF) {
        if (i == 0)
            break;
        i--;
    }
```
And actually the first loop is different because in the WinCNG backend the xor happens in place.

src/wincng.c
```diff
@@ -253,6 +253,17 @@ _libssh2_wincng_init(void)
         }
     }
 
+    ret = BCryptOpenAlgorithmProvider(&_libssh2_wincng.hAlgAES_ECB,
+        BCRYPT_AES_ALGORITHM, NULL, 0);
+    if (BCRYPT_SUCCESS(ret)) {
+        ret = BCryptSetProperty(_libssh2_wincng.hAlgAES_ECB, BCRYPT_CHAINING_MODE,
+            (PBYTE)BCRYPT_CHAIN_MODE_ECB,
+            sizeof(BCRYPT_CHAIN_MODE_ECB), 0);
+        if (!BCRYPT_SUCCESS(ret)) {
```
mback2k: Please align the lines with the opening ( , just like in lines 249 and 250. Otherwise this patch looks really good. I am looking forward to testing and trying it soon.

src/wincng.c
```diff
@@ -1707,67 +1718,114 @@ _libssh2_wincng_cipher_init(_libssh2_cipher_ctx *ctx,
         return -1;
     }
 
-    if (type.dwUseIV) {
-        pbIV = malloc(dwBlockLength);
-        if (!pbIV) {
+    pbIV = NULL;
+    pbCtr = NULL;
+    dwIV = 0;
+    dwCtrLength = 0;
```
mback2k: Thanks, I like your approach to setting up these variables. I think it is more flexible than my original approach.

mback2k:
Thanks for looking into this. Are you are right about the first loop. But maybe at least the counter increment function could be made generic, even though openssl.c only uses a fixed length of 16 bytes. I am not sure about the best place to put such a function, maybe misc.c or crypt.c?

madwizard-thomas:
I've moved both loops to functions in misc.c. This seemed the most appropriate since crypto.c pretty much just contains the high level crypto method abstractions and not a generic collection of crypto functions.
I've also fixed the code style issue.

src/wincng.h
```diff
@@ -299,21 +302,20 @@ struct _libssh2_wincng_cipher_ctx {
 struct _libssh2_wincng_cipher_type {
     BCRYPT_ALG_HANDLE *phAlg;
     unsigned long dwKeyLength;
-    unsigned long dwUseIV;
+    int useIV;
+    int ctrMode;
```
mback2k: Thanks for your changes. What do you think about using bool instead of int? I know that bool usually maps to int on most plattforms/toolchains, but the compiler could optimize it since only a boolean value is stored.

madwizard-thomas: Bool would be fine by me, I wasn't sure if it was compatible with the C specs libssh2 adheres to, I didn't see boolean being used anywhere.

src/misc.c
```diff
@@ -643,3 +643,30 @@ void *_libssh2_calloc(LIBSSH2_SESSION* session, size_t size)
...
+}
+
+/* Increments an AES CTR buffer to prepare it for use with the next AES block */
+void _libssh2_aes_ctr_increment(unsigned char *ctr,
+                                size_t length)
+{
+    if (length == 0)
+        return;
+    size_t i = (length - 1);
+    while (ctr[i]++ == 0xFF) {
```
madwizard-thomas: Actually this loop should probably not break early depending on cryptographic data in the counter bytes. It should run in constant time to prevent timing attacks.

src/wincng.h
mback2k: @bagder What do you think about using 'bool'?

bagder: I actually like using bool, but as we have no code already using it within libssh2 and it isn't a native type in C89 it needs to be used with care if we're going to introduce it now. So, the easy route is to add a comment about it but avoid it for now until we have someone introduce it properly!

mback2k: Okay, thanks for your comment. Sounds like a good plan to me.

src/misc.c
mback2k: That's right, but we have to find a way to do something that cannot be optimized away by compilers inside the loop instead of running no-ops.

src/misc.c
```diff
@@ -643,3 +643,30 @@ void *_libssh2_calloc(LIBSSH2_SESSION* session, size_t size)
...
+        *output++ = *input1++ ^ *input2++;
+}
+
+/* Increments an AES CTR buffer to prepare it for use with the next AES block */
+void _libssh2_aes_ctr_increment(unsigned char *ctr,
+                                size_t length)
+{
+    if (length == 0)
+        return;
+    size_t i = (length - 1);
```
mback2k: This line causes issues with some compilers, because size_t i is declared after code. Please move size_t i, to the top of the function. E.g. https://ci.appveyor.com/project/alamaison/libssh2-nkc47/build/168/job/4sapbrepqyhvx2y3

madwizard-thomas:
I've implemented the AES CTR code in a time constant manner and added a note about the boolean type. I think everything is fixed now?

mback2k:
@madwizard-thomas Thanks a lot for your contribution. I just reordered and reformatted the commits a little bit and just pushed your changes to the master branch on this repository. Consider this pull request merged.

PR #178, opened 2017-05-07
Generate debug info when building with MSVC

jethrogb:
I'm not sure why the Visual Studio generators wouldn't do this automatically. I copied what libgit2 is doing.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -355,6 +355,11 @@ elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
   target_compile_definitions(libssh2 PRIVATE LIBSSH2_DARWIN)
 endif()
 
+if(MSVC)
+  set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} /Zi /Od")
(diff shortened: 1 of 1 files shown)


jethrogb:
@alamaison @wildart @dbyron0 ping

dbyron0:
Sorry, I imagine this is right but know very little about cmake.

jethrogb:
@alamaison @wildart @dbyron0 ping

alamaison:
Don't you just need to switch to the Debug or RelWithDebugInfo configuration in the generated visual studio project?

jethrogb:
It's been a long time. I think I tried that but it didn't work. Do you want me to check again?

jethrogb:
Instead of marking it stale, maybe just merge it?

bagder:
Anyone (else) building on Windows who can say something :+1: or :-1: about this PR?

jethrogb:
not stale

willco007:
@jethrogb did you happen to test the suggestion from @alamaison about changing the project configuration?

jethrogb:
Ok I finally booted into Windows to test this flow again. I'm not using the graphical Visual Studio interface to build anything. I'm just using CMake for the entire build. I confirmed that it still doesn't generate symbols without this change.

Issue #193, opened 2017-07-02
Windows build errors


simon-p-r:
Hi

I am having lots of errors when trying to compile libssh2 on Windows 10, it appears I cannot use a custom install location for OpenSSL to link against.  

Thanks
Simon

simon-p-r:
I was about to ask on stackoverflow before someone but this [question](https://stackoverflow.com/questions/44889540/compiling-libssh2-on-windows) up, it appears building with openssl crypto backend is the problem.

Samega7Cattac:
Ya, I made that question.
Using OpenSSL is tricky, now work well.
Is better to see the configs if you use cmake-gui.

pkittenis:
There's a [conda package for libssh2](https://github.com/conda-forge/libssh2-feedstock) that seems to use OpenSSL backend.

PR #194, opened 2017-07-04
proposed fix for issue #193

simon-p-r:
Allows a define of OPENSSL_CUSTOM_DIR to set the CMake variable OPENSSL_ROOT_DIR which allows build


tiennou:
Did you try setting `OPENSSL_ROOT_DIR` directly ? I have a hard time understanding how adding an indirect variable.

Note that, in my quest to make it build cleanly against various specific `openssl` versions on macOS, I had to use `cmake .. -DOPENSSL_INCLUDE_DIR=$OPENSSL_ROOT_DIR/include -DOPENSSL_CRYPTO_LIBRARY=$OPENSSL_ROOT_DIR/lib/libcrypto.dylib -DOPENSSL_SSL_LIBRARY=$OPENSSL_ROOT_DIR/lib/libssl.dylib`, with `OPENSSL_ROOT_DIR` set to my openssl install prefix, so I do realize there might be something fishy with the default CMake OpenSSL module...

simon-p-r:
Yes I have tried to set OPENSSL_ROOT_DIR directly, I agree there is something very fishy about CMake's FindOpenSSL module.  I am compiling 32-bit and FindOpenSSL always finds and tries to link to 64-bit version I have installed on my machine.  I will try your defines out and do a PR for the build documents if it works.

simon-p-r:
@tiennou I am afraid that does not work, I tried to define the following

- DOPENSSL_ROOT_DIR=path/to/my/local/install
- DOPENSSL_LIBRARIES=path/to/libs
- DOPENSSL_INCLUDE_DIR=path/to/includes

However CMake still only finds the 64 bit install.

PR #199, opened 2017-07-11
Memory leaks / heap corruption

tiennou:
This is a PR for the changes in #147. Note that I tried to rebase them against `master` because they seemed to be based on *some* released version, but might have failed to do so.

@muton85 I'd appreciate if you could check that I didn't munge your changes too much.


tiennou:
IIRC, I was under the impression there might be a better way of tackling 4ad0c16.

colinw-smartrg:
Some of these fixes look important,

willco007:
The PR needs to be fixed up to pass the style linter checks; I'd be willing to review it once it's passing.

tiennou:
Sure, since the linter was enabled after that PR's commits were actually collected, and it's currently a pain to run the linter locally, here's a rebased version on top of #372.

Be aware that I'm not in any position to understand or debug any of that code — I've collected the parts that looked correct from #147, and only did that to keep the ball rolling because we saw some related crashes. I'm honestly not sure I care anyhow, since I wrote #243 and am still waiting for review there.

Issue #41, opened 2015-09-06
Add support for ECDH key exchange


bk2204:
ECDH key exchange is faster than regular Diffie-Hellman with equivalent levels of security. It's been supported for some time in OpenSSH. It also uses longer session hashes, resulting in harder-to-guess key material.  It would be nice if libssh2 supported this as well.

The NIST curves can probably be trivially implemented with OpenSSL and libgcrypt, but using curve25519-sha256@libssh.org would probably require custom code in libssh2.

svoop:
Will Cosgrove worked on diffie-hellman-group-exchange-sha256 support in January:
http://www.libssh2.org/mail/libssh2-devel-archive-2015-01/0004.shtml

fgouget:
This is getting somewhat urgent. libssh2 only supports the following key exchange protocols: diffie-hellman-group14-sha1,diffie-hellman-group-exchange-sha1,diffie-hellman-group1-sha1

But OpenSSH now disables [diffie-hellman-group1-sha1](http://www.openssh.com/legacy.html) by default, and [SHA1](https://www.schneier.com/blog/archives/2005/02/cryptanalysis_o.html) in general is considered broken so the rest of libssh2's key exchange protocols are bound to follow suite. Indeed some [SSH hardening guides](https://stribika.github.io/2015/01/04/secure-secure-shell.html) recommend only supporting curve25519-sha256@libssh.org and diffie-hellman-group-exchange-sha256, neither of which libssh2 supports.

bagder:
We accept pull requests as well as good old patches mailed to the mailing list.

jakob:
@fgouget libssh2  does support `diffie-hellman-group-exchange-sha256` in the most recent version (1.7.0). The website hasn't been updated yet with this info.

@badger Could you update the website to include this info?

bagder:
done!

willco007:
I have ECDH using OpenSSL support done on a private fork; I just need some time to merge it back into master and create a pull request.

prem-nm:
I am also looking for ECDH support in libssh2. Hope you find the time to merge back your changes :-)

ehamberg:
Any news on this, @willco007? :)

willco007:
@ehamberg Thanks for the reminder.  I finally shipped the project I've been working on and will have some time in the coming weeks to get my changes into a pull request.

[...]

PR #206, opened 2017-08-31
Adds support for ECDSA keys and host keys #41

willco007:
Has been in shipping code for a while.


willco007:
Will look into the test_hostkey failures tomorrow.

willco007:
As I noted on the email list, I suspect the host key tests are failing due to the fact libssh2 will now prefer ecdsa-sha2-nistp256 host keys when connecting to sshd.

prem-nm:
You can combine the hash and verify functions to a single function, i.e kex_ecdh_sha512_hash_create_verify() can have a extra argument on the hash size and you can have a single function kex_ecdh_sha_hash_create_verify() for all hash sizes. You have done something similar and made inline function LIBSSH2_KEX_METHOD_SHA_VALUE_HASH, so you could try to do this.

willco007:
Thanks for the feedback. I had considered that and even started down that route at one point but it was getting a bit unwieldy. I will give it another shot and see what I can do.

willco007:
Changed hash verifies to macros to avoid so much boilerplate code. Not a huge fan of large macros, but in this case it should be fine.

willco007:
Whew, finally got the tests updated correctly. Anyone have any other feedback on this?

sbredahl:
Impressive work you have done here! As for WINCNG support, is anything new needed apart from SHA384 support? The hash already exists (BCRYPT_SHA384_ALGORITHM) so it is not difficult to implement if this is the only hurdle?

willco007:
@sbredahl Yes, a lot more unfortunately. All the ECDSA function prototypes in crypto.h need to be implemented which does the actual ECDSA encryption.

willco007:
@sbredahl Furthermore, I believe the WINCNG backend uses a custom file parser for reading key files, which would also need to be updated to handle the ECDSA key file format.

willco007:
@sbredahl, it uses the same order of kex preference as OpenSSH, which is 256, 384, 512. However, clients can specify their own custom order using the libssh2 API.

src/openssl.c
```diff
@@ -780,43 +1034,129 @@ _libssh2_sha256(const unsigned char *message, unsigned long len,
...
 int
-_libssh2_md5_init(libssh2_md5_ctx *ctx)
+_libssh2_sha384_init(libssh2_sha384_ctx *ctx)
 {
 #ifdef HAVE_OPAQUE_STRUCTS
     *ctx = EVP_MD_CTX_new();
 
     if (*ctx == NULL)
-        return 0;
+    return 0;
```
vszakats: The return lines need indenting to avoid warnings with new C compilers.

src/openssl.c
```diff
@@ -780,43 +1034,129 @@ _libssh2_sha256(const unsigned char *message, unsigned long len,
...
 
-    if (EVP_DigestInit(*ctx, EVP_get_digestbyname("md5")))
-        return 1;
+    if (EVP_DigestInit(*ctx, EVP_get_digestbyname("sha384")))
+    return 1;
```
vszakats: The return line (also) needs indenting to avoid warnings with new C compilers.

src/openssl.c
```diff
@@ -780,43 +1034,129 @@ _libssh2_sha256(const unsigned char *message, unsigned long len,
...
+    unsigned char *out)
 {
-    unsigned char *p = buf;
+#ifdef HAVE_OPAQUE_STRUCTS
+    EVP_MD_CTX * ctx = EVP_MD_CTX_new();
 
-    /* Left space for bn size which will be written below. */
-    p += 4;
+    if (ctx == NULL)
+    return 1; /* error */
```
vszakats: The missing indentation issue is present here and in a bunch of further blocks later in the new code.

willco007:
Thanks @vszakats for taking a look at this. I've fixed the indentation issues in openssl.c (likely caused by my mass tabs to spaces conversion).

src/openssl.c
```diff
@@ -1088,6 +1694,13 @@ _libssh2_pub_priv_keyfile(LIBSSH2_SESSION *session,
         break;
 #endif /* LIBSSH_DSA */
 
+#if LIBSSH2_ECDSA
+    case EVP_PKEY_EC :
+    st = gen_publickey_from_ec_evp(
+            session, method, method_len, pubkeydata, pubkeydata_len, pk);
+    break;
```
vszakats: Thanks @willco007! Just one more block to be indented above.

vszakats:
Thank you @willco007 for all your fixes and for implementing this. Great job!

willco007:
The Travis tests are failing due to a Java installation error (404).

willco007:
@bagder Do you have time to review this?  I'd like to get this landed as my ED25519 key support relies on this code. Let's keep the ball rolling. :)

vszakats:
Restarted the Travis CI session. Package issues were apparently solved by now.

ehamberg:
Any news on this? :)

topilski:
Please merge it.

willco007:
@vszakats Do you feel like this is ready to merge, care to review it one last time and merge?

vszakats:
@willco007 I'm ready to merge this, just some minor formalities to fix with it, such as tabs, line-ending spaces and non-ANSI89 comments.

Here's a grep highlighting these:
```shell
$ curl -LO https://github.com/libssh2/libssh2/pull/206.patch
$ grep '^\+' 206.patch | grep -n -E '(\t|//)'
$ grep '^\+' 206.patch | grep -n -E ' $'
```

```
6:+#define LIBSSH2_KNOWNHOST_KEY_ECDSA<>   (4<<18)
29:+_libssh2_ecdsa_create_key(_libssh2_ec_key **out_private_key,.
108:+    //domain length
125:+    //public key
794:+        unsigned char *s = data + 1; //advance past packet type
1448:+   <->_libssh2_ec_key *private_key;   /* SSH2 ecdh private key */
1933:+    //key type
1936:+    //name domain
1939:+    //public key
2168:+<>LIBSSH2_EC_CURVE_NISTP256 = NID_X9_62_prime256v1,
2169:+<>LIBSSH2_EC_CURVE_NISTP384 = NID_secp384r1,
2170:+<>LIBSSH2_EC_CURVE_NISTP521 = NID_secp521r1,
2465:+            //return 1;
2467:+            //return 1;
2469:+            //return 1;
```

vszakats:
Linking to https://github.com/libssh2/libssh2/issues/41

vszakats:
@willco007 One more request: A short summary of what's been done, which new KEX names, host keys and curve names got support with your patch, would be great.

willco007:
@vszakats Thanks; I _think_ I got everything you noted fixed up.  As for a summary; this commit lands full ECDSA key support when using the OpenSSL backend. Which includes: 

New KEX methods:
ecdsa-sha2-nistp256, ecdsa-sha2-nistp384, ecdsa-sha2-nistp521

Can now read OpenSSL formatted ECDSA key files.

Now supports known host keys of type ecdsa-sha2-nistp256.

New curve types:
NID_X9_62_prime256v1, NID_secp384r1, NID_secp521r1

Default host key preferred ordering is now nistp256, nistp384, nistp521, rsa, dss.

vszakats:
Thank you @willco007! The PR is now merged and pushed into master.

kdudka:
This seems to introduce new compiler warnings.  Could you please have a look?

https://travis-ci.org/kdudka/libssh2/builds/351472344#L1377

willco007:
@kdudka It would seem your fork isn't up to date. For example, HAVE_OPAQUE_STRUCTS as been defined in the main project for a while now.

vszakats:
@kdudka's concerns are valid (I'm not sure where to find them in the mainline CI output though). It's the `#ifdef` vs `#if` issue, an unused argument warning and an uninitialized variable warning.

willco007:
Yeah, this would be a concern. I also never saw any warnings about this in any builds @vszakats noted. I will submit a patch changing the three offenders in openssl.c to #ifdef.

vszakats:
Thanks @willco007. Can you also look into the other two warnings?:

* `secret` variable can truly be uninitialized in certain cases in `libssh2_ecdh_gen_k()`, so it's definitely something wrong there.
* In `ecdh_sha2_nistp()`, `data_len` argument is indeed unused, so the question is if it's a mistake and it _should_ be used (but isn't), or if not necessary, the argument should be remove from that `static` function.

willco007:
I'm on it; I just committed a fix for the `secret` warning and will put in additional bounds checking in `ecdh_sha2_nistp` to use the data_len value.

mback2k:
@vszakats This commit seems to have broken all of my builds targeting Windows. Could you please check the commit status? Github links to builds on the following site with more details: https://libssh2build.uxnr.de/waterfall

willco007:
@mback2k When I wrote this code there was no winCNG backend and it appears it isn't stubbed out. In wincng.h you need to add: 

#define LIBSSH2_ECDSA 0

#if LIBSSH2_ECDSA
#else
#define _libssh2_ec_key void
#endif

That should fix the build when using this backend. I will do this on Monday, but I don't have my dev env set up here at home to test the fix.

It would be nice if whomever wrote this new backend to write some tests to ensure it actually works.

willco007:
@mback2k Alright, I created a pull request #245. Could you try and build it in your system and see if it works? Sorry for breakage.

mback2k:
@willco007 Thanks, your pull request solved my issue.

PR #208, opened 2017-09-08
scp: do not NUL-terminate the command for remote exec

kdudka:
It breaks SCP download/upload from/to certain server implementations.

The bug does not manifest with OpenSSH, which silently drops the NUL
byte (eventually with any garbage that follows the NUL byte) before
executing it.

Bug: https://bugzilla.redhat.com/1489736


willco007:
Looks fine to me, have you've tested this directly on BrickFTP?

kdudka:
@willco007 Sure, the patch makes the SCP/SFTP download work with BrickFTP.

kdudka:
@willco007 Sorry, I meant SCP download/upload, of course...

PR #216, opened 2017-10-16
Support for direct-streamlocal@openssh.com UNIX socket connection

gjalves:
This patch allow to use direct-streamlocal service from OpenSSH 6.7, that allows UNIX socket connections.

### DIFF
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -716,6 +716,10 @@ libssh2_channel_direct_tcpip_ex(LIBSSH2_SESSION *session, const char *host,
 #define libssh2_channel_direct_tcpip(session, host, port) \
   libssh2_channel_direct_tcpip_ex((session), (host), (port), "127.0.0.1", 22)
 
+LIBSSH2_API LIBSSH2_CHANNEL *
+libssh2_channel_direct_streamlocal_ex(LIBSSH2_SESSION * session, const char *socket_path,
+                                const char *shost, int sport);
+
 LIBSSH2_API LIBSSH2_LISTENER *
 libssh2_channel_forward_listen_ex(LIBSSH2_SESSION *session, const char *host,
                                   int port, int *bound_port, int queue_maxsize);
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -717,7 +717,8 @@ libssh2_channel_direct_tcpip_ex(LIBSSH2_SESSION *session, const char *host,
   libssh2_channel_direct_tcpip_ex((session), (host), (port), "127.0.0.1", 22)
 
 LIBSSH2_API LIBSSH2_CHANNEL *
-libssh2_channel_direct_streamlocal_ex(LIBSSH2_SESSION * session, const char *socket_path,
+libssh2_channel_direct_streamlocal_ex(LIBSSH2_SESSION * session,
+                                const char *socket_path,
                                 const char *shost, int sport);
 
 LIBSSH2_API LIBSSH2_LISTENER *
--- a/src/channel.c
+++ b/src/channel.c
@@ -434,6 +434,81 @@ libssh2_channel_direct_tcpip_ex(LIBSSH2_SESSION *session, const char *host,
 }
 
 /*
+ * libssh2_channel_direct_streamlocal_ex
+ *
+ * Tunnel TCP/IP connect through the SSH session to direct UNIX socket
+ */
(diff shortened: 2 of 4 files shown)


Red-M:
Can this please be reopened?

Red-M:
This looks like the travis build failed due to some internal issue in the test suite?

gjalves:
I think travis is mumbling about cosmetic aspects, like line lengths and misplaced spaces. I'll review this patch to overcome those errors.

Red-M:
Yeah, I can see that now, its failing on `perl src/checksrc.pl`.

Red-M:
Is this allowed to be merged now?

src/channel.c
```diff
@@ -2222,6 +2225,9 @@ libssh2_channel_write_ex(LIBSSH2_CHANNEL *channel, int stream_id,
     if(!channel)
         return LIBSSH2_ERROR_BAD_USE;
 
+    if(!channel->session)
+        return LIBSSH2_ERROR_SOCKET_RECV;
```
jakob: Can you explain the reason for this check? As far as I understand, there should never be a channel with a NULL session.

Also, the choice of error code here seems odd. Why return this error code?

gjalves: This check is unrelated with this patch and was wrongly pushed. I've stomped with channel->session == NULL and put this check to came along without crash, but I do not know why this error happened. I got the same impression that is an impossible situation, but I do not bothered to go deeper.

willco007:
Could you please add documentation for this new API when you get a chance? Thanks!

Red-M:
@willco007 Sorry to be a pest but this looks like its done now and can be merged?

doorsdown:
Agreed this looks ready for merging.

src/channel.c
```diff
@@ -433,6 +433,84 @@ libssh2_channel_direct_tcpip_ex(LIBSSH2_SESSION *session, const char *host,
...
+                              LIBSSH2_CHANNEL_WINDOW_DEFAULT,
+                              LIBSSH2_CHANNEL_PACKET_DEFAULT,
+                              session->direct_message,
+                              session->direct_message_len);
+
+    if(!channel &&
+        libssh2_session_last_errno(session) == LIBSSH2_ERROR_EAGAIN) {
+        /* The error code is still set to LIBSSH2_ERROR_EAGAIN, set our state
+           to created to avoid re-creating the package on next invoke */
+        session->direct_state = libssh2_NB_state_created;
```
willco007, line 482: IF `EAGAIN` is returned here and this function is called again, `_libssh2_channel_open` isn't re-called. Is this on purpose? It looks like it would then return a junk pointer of channel.

willco007: Disregard, I see the _libssh2_channel_open is outside of the idle if block.

willco007:
The `libssh2_channel_direct_streamlocal_ex.3` man page needs to also be added to the CMakeLists.txt file. But it looks good after that is resolved. Thanks!

willco007:
I'd like to land this, but we're still waiting for final update to the CMakeLists.

lmtr0:
Hello there good day!
> I'd like to land this, but we're still waiting for final update to the CMakeLists.
is the rewrite ready?

src/channel.c
vszakats: Unless I'm misunderstanding something, should we delete this check if it isn't related to this PR, or is this something we need?

vszakats:
This would need a `docs/Makefile.am` update to due conflicts, plus adding the man page also to `docs/CMakeLists.txt`. With these and clearing the question with the unrelated `if()` condition, this is good for merge.

include/libssh2.h
```diff
@@ -716,6 +716,11 @@ libssh2_channel_direct_tcpip_ex(LIBSSH2_SESSION *session, const char *host,
 #define libssh2_channel_direct_tcpip(session, host, port) \
   libssh2_channel_direct_tcpip_ex((session), (host), (port), "127.0.0.1", 22)
 
+LIBSSH2_API LIBSSH2_CHANNEL *
+libssh2_channel_direct_streamlocal_ex(LIBSSH2_SESSION * session,
+                                const char *socket_path,
+                                const char *shost, int sport);
```
vszakats, line 722: I'd be nice to align these.
```suggestion
                                      const char *socket_path,
                                      const char *shost, int sport);
```

PR #218, opened 2017-10-20
Fix typo in crypt.c

pkittenis:
Master branch cmake compile currently failing with 

```
src/crypt.c:57:5: error: expected ‘}’ before numeric constant
     8,                     /* blocksize (SSH2 defines minimum blocksize as 8) */
     ^
```


pkittenis:
That would explain why this was not caught in autoconf builds then - I had enabled crypt none in cmake only. Thanks for the heads up also 👍

PR #223, opened 2017-10-28
[WIP] Fixed twice assigned values

lunixoid:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A warning, found using PVS-Studio:
libssh2/src/packet.c        1005    warn    V519 The 'session->packAdd_state' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 994, 1005.

Which case is true:
1) `session->packAdd_state` always should be `libssh2_NB_state_idle`
or
2) `session->packAdd_state` equals` libssh2_NB_state_sent2`  if condition `if (session->packAdd_state == libssh2_NB_state_sent1)` is true?

### DIFF
--- a/src/packet.c
+++ b/src/packet.c
@@ -993,6 +993,8 @@ _libssh2_packet_add(LIBSSH2_SESSION * session, unsigned char *data,
 
             session->packAdd_state = libssh2_NB_state_sent2;
         }
+        else
+          session->packAdd_state = libssh2_NB_state_idle;
 
         /*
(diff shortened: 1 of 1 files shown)


willco007:
I'm not sure this fix is 100% correct. If EAGAIN is returned from _libssh2_kex_exchange() on 1014 this fix will reset the packAdd_State to idle the second time it's called which wouldn't be correct if EAGAIN is once again returned from _libssh2_kex_exchange(). That said, I can't hit this code in the wild so I'm not comfortable making an alternate suggestion.

PR #224, opened 2017-10-30
buildsystem: drop custom buildconf script, rely on autoreconf

yann-morin-1998:
The buildconf script is currently required, because we need to copy a
header around, because it is used both from the library and the examples
sources.

However, having a custom 'buildconf'-like script is not needed if we can
ensure that the header / exists by the time it is needed. For That, we
can just append the src/ directory to the headers search path for the
examples.

And then it means we no longer need to generate the same header twice,
so we remove the second one from configure.ac.

Now, we can just call "autoreconf -fi" to generate the autotools files,
instead of relying on the canned sequence in "buildconf", since
autoreconf has now long known what to do at the correct moment (future
versions of autotools, automake, autopoint, autoheader etc... may
require an other ordering, or other intermediate steps, etc...).

Eventually, get rid of buildconf now it is no longer needed. In fact, we
really keep it for legacy, but have it just call autoreconf (and print a
nice user-friendly warning).

Update gitignore accordingly, too.

Signed-off-by: "Yann E. MORIN" <yann.morin.1998@free.fr>
Cc: Sam Voss <sam.voss@rockwellcollins.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ Makefile
 Makefile.in
 aclocal.m4
 autom4te.cache
+compile
+test-driver
 config.guess
 config.log
 config.status
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,7 @@ script:
       fi
   - |
       if [ "$B" = "configure" ]; then
-        ./buildconf
+        autoreconf -fi
         ./configure --enable-debug --enable-werror
         make
         make check
--- a/Makefile.am
+++ b/Makefile.am
@@ -43,7 +43,7 @@ os400/libssh2rpg/libssh2_publickey.rpgle				\
 os400/libssh2rpg/libssh2_sftp.rpgle					\
 Makefile.os400qc3.inc
 
-EXTRA_DIST = $(WIN32FILES) buildconf $(NETWAREFILES) get_ver.awk \
+EXTRA_DIST = $(WIN32FILES) $(NETWAREFILES) get_ver.awk \
  maketgz NMakefile RELEASE-NOTES libssh2.pc.in $(VMSFILES) config.rpath \
  CMakeLists.txt cmake $(OS400FILES)
 
--- a/buildconf
+++ b/buildconf
@@ -1,22 +1,8 @@
 #!/bin/sh
 
-LIBTOOLIZE="libtoolize"
+echo "***" >&2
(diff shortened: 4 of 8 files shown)


yann-morin-1998:
@smvoss Note for you! ;-)

smvoss:
Tested +1

yann-morin-1998:
@willco007 Ping? ;-)

willco007:
I can test it out for you, but I don't feel like I know how autoconf, etc work with this project well enough to land this.

yann-morin-1998:
@willco007 I pinged you because you're the last one to have had some activity here...
If you know someone better suited to ping, then be my guest! ;-)
Anyway, I appreciate the testing.
Thanks. :-)

yann-morin-1998:
Gentle "Ping?" ;-)

yann-morin-1998:
Hello! Not trying to pres on this almosr-a-year old PR... But any feedback, please?

yann-morin-1998:
@bagder Thanks for activating the stale bot, it made me notice this PR of mine! ;-)

Any chance someone eventually looks at it now?

Thanks!

yann-morin-1998:
@bagder Thanks! I'll fix up and resubmit. :-)

yann-morin-1998:
As far as I can understand, the Travis-CI failures fdo not seem to be related to this changes at all. They are mostly all network errors.

yann-morin-1998:
@bagder I've rebased and repushed the changes. Any chance of them being applied anytime soon now?

yann-morin-1998:
OK, so this is a relatively simple change that has been lying around for more than two
years now...

autoreconf has been able to call all of automake, autpoint, and other autostuff in the
correct order with the right args for the good part of the past two decades now. Hopefully
this change will not need that much time before it lands... ;-]

Sorry for the rant, but I have no idea what I can do to get this considered further. Even
stating "no" would be fine.

willco007:
Anyone more versed in autoconf have any comments on this? Seems OK to me?

yann-morin-1998:
> Anyone more versed in autoconf have any comments on this? Seems OK to me?

I'll rebase on master and repush, so as to re-trigger the testing jobs, if that can give a bit more confidence...

However, the PR is still marked "changes requested", even though I indeed fixed the two additional files pointed out by @bagder.

Hopefully, we won't have to wait for stale-bot to pop in again before this gets applied. ;-)

PR #229, opened 2017-12-01
update crypt.c to fix a bug in FIPS scenario

jitu059:
Problem Description: when FipsAlgorithmPolicy is ON(1) in Windows, remote linux connect through libssh2, fails.
Post cipher exchange, packet length becomes garbage value.
Decrypt of packet failed or / gave errored result in libssh2, post cipher / key exchange happened.- Ciphers used -aes-128-ctr [erroneous]
Fix:
We should disable CTR methods in windows FIPS case. But reordering will solve the issue. as it will give priority to AES_CBC ciphers.
Tested: Windows to Linux , Linux to Linux

### DIFF
--- a/src/crypt.c
+++ b/src/crypt.c
@@ -311,11 +311,6 @@ static const LIBSSH2_CRYPT_METHOD libssh2_crypt_method_3des_cbc = {
 #endif
 
 static const LIBSSH2_CRYPT_METHOD *_libssh2_crypt_methods[] = {
-#if LIBSSH2_AES_CTR
-  &libssh2_crypt_method_aes128_ctr,
-  &libssh2_crypt_method_aes192_ctr,
-  &libssh2_crypt_method_aes256_ctr,
-#endif /* LIBSSH2_AES */
 #if LIBSSH2_AES
     &libssh2_crypt_method_aes256_cbc,
     &libssh2_crypt_method_rijndael_cbc_lysator_liu_se,  /* == aes256-cbc */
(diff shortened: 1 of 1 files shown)


willco007:
Changing the order of the default cipher list would weaken it in the base case as ctr is preferred over cbc.

jitu059:
willco007, thanks for the comment. I was not aware of that.Got your point. But CTR itself does not work, when FIPS_mode of openssl is set to ON. Is there any other suggestion you would like to give?

willco007:
Just curious, what does OpenSSH do in this scenario?

jitu059:
well, in this case I need to do some work. I will build openssh and try the similar things and capture how it works. Will need some time for this. Will get back in some days, as setup is not handy for me now.
Thanks for the valuable question.

sbredahl:
Enabling FIPS mode disables modern crypto (such as aes-ctr) and is a really bad idea. Many of the "approved" FIPS ciphers are considered legacy and not supported by todays OpenSSH servers. 

If you badly need FIPS approved crypto for some reason (government use?), I suggest making your own FIPS-compliant fork of libssh2 rather than introducing this abomination into libssh2.

kdudka:
On Friday, December 1, 2017 7:44:25 PM CET Lalata Indu Sahu wrote:
> But CTR itself does not work, when FIPS_mode of openssl is set to ON.

The last time I tried, aes-ctr worked with OpenSSL in FIPS mode.  Actually,
I had to make aes-ctr work in order to be able to use libssh2 in FIPS mode:

https://github.com/libssh2/libssh2/commit/9f6fd5af

willco007:
I'm going to close this as we want to keep the default ciphers as modern as possible.  If you need to change the default cipher list for your project I would recommend using the libssh2_session_method_pref() function to set them to the order you prefer.

PR #243, opened 2018-03-22
Testcases

tiennou:
This imports two new test cases for RSA & DES key parsing (see a191a2bd855ab4d2d5fb1b60b98812bf0735fb0d), developed while doing the SecureTransport backend (not by me though).

Since it tests low-level functions of the API, this overhauls the test harness so we can import and link against the private libssh2 API. Also included are tweaks for some minor discrepancies uncovered when checking all the backends.

This covers OpenSSL, mbedTLS, and libgcrypt. Let's 🤞 the WinCNG backend knows about all these key types 😉.


tiennou:
I've pondered bringing in [clar](https://github.com/vmg/clar), since that's what I'm used to work with. If there's interest, I can switch the test harness to that. IMHO, 👍 are that most of the assertion API is already there, 👎 are that it's built as a one-process system (you don't get process isolation, but it's a 👍 on Windows because processes are expensive).

Opinions ?

bagder:
Maybe bring attention to this work on the mailing list? We seem to lack manpower to deal even with good ideas...

tiennou:
Rebased, I've actually yanked the SecureTransport tests because of incomplete backend support (IIRC, all/some DER/PEM/thingamajig formats aren't recognized by all/some backends).

On the topic of the testsuite, given that AFAICS there's only a few of them, and they look like integration tests, I'd really like to switch to something more developer-friendly, especially since I usually get cold feet if I can't testcase fixes. In all, I don't think it would be that hard (the issue will be autotools 😉), so if you're fine with using clar, I can switch to that.

tiennou:
Okay, this is in a better shape. This is heavily inspired from libgit2's own pre-Azure CI setup, highlights are :
- CI scripts have been extracted from `.travis.yml` into `ci/` for ease of editing.
- testcases have been rewritten using `clar`.
- Travis's CI matrix has been turned down a little : we now do all 3 "POSIX" backends 32/64bits, a couple with ZLIB enabled/disabled, times 2 toolchains (gcc/clang), plus checksrc.pl & packaging.
- `valgrind` run on top of the 64bit CI runs.

Travis is red because of both [failures](https://travis-ci.org/libssh2/libssh2/jobs/514610222), and/or [valgrind](https://travis-ci.org/libssh2/libssh2/jobs/514610216#L9079).

Known issues are :
- I broke the build on AppVeyor, because of too much CMake-ry going on.
- Autotools has been disabled entirely on Travis, because I've not used it in years and I'm not even sure it ran the old test suite.
- Valgrind's suppression file is empty. This is suspicious 😉.

bagder:
> Autotools has been disabled entirely on Travis

Can we help out there somehow?

> Valgrind's suppression file is empty. This is suspicious

If we're using OpenSSL <= 1.0.x using valgrind is just a pain and supressions won't help IMO. With other backends or >= 1.1.1 I don't think we should need any supressions?

tiennou:
> Can we help out there somehow?

I'd be glad, though my personal preference would be to bless CMake as our main buildsystem, instead of carrying around two. I'm not sure how easy it would be to reproduce the test harness configuration in autotools-speak though, or if providing the CMake-generated Makefiles instead of `autoconf` would be sufficient.

> If we're using OpenSSL <= 1.0.x using valgrind is just a pain and supressions won't help IMO. With other backends or >= 1.1.1 I don't think we should need any supressions?

I haven't looked closely at what's causing the leaks yet, obviously most ought to be caused by failed tests (clar aborts cases when one test fails).

Right now what's bothering me is that I suspect the testsuite wasn't actually running when I started, and the few testcases I have seem to rely on backends capabilities which aren't uniform. Having a bunch of more stable testcases on more "basic" stuff, like read data/write data, stuff that doesn't really depend on backends.

For the record (which is why I'm still going at it 😆), my OpenSSL runs are green here. Travis stubbornly [fails](https://api.travis-ci.org/v3/job/514610216/log.txt) with strange -19/-16/-19 errors, which I'm not sure I understand :

```
  1) Failure:
userauth::publickey::auth_succeeds_with_correct_rsa_openssh_key [/home/travis/build/libssh2/libssh2/tests/testcases/userauth/publickey.c:53]
  Test failed.
  Unexpected failure: Callback returned error (-19)

  2) Failure:
userauth::publickey::auth_succeeds_with_correct_rsa_openssh_key [/home/travis/build/libssh2/libssh2/tests/testcases/userauth/publickey.c:75]
  Test failed.
  Unexpected failure: Unable to extract public key from private key. (-16)

  3) Failure:
userauth::publickey::auth_succeeds_with_correct_rsa_openssh_key [/home/travis/build/libssh2/libssh2/tests/testcases/userauth/publickey.c:135]
  Test failed.
  Unexpected failure: Callback returned error (-19)
```

Haven't yet determined if that's a backend-capability issue, or a test-harness setup (not putting the keyfiles in the sandbox, which could cause `-16`, but then I'd have had no private key as well 😕).

willco007:
It appears these OpenSSL 1.1.0 tests are being run with the OpenSSL 1.0.x backend and failing. They are currently skipped when building with anything other then OpenSSL 1.1.x. Hope this helps. 

Will

willco007:
Please disregard what I said. These tests should always be run; I didn't look close enough at there names. That's what I get for replying before coffee. 

Thanks,
Will

tiennou:
Okay, I've added some stat checks on those keys, and it still fails about key stuff. Travis seems to be using `Found OpenSSL: /usr/lib/x86_64-linux-gnu/libcrypto.so (found version "1.0.1f")`, so maybe that's what you alluded to @willco007 ? I'll try to see how hard it would be to bump Travis to use Ubuntu 16.04 instead of 14.04. Hopefully that has a recent-ish OpenSSL version.

(FTR, I'm cross-checking using my [fork's Travis](https://travis-ci.org/tiennou/libssh2) instead of using this PR, because I didn't want to hammer too much the uxnr builders/AppVeyor).

tiennou:
Rebased, a bit more work on this.

- It's running against Travis's Ubuntu 16.04.
- It's building against a source-build of OpenSSL 1.1.1b (apart from the ZLIB builds, which autodetect and thus fallback to the distrib-provided version).
- 32bit builds against libgcrypt & mbedTLS are failing, and I'm not sure if that's something I missed or the result of the switch to 16.04.

I'd like to know if there's any interest into something like it. Now, seeing the test results, I don't feel confident enough to keep moving forward, because I don't quite know why I can run them alright, but they're failing on Travis.

I've been going around in circles trying to tackle #240, but wanted to have a good test baseline before touching it. Is that something that would be appreciated ? Right now I'm trying to cut down on the number of `void *abstract` pointers, and I've ended up with something like [this](https://github.com/libssh2/libssh2/compare/master...tiennou:fix/cleanups).

Since I've been mostly all over the crypto interface, I've found some really confusing things: from the top of my head, the PEM parser for public keys + the one from the backend was the most interesting find.

tiennou:
I'm done, the build should be green. The remaining Appveyor build failed while fetching Docker.

I've made Valgrind run on mbedTLS only at the moment, given that the other two backends don't run under it without causing a failure.
It fixes 2 BN leaks from the gcrypt backend, one which looked pretty bad.

tiennou:
Is there anything I can do to keep that moving forward ? I feel bad for being pushy, but I invested some time into this, and was planning to work on other things (someone downstream requested "the .ssh/config parser" 😉), but that silence is somewhat disconcerting…

willco007:
Any ideas why the AppVeyor test isn't fetching Docker anymore? That needs to be sorted out.

tiennou:
Can you bump it ? I don't have access to the buildbots, and it really looks like a fluke :

```
[NuGet] An error occurred while loading packages from 'https://chocolatey.org/api/v2/': The remote name could not be resolved: 'chocolatey.org'
docker not installed. The package was not found with the source(s) listed.
 Source(s): 'https://chocolatey.org/api/v2/'
 NOTE: When you specify explicit sources, it overrides default sources.
```

I worked on the Win32 part using Visual Studio 2019, and it *built* fine.

tiennou:
✨

tiennou:
Rebased. Any change of this getting looked at ?

willco007:
This looks very promising! I'm OOO for the next couple weeks but I'll take a deeper look at it when I return and hopefully get it landed.

willco007:
@tiennou I've been looking this over and correct me if I'm wrong, but the tests seem to be testing against OpenSSL 1.1.1+ only.  It would be great if it could also test against 1.0.x as well to ensure compatibility.

tiennou:
An OpenSSL 1.0.1 bot has been setup — I've squashed all necessary changes to the first commit, so it's now "possible" to grab a specific version and build against that. Ideally, I'd also like to check the README for documentation discrepancies, but that can wait for a review.

.travis.yml
```diff
@@ -77,3 +77,4 @@ script:
 branches:
     only:
         - master
+        - /ci\/.*/
```
tiennou: I've added this so I can push branches to my fork and get Travis to double-check my WIPs. I've not been able to find a less invasive way to do that short of carrying this change around, so if that's not wanted it can be dropped from the PR.

tiennou:
Good to go again. I've also tried a more recent `valgrind` (3.15, Travis has 3.11) and it "works" — we *only* have a leak of the loaded engines in OpenSSL, as far as the test suite can tell.

Note that I brought in a few of libgit2's CMake modules, because I had different sets of warnings showing up. Now it's a selection, instead of a blanket `-Wall`. I also had to correct the printf specifier because reading a `long int` from an `uint32_t` is dubious, and this caused Coverity to start catching all those discrepancies. So there's a later commit for that 7b822bbd59c4fd50d68149e8aa08b2e567bd8122, and it seems to please both CI and Coverity; in case it's helpful.

tiennou:
That looks like `ctest` output or some other CMake incantation ? I *think* its okay (the 40+ runtime), and that having no output, though a little strange, is because clar doesn't really fit CMake, so the integration is a little bit "simple".

Technically the testsuite is the single `libssh2_clar` executable that should be built, and the `ci/test.sh` script has the "steps" taken by CI.

tiennou:
`libssh2_clar -v`, or a direct `ci/test.sh` invocation, provided you set some ivars should be okay.

willco007:
How are we feeling about this PR? Seems like it's in pretty good shape. @tiennou Do you mind updating the docs on how to run the unit tests using Clar?

tiennou:
@willco007 I've added a quickstart guide as `tests/README.md`, will that be sufficient ? I haven't found if there's already a description somewhere.

Relatedly, if you want more Cool supporting stuff™, I can try to spinup docurium on the source, and see how the docs looks — that's what libgit2 uses for its API reference : https://libgit2.org/libgit2/#HEAD. It's likely it would benefit from a manpage formatter though 😉.

willco007:
That'd be great, there is also a note about running the tests in `docs/INSTALL_CMAKE`. Noting something about Clar there might be useful too.

tiennou:
I've converted most files under docs to Markdown, and did some proofreading (fixed `HACKING.md` and its out-of-date guidelines). Hope this is fine !

PR #245, opened 2018-03-31
Added ECDSA defines

willco007:
Fixed missing defines preventing building


mback2k:
I am going to merge this change since it looks good and aligns WinCNG with the mbedTLS backend regarding the implementation state of ECDSA keys. Thanks.

PR #248, opened 2018-04-25
ED25519 Key Support #39

willco007:
This pull request adds full ED25519 support #39. To support this I also had to add OpenSSH key file reading support which makes this a very large pull request, sorry.  RSA/DSA/ECDSA/ED25519 OpenSSH formatted keys are all supported as well.  The only backend supported is OpenSSL. I also added 3 tests to make sure they work using the OpenSSL backend and fixed a bug in the test cmake file when using an implicit backend.


willco007:
I should mention, the ED25519 curve implementation is from BoringSSL and the other helper files are from BSD, all licenses are compatible with libssh2.

willco007:
Off to fix the mountain of warnings on windows...

topilski:
Hi @willco007 , is this implementation will work only with BoringSSL?

willco007:
@topilski It works with the OpenSSL backend only. The implementation of the curve is from BoringSSL, which uses a handful of OpenSSL functions directly. It could be 'ported' to be backend agnostic by someone who uses the other backends if there is interest.

topilski:
Hi @willco007 , thank you for your answer, do you know when this PR will be merged?

willco007:
Well, when I get it running on old Windows, almost there... 😂 Given the size of the PR I'm guessing it will take a while to review.  Hopefully people interested can start playing with it asap.

willco007:
I think the PR is at a point where I'm happy with it. A variation of this code has been shipping for a while and is known stable. I did have to update it for the current master of libssh2 (it was based on the shipping version). Specifically adding the reading *_from_memory functions. I'm going to write an additional test using libssh2_userauth_publickey_frommemory() using the new key format. Auth from key file and the rest of the PR can be reviewed.

vszakats:
This definitely looks useful and having ED25519 key support is quite cool — Though, if the newly added, low-level, ED25519 code is already present in OpenSSL 1.1.0 and upper (which versions do have ED25519 support built-in), I'd side on using that instead of rolling this significant chunk of sensitive code and keep maintaining it inside libssh2.

willco007:
@vszakats Thanks for your comments; I fully agree adding crypto code directly into the library isn't ideal. As for ED25519 support in OpenSSL, it looks OpenSSL 1.1.1+ will ship with ED25519 support but it's currently pre-release.

Updating the code to the ED25519 OpenSSL context should be no problem. On the plus side, keeping the curve in the library would allow other backends to use it with minor modifications, if that is a priority.

It looks like the OpenSSH key reading code would still be required as OpenSSL doesn't support it (not surprising).

jakob:
@willco007 THANK YOU!!!

I tried building your branch (with autotools), and got an error:

```
Undefined symbols for architecture x86_64:
  "_bcrypt_pbkdf", referenced from:
      __libssh2_bcrypt_pbkdf in libssh2.a(misc.o)
```

It looks like these two source files are missing from Makefile.inc: bcrypt_pbkdf.c blowfish.c

Adding them seems to fix the build (see jakob/libssh2@31ebd120723895bd99c69bed9b2c93be9bdb7730)

I couldn't find any place that actually uses bcrypt_pbkdf. What's the motivation for adding libssh2_bcrypt_pbkdf? We should probably add docs for this new function.

willco007:
@jakob Thanks for noticing the missing C files in the makefile, I've updated the PR with the fixed file. As far as bcrypt_pbkdf, it's used in pem.c to decrypt the new OpenSSH key format. I added the wrapper function _just in case_ a backend couldn't build the various blowfish files and it seemed like convention in libssh2 when dealing with external code. I could just as easily remove it and call bcrypt directly from pem.c if we feel like it's not a big deal.

jakob:
@willco007 Oh, I didn't see that libssh2_bcrypt_pbkdf() is used in pem.c. All good :)

I found a couple of issues related to host keys and known_hosts.

1. There is no LIBSSH2_KNOWNHOST_KEY_ED25519
2. LIBSSH2_KNOWNHOST_KEY_MASK is not expanded to include LIBSSH2_KNOWNHOST_KEY_ECDSA
3. I think we should have separate hostkey types LIBSSH2_HOSTKEY_TYPE_ECDSA_256, LIBSSH2_HOSTKEY_TYPE_ECDSA_384 and LIBSSH2_HOSTKEY_TYPE_ECDSA_521 instead of just LIBSSH2_HOSTKEY_TYPE_ECDSA, and corresponding LIBSSH2_KNOWNHOST_KEY_* constants. Otherwise we don't know which host key method we have to request from the server if we already have a known_host entry.
4. knownhost.c needs to be made aware of the new key types

jakob:
(I just realised my point 2 is incorrect, LIBSSH2_KNOWNHOST_KEY_MASK does include LIBSSH2_KNOWNHOST_KEY_ECDSA. But it would need to be bigger if we treat all the ECDSA keys as different types)

willco007:
I've updated knownhosts, I missed it when I was merging my private fork, thanks. There is no need to break out ECDSA host key types because the only supported ECDSA host key type is 256, at least for sshd.

src/knownhost.c
```diff
@@ -778,6 +778,10 @@ static int hostline(LIBSSH2_KNOWNHOSTS *hosts,
             key_type = LIBSSH2_KNOWNHOST_KEY_SSHDSS;
         else if(!strncmp(key_type_name, "ssh-rsa", key_type_len))
             key_type = LIBSSH2_KNOWNHOST_KEY_SSHRSA;
+        else if (!strncmp(key_type_name, "ecdsa-sha2-nistp256", key_type_len))
```
jakob: We should also test for the strings ecdsa-sha2-nistp384 and ecdsa-sha2-nistp521

jakob:
@willco007 I don't think that is accurate. On my machine (macOS 10.13, OpenSSH_7.6p1, LibreSSL 2.6.2) the default host key is 256 bits, but I can create a key using `ssh-keygen -t ecdsa -b 521` and use that as host key.

It's even possible to have multiple ecdsa host keys with different bit lengths, and you can tell the server which one you want by using eg. `HostKeyAlgorithms=ecdsa-sha2-nistp521`.

However, I've discovered that OpenSSH always requests host keys in the order `ecdsa-sha2-nistp256`, `ecdsa-sha2-nistp384`, `ecdsa-sha2-nistp521` when it finds *any* ecdsa host key in the knownhosts file. So I guess that's an argument for not using separate constants.

But we do need to support `ecdsa-sha2-nistp384` and `ecdsa-sha2-nistp521` host keys (see my inline comments on your commit)

src/knownhost.c
```diff
@@ -1017,6 +1021,14 @@ knownhost_writeline(LIBSSH2_KNOWNHOSTS *hosts,
         key_type_name = "ssh-dss";
         key_type_len = 7;
         break;
+    case LIBSSH2_KNOWNHOST_KEY_ECDSA:
+        key_type_name = "ecdsa-sha2-nistp256";
```
jakob: If we support other key lengths, we can't assume ecdsa-sha2-nistp256 here. Should we check key length to determine name (ecdsa-sha2-nistp256, ecdsa-sha2-nistp384, ecdsa-sha2-nistp521)?

willco007:
Given the large changes in openssl.c (mainly just moving code around), my preference would be to take this PR as-is. Then when OpenSSL 1.1.1 ships I can move the ED25519 code over to it and remove the custom curve file, if that is what we want to do. Merging master onto this PR is not fun, so getting this committed sooner rather than later would be ideal. @bagder Do you have any thoughts?

vszakats:
Maybe I'm missing something (or the support is not appropriate for this particular use-case), but it appears that the 25519 curve is supported since OpenSSL 1.1.0, released in September 2016.

1.1.1 will add (due out in two weeks) ED25519 support to that.

I'm all for having the latest and greatest cypto tech, but my rather strong opinion is that copy/pasting code between projects rarely pans out good and generally a red flag. If someone plans to be on the edge of technology it'd be essential to use the always latest (stable) OpenSSL anyway, and it's not libssh2's job to offer the missing bits for older OpenSSL versions. Such hybrid/intermixed solution will be very difficult to assess from a security POV and as I mentioned above, it will be an arduous maintenance task, which will need to be completed by "someone" till the code is vendored here.

(Performance is a distant second in importance here, but it may be worth a mention, that OpenSSL's ED25519 speed is significantly better than that of BoringSSL: https://github.com/openssl/openssl/pull/3503#issuecomment-303089137)

willco007:
@vszakats The curve is in 1.1.0 but no API exposing it is. I very much agree looping in code is generally bad form. And in this case we can avoid it. However, to stay relevant we may have to in the future. For example, the way OpenSSL implemented chacha20-poly is not compatible with the way it's used in ssh.

topilski:
Hi folks, when you will merge this branch? Many users waiting of this implemenatation, thank you for this work!

willco007:
I'm almost done moving the PR over to OpenSSL 1.1.1. Should be by end of week at the absolute latest.

jakob:
ED25519 public key authentication is broken for us since 2ee3dfe (when building libssh2 with openssl-1.1.1-pre6)

(ed25519 host keys seem to work, but client keys do not)

Debug output from sshd shows an incorrect hash for the client key.

jakob:
Commit 1e8de6bdb26349839d84e080de7bd4d52805707b still doesn't work for me. The last commit that works with ed25519 keys is 8c121379aa232091eca3a7b82cb519a9b44ebbdd (that's just before switching to the OpenSSL implementation)

jakob:
Also, support for passphrase protected OpenSSH format keys was broken somewhere in this pull request. Authentication with a passphrase protected ECDSA key works in master, but it doesn't in this pull request. I'll follow up with more info as soon as I get a chance to dig into it.

willco007:
@jakob I plan on getting to the bottom of this tomorrow. I thought my signing changes fixed client keys but I figured out late this afternoon I was authing via my ssh-agent key instead of hitting the new code in all cases in my app. I'll make sure passphrases are working for all key types as well.

jakob:
Followup on my previous comment: Passphrase protected keys do work, but the problem is how my app uses them. My app first tries calling `libssh2_userauth_publickey_fromfile_ex` without a passphrase. If that function returns `LIBSSH2_ERROR_FILE`, it checks the OpenSSL error codes and prompts for a passphrase. The function no longer returns `LIBSSH2_ERROR_FILE` if there was a problem reading the private key file.

The problem is in _libssh2_pub_priv_keyfile (and _libssh2_pub_priv_keyfilememory). It seems that when PEM_read_bio_PrivateKey fails, you simply try calling _libssh2_pub_priv_openssh_keyfile. This works in the happy case, but it discards error messages in the not happy case.

Also, _libssh2_pub_priv_openssh_keyfile just prints out a generic error message whenever reading a private key fails. We should probably try to output an error message with more details about the failure.

And finally, there is a lot of code duplication in _libssh2_pub_priv_openssh_keyfile / _libssh2_pub_priv_openssh_keyfilememory. Is there a reason why _libssh2_pub_priv_openssh_keyfile doesn't just read the key into a buffer and calls _libssh2_pub_priv_openssh_keyfilememory?

jakob:
Here's a patch that makes `libssh2_userauth_publickey_fromfile_ex` return `LIBSSH2_ERROR_FILE` in case the private key cannot be read, like it used to do.

[libssh2_pub_priv_keyfilememory.patch](https://github.com/libssh2/libssh2/files/2034564/libssh2_pub_priv_keyfilememory.patch.txt)

In the future this should probably return a more detailed error message (eg. unsupported format, opening file failed, passphrase required, passphrase incorrect, etc). But this patch at least makes it work like before.

willco007:
@jakob I've finally fixed the last of the OpenSSL port-over issues.  I added your patch to return `LIBSSH2_ERROR_FILE` error if the OpenSSH key files also fail to be parsed to match the OpenSSL code.

`_libssh2_pub_priv_openssh_keyfile` should return a decent error message as they are set throughout the parsing process. Is there a specific failure where you are you seeing a generic message? I tried to return -1 but set an error message string if the error was known.

As for `_libssh2_pub_priv_openssh_keyfile` vs `_libssh2_pub_priv_openssh_keyfilememory` code. I mainly was following the structure of the other `*_keyfilememory` calls which have slightly different prototypes/requirements. The existing OpenSSL `keyfile` functions could all be bottlenecked to the `filememory` calls as well as they have similar duplication.

jakob:
Thanks for the update! Looks like you forgot to commit crypto.h, I had to apply this patch to make it compile:

```diff
diff --git a/src/crypto.h b/src/crypto.h
index 8f9ad98..fdf3e2a 100644
--- a/src/crypto.h
+++ b/src/crypto.h
@@ -189,7 +189,7 @@ _libssh2_ed25519_new_private(libssh2_ed25519_ctx **ec_ctx,
                             const char *filename, const uint8_t *passphrase);
 int
 _libssh2_ed25519_sign(libssh2_ed25519_ctx *ctx, LIBSSH2_SESSION *session,
-                      uint8_t out_sig[LIBSSH2_ED25519_SIG_LEN],
+                      uint8_t **out_sig, size_t *out_sig_len,
                       const uint8_t *message, size_t message_len);
 
 int
```

willco007:
@jakob Yup, sure did. :)

jakob:
Thanks! So just to confirm, this build now works correctly on my machine.

Concerning the keyfile vs. filememory functions, I think this is something we could refactor at some point after this patch has landed. Aside from code duplication, I think it would also be better to read files only once, otherwise the code won't work when we get special files like `/dev/fd/63`. But one thing at a time...

topilski:
Hi @jakob , @willco007  any news? Or this PR in freezed state?

jakob:
Against all better judgement, I released a build of my app based on this pull request. It's mostly looking good, but I've received one crash report from a customer:

```
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000008
Exception Note:        EXC_CORPSE_NOTIFY
```

Backtrace:
```
0 BN_num_bits (bn_lib.c:136)
1 curve25519_sha256 (kex.c:2678)
2 kex_method_curve25519_key_exchange (kex.c:3043)
3 _libssh2_kex_exchange (kex.c:3990)
4 session_startup (session.c:732)
5 libssh2_session_handshake (session.c:814)
...
```

Looks like a NULL pointer somewhere?

(Server is OpenSSH_7.2p2 Ubuntu-4ubuntu2.4, host key is ssh-ed25519 and kex is curve25519-sha256@libssh.org)

I'll investigate and follow up if I find anything.

src/openssl.c
```diff
@@ -1679,6 +2571,244 @@ _libssh2_ecdh_gen_k(_libssh2_bn **k, _libssh2_ec_key *private_key,
...
+
+    if(out_len != LIBSSH2_ED25519_KEY_LEN) {
+        rc = -1;
+        goto cleanExit;
+    }
+
+    rc = EVP_PKEY_derive(server_key_ctx, out_shared_key, &out_len);
+
+    if(rc == 1 && out_len == LIBSSH2_ED25519_KEY_LEN) {
+        BN_bin2bn(out_shared_key, LIBSSH2_ED25519_KEY_LEN, *k);
```
jakob: At this point, *k would be NULL (set in line 2634)

Maybe this line should be:
```C
        *k = BN_bin2bn(out_shared_key, LIBSSH2_ED25519_KEY_LEN, NULL);
```

jakob: (or line 2634 should be removed if *k is already a BIGNUM)

jakob:
OK, so I've looked at the code in more detail. It looks like `*k` is already initialised before, so `_libssh2_curve25519_gen_k` should not set `*k` to NULL. It would be better if it checked to make sure it is not NULL instead. This should fix it:

```diff
diff --git a/src/openssl.c b/src/openssl.c
index 2a3f14f..6b98673 100644
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -2631,7 +2631,8 @@ _libssh2_curve25519_gen_k(_libssh2_bn **k,
     if(k == NULL)
         return -1;
 
-    *k = NULL;
+    if (*k == NULL)
+        return -1;
 
     bn_ctx = BN_CTX_new();
     if(bn_ctx == NULL)
```

I'm not sure why the argument is `_libssh2_bn **k` -> since the BIGNUM is already initialised, it would be simpler to just use `_libssh2_bn *k`. Or is this for compatibility with other crypto libraries that have a different bignum API?

willco007:
@jakob Thanks for the fix; and the live beta testing. 😄 I've added the null value check and removed the incorrect nulling. Please give it a shot.

topilski:
Hi @jakob , good news thank you for your work!

jakob:
Found a problem when building against OpenSSL 1.1.1-pre7:

`EVP_PKEY_verify_init` returns -2 in `_libssh2_ed25519_verify`. According to the [man page](https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_verify.html):

> a return value of -2 indicates the operation is not supported by the public key algorithm

Everything works when compiling against OpenSSL 1.1.1-pre6. Looking at diffs, I'm not sure what has changed in pre7?

jakob:
OK, it looks like they quietly removed support for using `EVP_PKEY_verify` with Ed25519 keys in pre7. We need to use `EVP_DigestVerify` instead.

I read the docs, and I think this is how we should do it:

```diff
diff --git a/src/openssl.c b/src/openssl.c
index 96170ba..0941b72 100644
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -2693,20 +2693,20 @@ int
 _libssh2_ed25519_verify(libssh2_ed25519_ctx *ctx, const uint8_t *s,
                         size_t s_len, const uint8_t *m, size_t m_len)
 {
-    int ret = -1;
+    int ret = 0;
 
-    EVP_PKEY_CTX *s_ctx = EVP_PKEY_CTX_new(ctx->public_key, NULL);
+       EVP_MD_CTX *md_ctx = EVP_MD_CTX_new();
+    if (NULL == md_ctx)
+        return -1;
 
-    if(s_ctx) {
-        ret = EVP_PKEY_verify_init(s_ctx);
-        if(ret != 1)
-            return -1;
+    ret = EVP_DigestVerifyInit(md_ctx, NULL, NULL, NULL, ctx->public_key);
+    if (ret != 1)
+        goto cleanup;
 
-        ret = EVP_PKEY_verify(s_ctx, s, s_len, m, m_len);
-    }
+       ret = EVP_DigestVerify(md_ctx, s, s_len, m, m_len);
 
-    if(s_ctx)
-        EVP_PKEY_CTX_free(s_ctx);
+cleanup:
+    EVP_MD_CTX_free(md_ctx);
 
     return (ret == 1) ? 0 : -1;
 }
```

willco007:
@jakob fix looks good, I've applied it to the PR.

jakob:
Can we merge this PR? I haven't reviewed it in detail, but it seems to be working great for my users. Haven't received any SSH related bug reports.

The issues raised by @vszakats have been addressed seeing that it's now based on OpenSSL 1.1.1.

The only thing potentially holding this PR back is that OpenSSL 1.1.1 isn't released yet, but I don't think that matters. Merging this into master sooner than later is definitely better, because then we can make a libssh2 1.9 release soon after OpenSSL 1.1.1 is released.

vszakats:
It LGTM, thanks for all the great work done here! OpenSSL 1.1.1 is indeed slipping, but I agree it doesn't matter on the grand scheme of things and once it will be out, it will be a great combo with libssh2 + this PR. Till then, OpenSSL 1.1.1 betas can fill the gap.

willco007:
OK, if it's all the same to everyone, I'm going to land this commit and we can start thinking about a 1.9 release when OpenSSL 1.1.1 is released. @jakob Thanks for the additional testing and help with this!

tjosten:
Since OpenSSL 1.1.1 is out now, do you plan on releasing this PR with a 1.9 release of libssh2? Thanks!

shaicoleman:
The website needs to be updated to reflect that ED25519 is now supported

willco007:
The website will be updated when 1.9 is released.

erikderzweite:
I don't see ED25519 being mentioned on the website.

Issue #221, opened 2017-10-26
Installation instructions?


Melkor462:
I cannot find instructions on how to install this anywhere. May be useful to include that somewhere.

bagder:
Fixed since #249 was merged.

PR #249, opened 2018-05-02
Add Instructions for building from Master

jakob:
I've previously only built libssh2 from tarballs, and it took me a lot of time to figure out how to build from master. 

Then I discovered that all I needed to do was call `./buildconf`

This pull request updates the build instructions to mention this.

(See also #221)


willco007:
Thanks, I ran into this the last time I tried to build with autotools too. 👍

PR #257, opened 2018-07-20
Add function to provide connected host information

jmbents:
Update to add 'libssh2_channel_read_host_info' function to provide
host information which is connected to remote server.

Signed-off-by: Paresh Chaudhary <paresh.chaudhary@rockwellcollins.com>
Signed-off-by: Jared Bents <jared.bents@rockwellcollins.com>

### DIFF
--- a/docs/libssh2_channel_read_host_info.3
+++ b/docs/libssh2_channel_read_host_info.3
@@ -1,0 +1,36 @@
+.TH libssh2_channel_read_host_info 3 "23 Jan 2019" "libssh2 1.8.0" "libssh2 manual"
+.SH NAME
+libssh2_channel_read_host_info - read host information.
+.SH SYNOPSIS
+#include <libssh2.h>
+
+LIBSSH2_API int libssh2_channel_read_host_info(LIBSSH2_CHANNEL *channel,
+                                               unsigned char *shost,
+                                               size_t shost_buffer_size,
+                                               unsigned int *sport,
+                                               unsigned int *shost_len);
+
+.SH DESCRIPTION
+This function is used to pass shost, sport, and shost_len from library to user space application.
+underlying function \fIlibssh2_channel_read_ex(3)\fP.
+
+\fIchannel\fP - Active channel stream to read from.
+
+\fIshost\fP - Pointer to host to tell the application where the connection originated from.
+
+\fIshost_buffer_size\fP - Size of buffer allocated for shost.
+
+\fIsport\fP - Pointer to port to tell the application where the connection originated from.
+
+\fIshost_len\fP - Pointer to length of shost.
+
+.SH RETURN VALUE
+Return 0 on successful and error values if failed.
+
+.SH ERRORS
+\fILIBSSH2_ERROR_BAD_USE\fP - channel/shost is not initialized.
+
+\fILIBSSH2_ERROR_INVAL\fP - shost_buffer_size is too small and overflow would occur.
+
+.SH SEE ALSO
+N/A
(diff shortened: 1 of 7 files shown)


jmbents:
@bagder Is this something we could get into the next release?

src/channel.c
```diff
@@ -53,6 +53,29 @@
...
+LIBSSH2_API int
+libssh2_channel_read_host_info(LIBSSH2_CHANNEL *channel, unsigned char *shost,
+                               unsigned int *sport, unsigned int *shost_len)
+{
+    int rc = 0;
+
+    if (!channel || !channel->shost)
+        return LIBSSH2_ERROR_BAD_USE;
+
+    memcpy(shost, channel->shost, strlen((char*)channel->shost));
```
bagder: this seems like asking for an overflow. I would prefer the API to pass in a buffer size too to make that impossible. Possibly then return a failure if the buffer was too small.

jmbents: I have updated to add a buffer_size for the user to provide and a check to make sure it is big enough.

willco007:
Could you use the paradigm of calling the function once with NULL to get the required buffer length and call it again with the alloc'd buffer to copy the value in to? That way the client would be much less likely to alloc a buffer that is too small.

docs/libssh2_channel_read_host_info.3
```diff
@@ -0,0 +1,36 @@
...
+                                               unsigned int *sport,
+                                               unsigned int *shost_len);
+
+.SH DESCRIPTION
+This function is used to pass shost, sport, and shost_len from library to user space application.
+underlying function \fIlibssh2_channel_read_ex(3)\fP.
+
+\fIchannel\fP - Active channel stream to read from.
+
+\fIshost\fP - Pointer to host to tell the application where the connection originated from.
```
bagder: Please elaborate on this. Where it originated from? Is that a host name? How can a connection originate from a host name? Is it an IP address? Can it be either IPv4 or IPv6? Is the string zero terminated?

docs/libssh2_channel_read_host_info.3
```diff
@@ -0,0 +1,36 @@
...
+
+\fIshost_buffer_size\fP - Size of buffer allocated for shost.
```
bagder: Is there a recommendation for how large this should be?

docs/libssh2_channel_read_host_info.3
```diff
@@ -0,0 +1,36 @@
...
+
+\fIsport\fP - Pointer to port to tell the application where the connection originated from.
```
bagder: Elaborate. What is a port it originates from. Is this the TCP source port for the connection? If so, why not say so?

docs/libssh2_channel_read_host_info.3
```diff
@@ -0,0 +1,36 @@
...
+
+\fIshost_len\fP - Pointer to length of shost.
```
bagder: And perhaps explain that the function will set this to the length of that returned string? Is that what it does? If the string is zero terminated, why is this information needed?

docs/libssh2_channel_read_host_info.3
```diff
@@ -0,0 +1,36 @@
+.TH libssh2_channel_read_host_info 3 "23 Jan 2019" "libssh2 1.8.0" "libssh2 manual"
+.SH NAME
+libssh2_channel_read_host_info - read host information.
+.SH SYNOPSIS
+#include <libssh2.h>
+
+LIBSSH2_API int libssh2_channel_read_host_info(LIBSSH2_CHANNEL *channel,
```
bagder: how about dropping `read` from the name and just make it `libssh2_channel_hostinfo` ? Is it even info about the *host* ?

PR #264, opened 2018-08-25
Improve sftp writeback performance

djelinski:
With the current implementation libssh2_sftp_write only acknowledges data when it is acknowledged by the sftp server. The proposed implementation acknowledges data as soon as it is put on the wire. This vastly improves write performance for large files on high-latency networks.
With this solution some write errors will only be reported when file handle is closed, and when an error is returned, there's no way to tell how much data was written. This matches the behavior of stdio functions fwrite/fclose.

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -424,11 +424,12 @@ sftp_packet_read(LIBSSH2_SFTP *sftp)
  * Remove all pending packets in the packet_list and the corresponding one(s)
  * in the SFTP packet brigade.
  */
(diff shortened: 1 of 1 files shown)


sbredahl:
Can't we just use a large buffer (say ~1MB), this seems to solve the same problem?

bagder:
I would like to see this work merged but I have not had the bandwidth to deal with it personally. I don't know how to proceed with this.

Jan-E:
See https://github.com/libssh2/libssh2/pull/335

Jan-E:
New PR https://github.com/libssh2/libssh2/pull/337

PR #268, opened 2018-10-23
SFTP: increase maximum packet size to 256K

zhwang-amad:
to match implementations like OpenSSH


willco007:
We also had a user with a directory listing that hit the max length due to many long file names. This should take care of it.

PR #269, opened 2018-10-23
SFTP: discard oversized packet

zhwang-amad:
To allow subsequent reads on the channel.

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -346,10 +346,14 @@ sftp_packet_read(LIBSSH2_SFTP *sftp)
             sftp->partial_len = _libssh2_ntohu32(sftp->partial_size);
             /* make sure we don't proceed if the packet size is unreasonably
                large */
-            if(sftp->partial_len > LIBSSH2_SFTP_PACKET_MAXLEN)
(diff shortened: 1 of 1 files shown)


Zenju:
Just for the record, the default of LIBSSH2_SFTP_PACKET_MAXLEN  800000 is too small.
E.g. during directory traversal AWS's SFTP always returns a single package, no matter how many items the directory contains. So in practice it occasionally hits 100kb package size, which is not an error.
"Fixed" by setting: LIBSSH2_SFTP_PACKET_MAXLEN  160000

willco007:
My quick glance at OpenSSH it would appear it will accept packets up to (256 * 1024) in length, should we go with that value to match them?

zhwang-amad:
There is another pull request proposing to increase maximum packet size to 256K: #268 .
The purpose of this pull request is to discard packets larger than whatever size permitted to not block the caller.

bagder:
I suppose we should still merge this right, even if the max is now much larger?

zhwang-amad:
Yes! This pull request allows the application to continue working with the channel after receiving an oversized frame.

PR #274, opened 2018-11-27
New function to retrieve pre-auth server banner

dmiller-nmap:
This is currently working privately to collect `SSH_MSG_USERAUTH_BANNER` contents via a NSE script for Nmap.

Reference: [RFC 4252, section 5.4](https://tools.ietf.org/html/rfc4252#section-5.4)

### DIFF
--- a/src/userauth.c
+++ b/src/userauth.c
@@ -53,6 +53,7 @@
 #include "session.h"
 #include "userauth.h"
 
+#define LIBSSH2_USERAUTH_MAX_BANNER 2048
 /* libssh2_userauth_list
  *
  * List authentication methods
@@ -63,8 +64,8 @@
 static char *userauth_list(LIBSSH2_SESSION *session, const char *username,
                            unsigned int username_len)
 {
-    static const unsigned char reply_codes[3] =
-        { SSH_MSG_USERAUTH_SUCCESS, SSH_MSG_USERAUTH_FAILURE, 0 };
+    static const unsigned char reply_codes[4] =
+        { SSH_MSG_USERAUTH_SUCCESS, SSH_MSG_USERAUTH_FAILURE, SSH_MSG_USERAUTH_BANNER, 0 };
     /* packet_type(1) + username_len(4) + service_len(4) +
        service(14)"ssh-connection" + method_len(4) = 27 */
     unsigned long methods_len;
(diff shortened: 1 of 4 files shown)


dmiller-nmap:
I had to guess on an upper safe limit for how  much memory to allocate for this banner (`LIBSSH2_USERAUTH_MAX_BANNER`), since we obviously shouldn't just trust the server not to spoof a huge number (4-byte length = 4.3GB). US Dept of Defense standard warning banner is probably about the longest you'll see at 1484 bytes, so 2KB is probably plenty.

sbredahl:
That could prove useful, thanks!

MOJNICK:
I think this functionality should be moved from userauth_list() and added to every authentication method (password, publickey, hostbased)  instead. btw. is the do-while necessary?

dmiller-nmap:
It would be nice to be able to have the banner available no matter what auth was attempted, since the spec says:

>The SSH server may send an SSH_MSG_USERAUTH_BANNER message at any
   time after this authentication protocol starts and before
   authentication is successful.

I chose to implement it most simply as part of `userauth_list` since that's a common way that SSH clients open a connection. I don't know if putting it into every userauth attempt would have made the change more difficult to review (it's been a while since I looked at the code now!).

The do/while loop is used in order to handle the possibility of multiple banners sent in one reply, as well as to loop back and grab the next message (userauth success or failure) after the banner. There's probably a better way to do this like dispatching based on the type of message extracted, but this worked at the moment.

src/userauth.c
```diff
@@ -118,21 +119,54 @@ static char *userauth_list(LIBSSH2_SESSION *session, const char *username,
...
+                _libssh2_error(session, LIBSSH2_ERROR_EAGAIN,
+                               "Would block requesting userauth list");
+                return NULL;
+            }
+            else if(rc) {
+                _libssh2_error(session, rc, "Failed getting response");
+                session->userauth_list_state = libssh2_NB_state_idle;
+                return NULL;
+            }
+            if (session->userauth_list_data[0] == SSH_MSG_USERAUTH_BANNER) {
```
willco007: Does not check userauth_list_data_len before accessing first byte of data.

src/userauth.c
```diff
@@ -118,21 +119,54 @@ static char *userauth_list(LIBSSH2_SESSION *session, const char *username,
...
+                methods_len = _libssh2_ntohu32(session->userauth_list_data + 1);
```
willco007: Does not check to see if length is > 5 before accessing next 4 bytes.

willco007:
I've added a couple comments to this code. There is a PR coming which resolves both of these issues but you'll need to rebase this feature after that fix is landed.

thesubtlety:
This feature is just what I'm looking for. Are there any plans to revisit this issue? Thanks!

PR #275, opened 2018-11-30
Fix error compiling on Win32 with STDCALL=ON

alex-weaver:
This is necessary for compiling libgit2 with SSH support on 32-bit windows


willco007:
Thank you for your contribution.

PR #280, opened 2019-01-11
mbedtls: Fix leak of 12 bytes by each key exchange.

AdrianMoranMontes:
Some of the leaks present on version 1.8.0 have been solved, but still there was one more leak.

Running tests before/after the patch can verify that the leak is solved with this simple modification.

Best regards,
Adrian M.

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -688,8 +688,7 @@ _libssh2_dh_secret(_libssh2_dh_ctx *dhctx, _libssh2_bn *secret,
 void
 _libssh2_dh_dtor(_libssh2_dh_ctx *dhctx)
 {
-    mbedtls_mpi_free(*dhctx);
-    mbedtls_free(*dhctx);
+    _libssh2_bn_free(*dhctx);
     *dhctx = NULL;
 }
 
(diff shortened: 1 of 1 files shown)


willco007:
Thanks for the patch. Looking at the code a bit, `dhctx` is being init'd using `_libssh2_mbedtls_bignum_init()`, so the dtor call should use `_libssh2_mbedtls_bignum_free()` which already includes the call to `mbedtls_free()`. If you'd update the patch that'd be super.

src/mbedtls.c
```diff
@@ -688,8 +688,7 @@ _libssh2_dh_secret(_libssh2_dh_ctx *dhctx, _libssh2_bn *secret,
 void
 _libssh2_dh_dtor(_libssh2_dh_ctx *dhctx)
 {
-    mbedtls_mpi_free(*dhctx);
-    mbedtls_free(*dhctx);
+    _libssh2_bn_free(*dhctx);
```
willco007: Unfortunately is the wrong call, it needs to be `_libssh2_mbedtls_bignum_free()` which calls `mbedtls_mpi_free()` as well as `mbedtls_free()`.

AdrianMoranMontes: Sorry, my bad.
The change was so easy that I did not try to compile it again. The funny thing is that in the commit comment, I specified the correct function... :)

Now it is correct and I have check that it compiles correctly.

Regards,
Adrian M.

PR #281, opened 2019-01-16
Should not clean up agent if agent is NULL pointer

ericcurtin:
Dereference after NULL check


willco007:
Thanks for bringing this to our attention. Instead of adding another goto declaration I will add a NULL agent check around the disconnect/free calls.

PR #284, opened 2019-02-06
Fix https://github.com/libssh2/libssh2/pull/271

Zenju:
Otherwise the previous "LIBSSH2_SFTP_DEFAULT_MODE" won't have any effect.


willco007:
Good catch, when I brought my change over from my private branch I missed deleting this line.

PR #304, opened 2019-03-04
_libssh2_string_buf_free: use correct free

bagder:
Use LIBSSH2_FREE() here, not free(). We allow memory function
replacements so free() is rarely the right choice...


bagder:
Without this fix, a debug-build of curl segfaults immediately since it uses the custom memory function to debug and track memory use/leaks.

willco007:
Good catch, thanks!

PR #311, opened 2019-03-11
allow private-key from memory using mbedtls

doublex:
And fix passphrase

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -417,17 +417,21 @@ _libssh2_mbedtls_rsa_new_private_frommemory(libssh2_rsa_ctx **rsa,
     if(*rsa == NULL)
         return -1;
 
-    mbedtls_pk_init(&pkey);
-
     /*
     Hack: mbedtls checks in "mbedtls/pkparse.c:1184" if "key[keylen - 1] != '\0'"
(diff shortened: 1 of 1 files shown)


src/mbedtls.c
```diff
@@ -411,15 +411,24 @@ _libssh2_mbedtls_rsa_new_private_frommemory(libssh2_rsa_ctx **rsa,
...
     mbedtls_pk_init(&pkey);
 
-    ret = mbedtls_pk_parse_key(&pkey, (unsigned char *)filedata,
-                              filedata_len, NULL, 0);
+    /*
+    Hack: mbedtls checks in "mbedtls/pkparse.c:1184" if "key[keylen - 1] != '\0'"
+    Using a private-key from memory will fail if the last byte is not a null terminator
+    */
+    filedata_null = mbedtls_calloc(1, filedata_len + 1);
+    memcpy(filedata_null, filedata, filedata_len);
```
willco007: There is no null check here before calling `memcpy` and `mbedtls_free`, probably worth adding. Same thing is replicated on line 652.

doublex:
See https://github.com/libssh2/libssh2/pull/358

PR #324, opened 2019-03-17
code style cleanup

bagder:
- add tool for verifying code style
- clean up code to adhere to code style
- add travis job that verifies the code style


bagder:
I'd be useful to also provide a `make checksrc` equivalent for the cmake build but I don't know how.

PR #326, opened 2019-03-18
String buffer improvements

willco007:
Changed `_libssh2_get_c_string` to `_libssh2_get_string` and now returns -1 on error and 0 on success. Also changed `_libssh2_get_bignum_bytes` to return -1 on error and 0 on success. Both now take `size_t` as a parameter to return the length of the buffer. This is to attempt to keep the data going in/out of these functions unsigned to better impedance match the existing API.

### DIFF
--- a/src/hostkey.c
+++ b/src/hostkey.c
@@ -65,7 +65,7 @@ hostkey_method_ssh_rsa_init(LIBSSH2_SESSION * session,
 {
     libssh2_rsa_ctx *rsactx;
     unsigned char *e, *n;
-    int e_len, n_len;
+    size_t e_len, n_len;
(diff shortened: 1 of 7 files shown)


bagder:
Will, please don't merge master into your branch. Rebase and force-push instead so that we then can just merge this fast-forward without merge commits. Makes the history and commit log much easier and simpler.

bagder:
I would also suggest you run the checksrc script locally before you push, as it will save you some roundtrips...

willco007:
@bagder Sorry about the merge; working on formatting at the moment.

willco007:
Ugh, I just merged it again. Starting over.

PR #327, opened 2019-03-19
Fixed misapplied patch

willco007:
Fixes for user auth failure on the 1.8.1 patched branch.


pghmcfc:
Works for me (test suite passes) on all Fedora releases from 13 to Rawhide (i686 and x86_64), and RHEL 6 and 7 (i686 and x86_64), and RHEL 8 Beta (x86_64).

kloczek:
FYI: looks like 1.8.2 is OK now :)

PR #331, opened 2019-03-20
Fix Error C2371 'ssize_t': redefinition; different basic types

Zenju:
This continues https://github.com/libssh2/libssh2/pull/256

Sorry, I believe I messed up by deleting my libssh2 branch, so I wasn't able to make changes to my prior pull request. I'm not (yet) a git expert...

>I would however maybe suggest that we could do the same courtesy and also define that when we do. What do you think @Zenju?

Sure, I'll add that.


bagder:
Thanks!

PR #332, opened 2019-03-20
String buffer improvements

willco007:
Renamed `_libssh2_get_c_string` to `_libssh2_get_string` and now returns -1 on error and 0 on success. Also changed `_libssh2_get_bignum_bytes` to return -1 on error and 0 on success. Both now take `size_t` as a parameter to return the length of the buffer. This is to attempt to keep the data going in/out of these functions unsigned to better impedance match the existing API.

### DIFF
--- a/src/hostkey.c
+++ b/src/hostkey.c
@@ -65,7 +65,7 @@ hostkey_method_ssh_rsa_init(LIBSSH2_SESSION * session,
 {
     libssh2_rsa_ctx *rsactx;
     unsigned char *e, *n;
-    int e_len, n_len;
+    size_t e_len, n_len;
     struct string_buf buf;
 
     if(*abstract) {
@@ -86,12 +86,10 @@ hostkey_method_ssh_rsa_init(LIBSSH2_SESSION * session,
     if(_libssh2_match_string(&buf, "ssh-rsa") != 0)
         return -1;
 
-    e_len = _libssh2_get_c_string(&buf, &e);
-    if(e_len <= 0)
+    if(_libssh2_get_string(&buf, &e, &e_len) != 0)
         return -1;
 
-    n_len = _libssh2_get_c_string(&buf, &n);
-    if(n_len <= 0)
+    if(_libssh2_get_string(&buf, &n, &n_len) != 0)
         return -1;
 
     if(_libssh2_rsa_new(&rsactx, e, e_len, n, n_len, NULL, 0,
@@ -285,7 +283,7 @@ hostkey_method_ssh_dss_init(LIBSSH2_SESSION * session,
 {
     libssh2_dsa_ctx *dsactx;
     unsigned char *p, *q, *g, *y;
-    int p_len, q_len, g_len, y_len;
+    size_t p_len, q_len, g_len, y_len;
     struct string_buf buf;
 
     if(*abstract) {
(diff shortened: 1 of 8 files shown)


src/hostkey.c
```diff
@@ -308,20 +306,16 @@ hostkey_method_ssh_dss_init(LIBSSH2_SESSION * session,
     if(_libssh2_match_string(&buf, "ssh-dss") != 0)
         return -1;
 
-    p_len = _libssh2_get_c_string(&buf, &p);
-    if(p_len < 0)
+    if(_libssh2_get_string(&buf, &p, &p_len) != 0)
```
bagder: personally, I'd prefer these lines without the `!= 0` part. Like this:
~~~c
if(_libssh2_get_string(&buf, &p, &p_len))
    return -1;

src/libssh2_priv.h
```diff
@@ -619,7 +619,7 @@ struct _LIBSSH2_SESSION
      * Or read from server in (eg) KEXDH_INIT (for client mode)
      */
     unsigned char *server_hostkey;
-    uint32_t server_hostkey_len;
+    size_t server_hostkey_len;
```
bagder: This is unsigned 32 bit data from the wire. I think it would be wise to keep the source size here, It can't be larger anyway!

src/misc.c
```diff
@@ -815,12 +822,15 @@ int _libssh2_get_bignum_bytes(struct string_buf *buf, unsigned char **outbuf)
     buf->dataptr += data_len;
     buf->offset += data_len;
 
-    return bn_len;
+    if (outlen != NULL)
```
bagder: Again, I think `!= NULL` here only makes this harder to read.

src/misc.c
```diff
@@ -815,12 +822,15 @@ int _libssh2_get_bignum_bytes(struct string_buf *buf, unsigned char **outbuf)
...
+        *outlen = (size_t)bn_len;
+
+    return 0;
 }
 
 int _libssh2_check_length(struct string_buf *buf, size_t len)
 {
-    return ((int)(buf->dataptr - buf->data) <= (int)(buf->len - len)) ? 1 : 0;
+    return ((ssize_t)(buf->dataptr - buf->data) <= (ssize_t)(buf->len - len)) ? 1 : 0;
```
bagder: shouldn't this function then also return a `ssize_t` ? Seems odd to just change the typecasts.

src/kex.c
```diff
@@ -2818,11 +2818,11 @@ curve25519_sha256(LIBSSH2_SESSION *session, unsigned char *data,
         }
 
         /* server signature */
-        if(_libssh2_get_string(&buf, &exchange_state->h_sig, &(exchange_state->h_sig_len)) != 0) {
+        if(_libssh2_get_string(&buf, &exchange_state->h_sig,
+           &(exchange_state->h_sig_len)) != 0) {
```
bagder: weird indent? the second line is expected to start under the '&'  from the previous line

src/hostkey.c
willco007: I'm fine with that and can make these changes. I typically use explicit value checks to ensure the caller knows what the return types are.

src/misc.c
willco007: It returns 1 or 0 regardless of free space, cleaner API IMHO. The casts are to allow larger values (but still signed), in reality these upper limits should never bit hit.

willco007:
The latest build failure doesn't seem related to the code.

src/misc.c
bagder: It'd be cool with a comment above `_libssh2_check_length()` explaining its use/purpose...

PR #333, opened 2019-03-21
make sftp_read() return 0 on zero bytes read

### LINKED ISSUE #312: Libssh2_sftp_read returns LIBSSH2_ERROR_SFTP_PROTOCOL when reading a file of size 0
Libssh2_sftp_read returns LIBSSH2_ERROR_SFTP_PROTOCOL when reading a file of size 0

doorsdown:
This is in response to #312 where files with size 0 respond with the invalid protocol error.

Fixes #312

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -679,7 +679,6 @@ sftp_bin2attr(LIBSSH2_SFTP_ATTRIBUTES *attrs, const unsigned char *p,
     buf.data = (unsigned char *)p;
     buf.dataptr = buf.data;
     buf.len = data_len;
-    buf.offset = 0;
 
     if(_libssh2_get_u32(&buf, &flags) != 0) {
         return LIBSSH2_ERROR_BUFFER_TOO_SMALL;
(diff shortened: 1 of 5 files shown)


src/sftp.c
```diff
@@ -1554,7 +1554,7 @@ static ssize_t sftp_read(LIBSSH2_SFTP_HANDLE * handle, char *buffer,
             if(chunk->lefttosend) {
                 /* if the chunk still has data left to send, we shouldn't wait
                    for an ACK for it just yet */
-                if(bytes_in_buffer > 0) {
+                if(bytes_in_buffer >= 0) {
```
bagder: Pointed out by the CI build: `bytes_in_buffer` is a `size_t` (unsigned). It will *always* be larger than or equal to zero...

doorsdown:
Right, and looking at this more I wonder if bytes_in_buffer should be made a ssize_t and initialized at -1 or leave bytes_in_buffer as a size_t and remove the is it greater than 0 check.

src/sftp.c
doorsdown: Right, and looking at this more I wonder if bytes_in_buffer should be made a ssize_t and initialized at -1 or leave bytes_in_buffer as a size_t and remove the is it greater than 0 check.

PR #334, opened 2019-03-21
style: make includes and examples code style strict

bagder:
make travis and the makefile rule verify them too


bagder:
travis is clearly having issues right now so lots of tests are timing out. I'll just wait a while and try to manually restart them later...

PR #335, opened 2019-03-21
Implement sftp writeback for 1.8.1

Jan-E:
@bagder @djelinski
As far as I know, this is the https://github.com/libssh2/libssh2/pull/264 PR for 1.8.1.

I adapted it a long time ago as a patch for the 1.8.0 release and updated it yesterday as a patch for the 1.8.1 branch. It should be carefully checked if I did not miss anything, because it was a lot of copy-paste.

I did not test it yet, but it might be a new start.

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -432,11 +432,12 @@ sftp_packet_read(LIBSSH2_SFTP *sftp)
  * Remove all pending packets in the packet_list and the corresponding one(s)
  * in the SFTP packet brigade.
  */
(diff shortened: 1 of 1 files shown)


bagder:
Don't use the 1.8.1 branch, that one will go nowhere. We will do the next release off master.

Jan-E:
> Don't use the 1.8.1 branch, that one will go nowhere. We will do the next release off master.

OK, I will try another PR for master

Jan-E:
New PR https://github.com/libssh2/libssh2/pull/337

PR #343, opened 2019-03-25
misc: remove 'offset' from string_buf

bagder:
It isn't necessary.


willco007:
Looks fine from here.

Issue #287, opened 2019-02-13
banner_receive() may leak memory if a client attempts  connection to  non-ssh2 servers/portts.


ismail-yilmaz:
Hello,

banner_receive() function in session.c may leak memory if the client accidentally attempts connection to non-ssh2 servers or ports.

Let's say that you accidentally try to connect to an FTP port (21 instead of 22), the libssh2 parser will search for "SSH-" characters in the server response: 

https://github.com/libssh2/libssh2/blob/03d068f00a28ee23497f1b6b1206b7857d48d06d/src/session.c#L726

And usually these characters are included in the, say, FTP error messages too (e.g. "Unknown command SSH-xxx bla bla..."

In such cases, banner_receive() will continue to allocate memory (session.c, ln, 166-182):

```
    /* From this point on, we are done here */
    session->banner_TxRx_state = libssh2_NB_state_idle;
    session->banner_TxRx_total_send = 0;

    if(!banner_len)
        return LIBSSH2_ERROR_BANNER_RECV;

    session->remote.banner = LIBSSH2_ALLOC(session, banner_len + 1);
    if(!session->remote.banner) {
        return _libssh2_error(session, LIBSSH2_ERROR_ALLOC,
                              "Error allocating space for remote banner");
    }
    memcpy(session->remote.banner, session->banner_TxRx_banner, banner_len);
    session->remote.banner[banner_len] = '\0';
    _libssh2_debug(session, LIBSSH2_TRACE_TRANS, "Received Banner: %s",
                   session->remote.banner);
return LIBSSH2_ERROR_NONE;
```
And this causes heap leaks.

EDIT: Just to be clear: This happens because some servers return the client banner  as an unknown/unrecognized command (libssh2.h, ln: 229:

`#define LIBSSH2_SSH_BANNER                  "SSH-2.0-libssh2_" LIBSSH2_VERSION`

willco007:
This is resolved with #356, thanks for the report.

PR #345, opened 2019-03-26
Prevent banner_receive() from leaking when accessing non ssh ports

doorsdown:
This pr is based on #287


willco007:
Your updated PR isn't quite right, it includes changes already on master. The easiest solution is probably just to close this PR and re-open a new one after you've pulled up and re-applied your fix.

doorsdown:
Opened #356

PR #347, opened 2019-04-01
1.8.x

Giszpenc:
Fix signature for when LIBSSH2_CRYPT_NONE is defined

### DIFF
--- a/RELEASE-NOTES
+++ b/RELEASE-NOTES
@@ -1,31 +1,29 @@
-libssh2 1.8.0
-
-This release includes the following changes:
-
- o added a basic dockerised test suite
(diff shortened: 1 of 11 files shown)


willco007:
This PR isn't quite right. You're trying to push changes from the 1.8.x branch into master which isn't going to work. What you'll want to do is clone master and make your changes on a fork and then submit a PR from there.

Giszpenc:
> This PR isn't quite right. 

Thank you for your patience and tolerance.  As is probably obvious, I am a novice with good intentions -- a dangerous combination!  I will continue to try and submit my fix.  It may even change as I have discovered it just compiles but fails during key exchange.

PR #350, opened 2019-04-03
Simplified _libssh2_check_length

willco007:
misc.c : _libssh2_check_length()

Removed cast and one-lined check. 

Credit : Yuriy M. Kaminskiy

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -811,10 +811,7 @@ int _libssh2_get_bignum_bytes(struct string_buf *buf, unsigned char **outbuf)
 
 int _libssh2_check_length(struct string_buf *buf, size_t len)
 {
-    if(len > buf->len)
-        return 0;
-
-    return ((int)(buf->dataptr - buf->data) <= (int)(buf->len - len)) ? 1 : 0;
+    return (len <= (size_t)((buf->data + buf->len) - buf->dataptr));
 }
 
 /* Wrappers */
--- a/src/misc.c
+++ b/src/misc.c
@@ -811,7 +811,9 @@ int _libssh2_get_bignum_bytes(struct string_buf *buf, unsigned char **outbuf)
 
 int _libssh2_check_length(struct string_buf *buf, size_t len)
 {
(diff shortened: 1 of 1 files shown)


src/misc.c
```diff
@@ -811,10 +811,7 @@ int _libssh2_get_bignum_bytes(struct string_buf *buf, unsigned char **outbuf)
 
 int _libssh2_check_length(struct string_buf *buf, size_t len)
 {
-    if(len > buf->len)
-        return 0;
-
-    return ((int)(buf->dataptr - buf->data) <= (int)(buf->len - len)) ? 1 : 0;
+    return (len <= (size_t)((buf->data + buf->len) - buf->dataptr));
```
kdudka: This weakens the check in case `buf->dataptr` is already behind `buf->data + buf->len` because when `((buf->data + buf->len) - buf->dataptr)` is negative, the conversion to `size_t` turns it into a big positive number.

willco007: It theoretically does, but `dataptr` should be greater than or equal to data if the API is used correctly (famous last words, I know).  Also, in the original case if `dataptr` was less than `data` it would be a negative number which would be less than the test and incorrectly return true, which is also bad. Furthermore, the cast to a signed value from unsigned isn't great and could loose precision.

bagder: I think the expression could be split up in several parts to become more readable and then the logic is easier to follow and confirm. Something like:
~~~c
 char *endp = &buf->data[buf->len];
 size_t left = endp - buf->dataptr;
 return (len <= left);
~~~
it could even protect against the wrap-around case @kdudka mentioned:
~~~c
 char *endp = &buf->data[buf->len];
 size_t left = endp - buf->dataptr;
 return ((len <= left) && (left <= buf->len));
~~~

willco007: Looks good to me. I'll go ahead and make the change to @bagder's last suggestion.

carnil:
https://blog.semmle.com/libssh2-integer-overflow/ is related to this pull request.

Issue #299, opened 2019-02-23
Memory leak  in `libssh2_session_handshake()`


dimmaq:
In my case, if set little timeout `libssh2_session_set_timeout(session, 5000)` then `libssh2_session_handshake()` failed with last error **(-9) API timeout expired** and have memory leak

`diffie_hellman_sha1()`
allocated
https://github.com/libssh2/libssh2/blob/f7dd7d1d434d73baa44a1e560232df9196aa2dd7/src/kex.c#L165-L166

return without freeing
https://github.com/libssh2/libssh2/blob/f7dd7d1d434d73baa44a1e560232df9196aa2dd7/src/kex.c#L234-L242

tested on example ssh2.c
```
libssh2_session_set_timeout(session, 5000)
    if (libssh2_session_handshake(session, sock)) {
		char *errmsg;
		int errlen;
		int err = libssh2_session_last_error(session, &errmsg, &errlen, 0);
		fprintf(stderr, "Unable to open a session: (%d) %s\n", err, errmsg);
        fprintf(stderr, "Failure establishing SSH session\n");
		system("pause");
        return -1;
    }
```

tiennou:
I've been toying at this one, and I caught another related leak. Sadly I forgot to grab the actual leakstack (a leak of the negotiated ecdsa key IIRC, starting under `session->kex->exchange_keys(…);`), but it seems to be the same issue: a timeout in the middle of a session being setup can leak parts of the setup done by `_libssh2_kex_exchange`.

doublex:
@tiennou 
Did you try this provisionally patch?
https://github.com/libssh2/libssh2/pull/357

tiennou:
Yes, and have the same opinion as @willco007: it's not quite the correct fix, and a more complete solution is needed (eg. you're only fixing the issue on timeouts — anything else that can cause the code to bail and never retry is still going to leak). Right now I'm still confused by the async state-machine code, but my gut feeling is that `session_create` needs a counter "destructor" that handles half-negotiated KEXes correctly, so if the code detects a timeout, it can tear down all transient data that would be overwritten on a retry. And the async wrapper macro (`BLOCK_ADJUST`?) need to take that destructor.

It sounds complex 😉.

doorsdown:
I tend to agree that my patch is very narrowly focused and a more complete fix would be best but @tiennou said, its complex.

PR #357, opened 2019-04-18
Fixing bug where if timeout was reached the memory was not cleand up

doorsdown:
While researching #299 I found that the LIBSSH2_ERROR_TIMEOUT value was being set in the session but being ignored. I added @dimmaq's test into the ssh2 example and was able to see memory being lost in valgrind and this no longer happening after I compiled this fix in.


doublex:
@bagder  @willco007
This patch fixed the memory-leaks reported by `address sanitizer`.
Could you merge this patch?

willco007:
This doesn't seem like quite the right solution. The value coming back from `_libssh2_kex_exchange` should be where the fix lands, not sniffing `session->err_code` directly. I'll dig into this a bit more and see what I can come up with.

PR #358, opened 2019-04-19
mbedtls: allow private-key from memory

doublex:
Updated https://github.com/libssh2/libssh2/pull/311

This patch fixes two issues using mbedtls:
a) private-key from memory
b) password encrypted private-key (again from memory)

I am using this patch since years.
Is there a chance to apply this patch to libssh2?


willco007:
@doublex could you fix the linter issues found by Travis? After it passes it looks good to go.

doublex:
https://github.com/libssh2/libssh2/pull/359

PR #359, opened 2019-04-19
mbedtls: allow private-key from memory

doublex:
@willco007 

Thanks a lot!
The "trailing whitespaces" from https://github.com/libssh2/libssh2/pull/358


willco007:
Looks like you still have some lines that are over 80 characters so the linter is failing. Also, there is no need to close this PR, just update your branch and it will automatically update the PR. :)

doublex:
@willco007 
Thanks a lot.
I have added a variable to store the "password-length"

doublex:
You simply add a another git-commit?

willco007:
Yup, just push another update to the same branch and it will be looped in automatically.

doublex:
@willco007 
Thanks for your help.

Issue #179, opened 2017-05-07
Various errors when unable to open Windows CNG providers


jethrogb:
When on Windows for some reason, the CNG providers can't be opened, various strange code paths with poor error messaging are taken.

1. `BCryptOpenAlgorithmProvider` can fail (as it does on my system for some reason). Return values are not checked and later crypto code will just try to use the NULL handles and fail.
https://github.com/libssh2/libssh2/blob/386e012/src/wincng.c#L216

2. With wincng support enabled, if you call `libssh2_init` with `LIBSSH2_INIT_NO_CRYPTO`, there is absolutely no way you can make the library work, since `_libssh2_wincng_init` will never get filled in. This manifests the same way as point no. 1 above.

3. In `diffie_hellman_sha1`, if `libssh2_dh_key_pair` fails (such as when no random numbers can be generated due to the previous bug), the function returns without setting an error code. The handshake will fail later without any error message once the client tries to request ssh-userauth without having done a key exchange.
https://github.com/libssh2/libssh2/blob/b3f9557/src/kex.c#L138-L139 introduced in 73493c6c78f13ca0bab30eaadd7c45306d540f1a

jethrogb:
I don't think this is fixed

willco007:
Any PRs to fix Windows CNG issues would be appreciated.

tiennou:
I can take a look soonish. Is there some other things that would be helpful to get 1.9 out ? Since (understandably) my far-reaching PRs can't land just before a long-needed release, I'm keen on keeping the ball rolling 😉.

tiennou:
Do you think [this](https://github.com/tiennou/libssh2/commit/a86ff16ca1f984706ca5e87322dfa75f51c1320e) would be acceptable as a 1st commit, or is it already too much ?

jethrogb:
@willco007 @tiennou is it safe to assume either of you have a Windows development environment setup? If so, could you take a look at https://github.com/libssh2/libssh2/pull/178 ?

willco007:
I'm on macOS so I'm no help I'm afraid.

PR #369, opened 2019-05-14
Report cryptographic initialization failures

tiennou:
This should alleviate the issues in #179, by actually returning a failure if any part of a backend failed to initialize. There should be a more precise way to do that, in case some of them are optional, but hopefully this is sufficient as a first step (I didn't want to touch anything else, because of the generic-backend changes).

I'm bringing in some of my other commits, as some of them are only developer-friendly. They should be mostly NFC, but if you feel they're not needed I'll keep them separate. They make access to `checksrc.pl` easier (`./ci/checksrc.sh`), move the iovec selection support out of crypto, and adds all backends source so they can be accessed via Xcode/VisualStudio.


tiennou:
Rebased, I've ended up removing the "dev" changes because it broke the mbedTLS detection.

PR #371, opened 2019-05-18
acinclude.m4: add mbedtls to LIBS

ffontaine:
This is useful for static builds so that the Libs.private field in
libssh2.pc contains correct info for the benefit of pkg-config users.
Static link with libssh2 requires this information.

Signed-off-by: Baruch Siach <baruch@tkos.co.il>
[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/libssh2/0002-acinclude.m4-add-mbedtls-to-LIBS.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -441,6 +441,7 @@ m4_case([$1],
 [mbedtls], [
   LIBSSH2_LIB_HAVE_LINKFLAGS([mbedcrypto], [], [#include <mbedtls/version.h>], [
     AC_DEFINE(LIBSSH2_MBEDTLS, 1, [Use $1])
+    LIBS="$LIBS -lmbedcrypto"
     found_crypto="$1"
     support_clear_memory=yes
   ])


tiennou:
I'm not familiar with the intricacies of Linux library locating, especially embedded ones, but I'm curious how you're working around the fact that [AFAIK](https://github.com/ARMmbed/mbedtls/pull/1753), mbedTLS does not provide a pkgconfig file. Doesn't adding a `Libs.private` entry means that there should be a corresponding pkgconfig file ?

ffontaine:
Nope, adding -lmbedcrypto to Libs.private does not imply that we need a pkgconfig file. Programs linking with libssh2 through pkgconfig will just add -lmbedcrypto when statically linking with libssh2.

tiennou:
I've re-read through the PR I linked, and I think that was about `Require`/`Require.private`, not `Libs`. As I said, not that well-versed 😉.

👍 from me then.

ffontaine:
Hello, could someone merge this PR?

PR #386, opened 2019-07-04
userauth: fix off by one error when loading public keys with no id

tiennou:
Caught by ASAN:

```
=================================================================
==73797==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60700001bcf0 at pc 0x00010026198d bp 0x7ffeefbfed30 sp 0x7ffeefbfe4d8
READ of size 69 at 0x60700001bcf0 thread T0
2019-07-04 08:35:30.292502+0200 atos[73890:2639175] examining /Users/USER/*/libssh2_clar [73797]
    #0 0x10026198c in wrap_memchr (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x1f98c)
    #1 0x1000f8e66 in file_read_publickey userauth.c:633
    #2 0x1000f2dc9 in userauth_publickey_fromfile userauth.c:1513
    #3 0x1000f2948 in libssh2_userauth_publickey_fromfile_ex userauth.c:1590
    #4 0x10000e254 in test_userauth_publickey__ed25519_auth_ok publickey.c:69
    #5 0x1000090c3 in clar_run_test clar.c:260
    #6 0x1000038f3 in clar_run_suite clar.c:343
    #7 0x100003272 in clar_test_run clar.c:522
    #8 0x10000c3cc in main runner.c:60
    #9 0x7fff5b43b3d4 in start (libdyld.dylib:x86_64+0x163d4)

0x60700001bcf0 is located 0 bytes to the right of 80-byte region [0x60700001bca0,0x60700001bcf0)
allocated by thread T0 here:
    #0 0x10029e053 in wrap_malloc (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x5c053)
    #1 0x1000b4978 in libssh2_default_alloc session.c:67
    #2 0x1000f8aba in file_read_publickey userauth.c:597
    #3 0x1000f2dc9 in userauth_publickey_fromfile userauth.c:1513
    #4 0x1000f2948 in libssh2_userauth_publickey_fromfile_ex userauth.c:1590
    #5 0x10000e254 in test_userauth_publickey__ed25519_auth_ok publickey.c:69
    #6 0x1000090c3 in clar_run_test clar.c:260
    #7 0x1000038f3 in clar_run_suite clar.c:343
    #8 0x100003272 in clar_test_run clar.c:522
    #9 0x10000c3cc in main runner.c:60
    #10 0x7fff5b43b3d4 in start (libdyld.dylib:x86_64+0x163d4)

[...]

PR #392, opened 2019-07-23
Add support for an OSS Fuzzer fuzzing target

cmeister2:
This adds support for an OSS-Fuzz fuzzing target in ssh2_client_fuzzer,
which is a cut down example of ssh2.c. Future enhancements can improve
coverage.

---

I've tested this with the oss-fuzz framework and it appears to find repeated READ access bugs in one particular function, so I think it's definitely worth getting in!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,7 @@ env:
    - ADDRESS_SIZE=32 CRYPTO_BACKEND=mbedTLS BUILD_SHARED_LIBS=ON ENABLE_ZLIB_COMPRESSION=OFF B=cmake
    - ADDRESS_SIZE=32 CRYPTO_BACKEND=mbedTLS BUILD_SHARED_LIBS=OFF ENABLE_ZLIB_COMPRESSION=ON B=cmake
    - ADDRESS_SIZE=32 CRYPTO_BACKEND=mbedTLS BUILD_SHARED_LIBS=ON ENABLE_ZLIB_COMPRESSION=ON B=cmake
+   - B=fuzzer
 
 before_install:
   - if [ $ADDRESS_SIZE = '32' ]; then sudo dpkg --add-architecture i386; fi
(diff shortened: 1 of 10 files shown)


tiennou:
Thanks for working on this, I had prodded oss-fuzz around but decided that more tests would be most helpful (and I've never done anything fuzzing-related yet so…) !

The read bugs you're seeing could be the ones from #386 (hopefully). Other possible ennhancements would be to wire your fuzzing socket to the test harness in #243, so you get those systems as well. I have a branch on top of it with more partially-written tests. Another low-hanging fruit would be the `SSH_AGENT_*` socket.

I can't do much more because I've only been involved in this project recently, so those are only suggestions 😉.

cmeister2:
> The read bugs you're seeing could be the ones from #386 (hopefully). 

Unfortunately not! I reported them to the security DL last week though :smile:

willco007:
@cmeister2 Thanks for all your work on this! There is a pending PR to overhaul the test system; would you mind updating this after that is landed?

cmeister2:
Sure, can you ping me when that happens?

cmeister2:
Ok, so just wondering what the status on this is. We're currently getting overnight issues off a branch I'm maintaining (and @willco007 is doing a stellar job of fixing them), but it'd be good to get this code into master if possible.

cmeister2:
(Or if you didn't want to get it into master, you could create a new libssh2-fuzzer repository or something and give me write access to that; similar to https://github.com/curl/curl-fuzzer/ or https://github.com/GNUAspell/aspell-fuzz/ )

willco007:
There is still work that needs to be done on #243 before it can be landed. I don't think it's necessary to create a new repro just for the fuzzer; but I could be convinced if others feel strongly about it.

cmeister2:
Think I'm still waiting on a merge here which @willco007 seemed to want to move forward with, as per https://github.com/libssh2/libssh2/issues/393#issuecomment-545070770

willco007:
I'm going to land this if there is no objections, it's been sitting here for a while and in important to get integrated.

cmeister2:
Still waiting... :smile:

Issue #388, opened 2019-07-10
Login to ssh server fails using wincng


steffenb7333:
**Describe the bug**
Since the May'19 update of Windows (1903) a login to remote SSH servers using WinCNG is not possible anymore. All my servers, running different versions of OpenSSH reject login attempts with:
- Disconnecting: Protocol error: expected packet type 32, got 5 [preauth]

Before the Windows update, there weren't any issues. I confirmed the cause by preparing a fresh Win10 installation (then working) and then applying the update.

**To Reproduce**
Apply may'19 update to existing Windows10 Installation. Use libssh2 with WinCNG. Try login to remote ssh server.

**Expected behavior**
Login to remote ssh server possible. No protocol violation.

**Version (please complete the following information):**
Windows 10, May 2019 update (10.0.18362.207)
libssh2 1.9.0, compiled using VS2019

wez:
I've been tracing through this same problem today; in my case I'm trying to connect from windows to an ubuntu host.  I've traced it to this call failing with NTSTATUS STATUS_UNSUCCESSFUL:

https://github.com/libssh2/libssh2/blob/master/src/wincng.c#L1982

```
    ret = BCryptImportKeyPair(_libssh2_wincng.hAlgRSA, NULL,
                              BCRYPT_RSAPUBLIC_BLOB, &hKey, key, keylen,
                              BCRYPT_NO_KEY_VALIDATION);
```

sbredahl:
It sounds like the same wincng bug has been found in the TotalCommander sftp plugin, specific to the Windows 10 1903 version: According to its author _"The problem is a bug in the Windows function BCryptImportKeyPair: It fails to import an RSA key where the public key exponent is as large as the modulus. It works with a small 3 byte modulus 65537."_  (Have not verified it that is also the case with libssh2 however.)  https://www.ghisler.ch/board/viewtopic.php?t=52563.

wez:
I figured out an alternative solution (#397) that should work on Windows 8.1 and up, although I can only test it on my Windows 10 1903 system.

GupTk:
Hi thanks for your fork.
As my libssh2 1.7.0 is crashing on Windows 1903 and the 1.9.0 is failing ot cannot, I just tried your fork.The other end is an embedded linux (OpenSSH 5.8 - protocol 2.0).

Your fork cannot connect and I get a SSH_MSG_DISCONNECT error with the msg

Protocol error: expected packet type 32, got 5

Any idea?
[...]

PR #397, opened 2019-07-31
Fix DH kex on Windows 10 1903

wez:
Since Windows 1903 the approach used to perform DH kex with the CNG
API has been failing.

This commit switches to using the `DH` algorithm provider to perform
generation of the key pair and derivation of the shared secret.

It uses a feature of CNG that is not yet documented.  The sources of
information that I've found on this are:

* https://stackoverflow.com/a/56378698/149111
* https://github.com/wbenny/mini-tor/blob/5d39011e632be8e2b6b1819ee7295e8bd9b7a769/mini/crypto/cng/dh.inl#L355

With this change I am able to successfully connect from Windows 10 to my
ubuntu system.

Fixes: https://github.com/libssh2/libssh2/issues/388
Refs: https://github.com/alexcrichton/ssh2-rs/issues/122

### DIFF
--- a/src/wincng.c
+++ b/src/wincng.c
@@ -291,6 +291,11 @@ _libssh2_wincng_init(void)
                                                0);
         }
     }
+
+#if LIBSSH2_USE_BCRYPT_DH
+    (void)BCryptOpenAlgorithmProvider(&_libssh2_wincng.hAlgDH,
+                                      BCRYPT_DH_ALGORITHM, NULL, 0);
+#endif
 }
 
 void
@@ -310,6 +315,9 @@ _libssh2_wincng_free(void)
     (void)BCryptCloseAlgorithmProvider(_libssh2_wincng.hAlgAES_CBC, 0);
     (void)BCryptCloseAlgorithmProvider(_libssh2_wincng.hAlgRC4_NA, 0);
     (void)BCryptCloseAlgorithmProvider(_libssh2_wincng.hAlg3DES_CBC, 0);
+#if LIBSSH2_USE_BCRYPT_DH
+    (void)BCryptCloseAlgorithmProvider(_libssh2_wincng.hAlgDH, 0);
+#endif
 
     memset(&_libssh2_wincng, 0, sizeof(_libssh2_wincng));
 }
@@ -2126,6 +2134,25 @@ _libssh2_wincng_bignum_free(_libssh2_bn *bn)
     }
 }
 
+#if LIBSSH2_USE_BCRYPT_DH
+/* We provide our own prototype for this function as the availability
+ * of the header that is documented to provide it is patchy across
+ * the various environments that the libssh2 CI builds on, and
+ * because the stdcall convention is important for the linker to
+ * be able to resolve the function in 32-bit MSVC compiler
+ * environments. */
+extern NTSTATUS __stdcall RtlGetVersion(OSVERSIONINFOW*);
+
+static int is_windows_10_or_later(void)
+{
+    OSVERSIONINFOW vers;
+    vers.dwOSVersionInfoSize = sizeof(vers);
+    if(RtlGetVersion(&vers) != 0) {
+        return 0;
+    }
+    return vers.dwMajorVersion >= 10;
+}
+#endif
 
 /*
  * Windows CNG backend: Diffie-Hellman support.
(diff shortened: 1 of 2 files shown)


JanFellner:
Looks promising, working on my side.
Needed to upgrade to a newer platform sdk as it failed to build with an older one (10.0.10240.0)
I did only take over wincng.c and wincng.h into my project
Suggestion (without pull request)

use:
static int is_windows_10_or_later(void)
{
	return IsWindows10OrGreater();
}

(Thats the microsoft suggested way of checking versions)
https://docs.microsoft.com/en-us/windows/win32/api/versionhelpers/nf-versionhelpers-iswindowsversionorgreater
https://docs.microsoft.com/en-us/windows/win32/api/versionhelpers/nf-versionhelpers-iswindows10orgreater

wez:
@JanFellner see https://github.com/wez/libssh2/commit/3dcd743bc08db9e71ce4ec1ea4bb742201c51aca and https://github.com/libssh2/libssh2/pull/397/commits/7fe4a7acfe2a5a865cf5ef8be9659880c508d32b
The TL;DR is that `IsWindows10OrGreater` surprisingly returns `false` on Windows 10 if the embedding application hasn't compiled in a manifest that explicitly indicates that it runs on Windows 10.  That would make the version check less effective when this library is linked into a simple application.

willco007:
This build failed because it it can't find `versionhelpers.h`. Could we get an updated PR @wez? Thanks!

wez:
I'm not sure that that header is even used in the latest version of this PR; can you try taking it out and building that?  I don't have the right machine handy right now; it's been months since I looked at this

src/wincng.c
```diff
@@ -2134,35 +2161,288 @@ _libssh2_wincng_bignum_free(_libssh2_bn *bn)
...
+    }
+#endif
     /* Compute the shared secret */
-    _libssh2_wincng_bignum_mod_exp(secret, f, *dhctx, p);
+    _libssh2_wincng_bignum_mod_exp(secret, f, dhctx->bn, p);
     return 0;
 }
 
-void
-_libssh2_dh_dtor(_libssh2_dh_ctx *dhctx)
```
mback2k: Why is this function being removed? I think it is still needed for the original approach.

src/wincng.h
```diff
@@ -385,7 +394,17 @@ _libssh2_bn *_libssh2_wincng_bignum_init(void);
...
+typedef struct {
+#if LIBSSH2_USE_BCRYPT_DH
+    /* holds our private+public key components */
+    BCRYPT_KEY_HANDLE dh_handle;
+    /* records the parsed out modulus and generator parameters that are shared
+    * with the peer */
+    BCRYPT_DH_PARAMETER_HEADER *dh_params;
+#endif
+    /* fallback if the newer DH api doesn't work on this system */
+    struct _libssh2_wincng_bignum *bn;
```
mback2k: Why aren't you falling back to the original define? I think there is no need to nest the bignum struct in another struct.

src/wincng.c
wez: it hasn't been removed, it just appears to have been moved due to the way that the diff renders.  Look for `_libssh2_dh_dtor` in the RHS of this diff.

src/wincng.h
wez: the determination about whether we can use the DH API is based on a runtime test so both approaches need to be compiled into the same binary.

jonathanturcotte:
This fixed the issue we were seeing in #456, but I had to add ntdll to CMakeLists to get it to build for me properly on Windows Server 2012 and Visual Studio 2017. I'm not familiar with cmake, but I just added:

image

jonathanturcotte:
Is there anything else needed to have this PR put in? I can add those lines to the CMakeLists if needed.

willco007:
@mback2k are you OK to sign off on these changes?

mback2k:
> @mback2k are you OK to sign off on these changes?

I haven't tested this myself yet. I am still uncertain about including an unpublished/undocumented API call in libssh2 before it is published/documented. I still don't understand why the existing approach fails with recent Windows versions.

My plan is to come back to libssh2 development once [my current work on curl is done](https://github.com/curl/curl/pulls/mback2k), but if you need to go ahead with this now, I guess you are free to do so. One last thing though: is checking for Windows 10 sufficient here? I think the new approach is only needed on Windows 1903 or later?

src/wincng.c
```diff
@@ -2126,6 +2134,25 @@ _libssh2_wincng_bignum_free(_libssh2_bn *bn)
...
+extern NTSTATUS __stdcall RtlGetVersion(OSVERSIONINFOW*);
+
+static int is_windows_10_or_later(void)
+{
+    OSVERSIONINFOW vers;
+    vers.dwOSVersionInfoSize = sizeof(vers);
+    if(RtlGetVersion(&vers) != 0) {
+        return 0;
+    }
+    return vers.dwMajorVersion >= 10;
```
mback2k: Is this specific enough? I thought Windows 1903 is the relevant version?

brandl-muc:
Thanks @jonathanturcotte , your diff saved me. Seems to me that it should be integrated in this PR.

jonathanturcotte:
Is there any way this can go in soon?

mback2k:
> Is there any way this can go in soon?

Maybe you could comment on my questions above?

mback2k:
I will take a look into this the next 2 days. Sorry for the delay.

mback2k:
Quick update: I have got a slightly modified version working in my test environment without the version check due to the manifest requirement. I plan to have this merge ready in the upcoming week. Thanks for your patience.

mback2k:
`Ref: `https://www.ghisler.ch/board/viewtopic.php?t=52563

mback2k:
Quick update: my reworked approach regarding automatic fallback from `BCRYPT_KDF_RAW_SECRET` to the classic approach is now working on Windows 10 (1903) and Windows 7. I hope to have something merge ready very soon. I will keep you posted.

src/wincng.c
mback2k: Will be solved via #484.

mback2k:
@wez @JanFellner @willco007 @jonathanturcotte @brandl-muc 

I am so sorry that it took me so long to look into this. But I now finally have integrated the work of @wez and adapted it to a feature-testing-based fallback-approach with PR #484 that does not rely on version checking (which still did not work for me with this original PR due to the requirement of manifests). Therefore I would like to ask you to take a look at the new PR, test it on as much Windows versions you have available and give me some quick Go or No-Go feedback. Thanks in advance!

I hope that we can merge this very soon and also do a release shortly after. cc @bagder

jonathanturcotte:
Awesome, things are a bit hectic right now, but I'll try and give this a try in the next little while if I can!

mback2k:
Sorry for the delay, I was quite busy the last week. I just merged this into master and plan to do some follow ups via separate PRs.

@wez thanks a lot for the contribution!

PR #416, opened 2019-10-03
ssh-ed25519: Support PKIX + calc pubkey from private

Zenju:
Hi,

for my software FreeFileSync I'm supporting the PuTTY ppk file format, so I'm parsing the pkk files into OpenSSL's EVP_PKEY. Then generate PKIX PEM streams (PEM_write_bio_PrivateKey) in order to inter-op with libssh2 (libssh2_userauth_publickey_frommemory). This works nicely for
ssh-rsa
ssh-dss
ecdsa-sha2-nistp256
ecdsa-sha2-nistp384
ecdsa-sha2-nistp521
key algorithms because libssh2 is using PEM_read_bio_PrivateKey (great!) as an alternative to the openssh key format.

However I've found things do not work well for "ssh-ed25519"

1. _libssh2_pub_priv_keyfile() is missing the code to extract the ed25519 public key from a given private key

2. _libssh2_ed25519_new_private_frommemory is only parsing the openssh key format but does not understand PKIX (as retrieved via PEM_read_bio_PrivateKey)

This pull request resolves both issues: Tested and working nicely with ssh-ed25519-protected OpenSSH servers.

Regards, Zenju

PS: I've also done a bit of clean-up and simplified the libssh2_ed25519_ctx/libssh2_x25519_ctx structures. For ssh-ed25519 the public key can be generated from the private key, so why where two EVP_PKEY keys used previously? Am I missing something? I couldn't see a reason to store both while reviewing the code.

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -1109,7 +1109,6 @@ _libssh2_rsa_new_private(libssh2_rsa_ctx ** rsa,
 
     pem_read_bio_func read_rsa =
         (pem_read_bio_func) &PEM_read_bio_RSAPrivateKey;
-    (void) session;
 
     _libssh2_init_if_needed();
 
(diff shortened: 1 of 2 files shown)


willco007:
This looks good! There are a couple linter issues to be fixed up to pass the tests. As for keeping the public/private keys around, that was a hold over from the initial BoringSSL implementation that could be removed without issue.

PR #417, opened 2019-10-03
Fix Unicode builds on Windows

Zenju:
Hi,

libssh2 currently fails to compile in Visual Studio 16.3.2 for Unicode builds. The attached patch fixes this issue. Other than that libssh2 + pageant works great on Windows.

Best, Zenju


willco007:
Any Windows users have a comment on this PR?

PR #418, opened 2019-10-04
Fix name clash with ZLIB macro "compress"

Zenju:
Hi,

when libssh2 is compiled with ZLIB support, and ZLIB happens to be compiled with the Z_PREFIX_ macro set, the compress function name (which is a macro) is replaced with "prefix_compress", and comp.c fails to compile:

comp.c(390,23): error C2039: 'wx_zlib_compress': is not a member of 'flags'

In my case the Z_PREFIX_ = "wx_zlib_" is coming from the GUI library (wxWidgets) which is packaging zlib. The attached patch fixes the problem in a minimal way. But if it is still too ugly, maybe the _LIBSSH2_SESSION :: flags :: compress member should be renamed instead?

Best, Zenju

### DIFF
--- a/src/comp.c
+++ b/src/comp.c
@@ -369,6 +369,7 @@ static const LIBSSH2_COMP_METHOD *no_comp_methods[] = {
 const LIBSSH2_COMP_METHOD **
 _libssh2_comp_methods(LIBSSH2_SESSION *session)
 {
+    #undef compress /* dodge name clash with ZLIB macro */
     if(session->flag.compress)
         return comp_methods;
     else
--- a/src/comp.c
+++ b/src/comp.c
@@ -38,7 +38,8 @@
 
 #include "libssh2_priv.h"
 #ifdef LIBSSH2_HAVE_ZLIB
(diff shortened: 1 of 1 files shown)


willco007:
Seems to me that zlib should be undefining it, not libssh2 library, to fix this issue. Have you raised this issue with them?

Zenju:
"compress" is part of zlib's official API (https://zlib.net/manual.html#Utility).

willco007:
Anyone have other thoughts on this?

willco007:
Could we check to see if compress is defined and undefine it around this call and redefine it after if needed?

Zenju:
> 
> 
> Could we check to see if compress is defined and undefine it around this call and redefine it after if needed?

compress as part of zlib's API is always defined via zlib.h is, but latter only conditionally if LIBSSH2_HAVE_ZLIB. libssh2 isn't using the compress macro, so it's not needed to redefine it and  also would introduce an even tighter coupling with zlib implementation details for no benefit. I've updated the fix accordingly.

PR #421, opened 2019-10-10
lock mechanism to protect the file hnown_host + new function to add host+ key

mickae1:
I needed for my project a way to protect the file known_host + a function that add to the file a unique host  to the file.

### DIFF
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -269,10 +269,8 @@ typedef off_t libssh2_struct_stat_size;
 #define LIBSSH2_FREE_FUNC(name)    void name(void *ptr, void **abstract)
 
 /* lock callbacks */
-#define LIBSSH2_LOCK_INIT_FUNC(name)   void* name()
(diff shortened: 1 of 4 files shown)


willco007:
Thanks for your contribution! Could these APIs be locked before access in the client application to avoid pushing the lock down into the library?

mickae1:
I analyzed the code of libssh2 and it looks like that its designed to be
independent from system functions.

i would like to use pthread lock in libssh2, but if I'm not sure it's the
best to do.

Le mar. 15 oct. 2019 à 20:23, Will Cosgrove <notifications@github.com> a
écrit :

mickae1:
Yes you are right. I will do the modifications.
But I need at least the function that remove the previous host before
adding the new host + fingerprint. It's a functionality that it is missing
in libssh2. Replacing the old host with the new key fingerprint.

Le mer. 16 oct. 2019 à 19:02, Will Cosgrove <notifications@github.com> a
écrit :

willco007:
This function will delete an entry `libssh2_knownhost_del`.

PR #437, opened 2019-12-11
scp: fix state when empty channel is returned

colinw-smartrg:
This allows the calling app to get the stderr data from the channel.
Then when it is closed and free'd, the next call to scp_recv will be in the correct state.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -985,7 +985,8 @@ session_free(LIBSSH2_SESSION *session)
     /* Free startup_key_state.key_state_low variables */
 #if LIBSSH2_ECDSA
     if(session->startup_key_state.key_state_low.private_key != NULL)
-        _libssh2_ecdsa_free(session->startup_key_state.key_state_low.private_key);
+        _libssh2_ecdsa_free(
+		session->startup_key_state.key_state_low.private_key);
 #endif
     if(session->startup_key_state.key_state_low.public_key_oct != NULL)
        free(session->startup_key_state.key_state_low.public_key_oct);
--- a/src/session.c
+++ b/src/session.c
@@ -986,7 +986,7 @@ session_free(LIBSSH2_SESSION *session)
 #if LIBSSH2_ECDSA
     if(session->startup_key_state.key_state_low.private_key)
         _libssh2_ecdsa_free(
-               session->startup_key_state.key_state_low.private_key);
+                         session->startup_key_state.key_state_low.private_key);
 #endif
     if(session->startup_key_state.key_state_low.public_key_oct)
         free(session->startup_key_state.key_state_low.public_key_oct);
--- a/src/session.c
+++ b/src/session.c
@@ -984,11 +984,11 @@ session_free(LIBSSH2_SESSION *session)
 
     /* Free startup_key_state.key_state_low variables */
 #if LIBSSH2_ECDSA
-    if(session->startup_key_state.key_state_low.private_key != NULL)
+    if(session->startup_key_state.key_state_low.private_key)
         _libssh2_ecdsa_free(
                session->startup_key_state.key_state_low.private_key);
 #endif
-    if(session->startup_key_state.key_state_low.public_key_oct != NULL)
+    if(session->startup_key_state.key_state_low.public_key_oct)
         free(session->startup_key_state.key_state_low.public_key_oct);
 
     /*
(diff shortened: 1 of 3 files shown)


colinw-smartrg:
Added two minor memory leak and double free errors.

colinw-smartrg:
So it fails if the C ident of the else { is not on the same line. That is a bit persnickety.
Personally I like the open { to be on the same indent level as the close. Just like the beginning of the function. Much easier to read.

colinw-smartrg:
Now a line is too long ? 
Is there a pre-processor I can run to see if I have any other persnickety style issues before I add another bunch of commits to fix code spacing ?
How about a CI engine that finds all the issues in one go ?

src/session.c
```diff
@@ -982,6 +982,15 @@ session_free(LIBSSH2_SESSION *session)
         LIBSSH2_FREE(session, session->remote.lang_prefs);
     }
 
+    /* Free startup_key_state.key_state_low variables */
```
willco007, line 985: It seems like this should be cleaned up in kex.c so consumers (the session) doesn't need to know about implementation details. Without being able to reproduce this, it's hard to say what the best approach would be. I'd also suggest NULL'ing the freed memory for safety.

willco007:
I left a review comment on this in May. Any chance of a follow up @colinw-smartrg, thanks!

src/session.c
colinw-smartrg: You are probably right.
I haven't looked at this in some time, but as I recall, the problem was a memory leak that was found using valgrind.
We have an application that connects and disconnects all the time and noticed a slow memory leak that was tracked down to those objects not being freed.
I am not familiar with the kex architecture, but something should trigger freeing those when the session is destroyed. 
I can't even find any reference to the allocation of the startup_key_state object.

colinw-smartrg:
I was hoping these would be pulled into the latest release.

We will not be moving our use of libssh2 to the latest without them.

willco007:
I was hoping this PR would be tidied up a bit before accepting. Namely, moving it into the correct location as per my comments back in May.

colinw-smartrg:
I'm still not familiar enough with the kex.c code / architecture to know where to put the free() that need to be added to clean up properly. 
Maybe you can get some help from someone that is familiar with where startup_key_state.key_state_low.private_key objects are allocated and supposed to be free'd. If they can point me to where the destructor code for it should be, I'd be happy to move it there.

src/session.c
```diff
@@ -982,6 +982,15 @@ session_free(LIBSSH2_SESSION *session)
...
+#if LIBSSH2_ECDSA
+    if(session->startup_key_state.key_state_low.private_key != NULL)
+        _libssh2_ecdsa_free(
+               session->startup_key_state.key_state_low.private_key);
+#endif
+    if(session->startup_key_state.key_state_low.public_key_oct != NULL)
+       free(session->startup_key_state.key_state_low.public_key_oct);
```
vszakats, line 992:
```suggestion
        free(session->startup_key_state.key_state_low.public_key_oct);
```

src/session.c
```diff
@@ -982,6 +982,15 @@ session_free(LIBSSH2_SESSION *session)
         LIBSSH2_FREE(session, session->remote.lang_prefs);
     }
 
+    /* Free startup_key_state.key_state_low variables */
+#if LIBSSH2_ECDSA
+    if(session->startup_key_state.key_state_low.private_key != NULL)
+        _libssh2_ecdsa_free(
+               session->startup_key_state.key_state_low.private_key);
+#endif
+    if(session->startup_key_state.key_state_low.public_key_oct != NULL)
```
vszakats, line 991:
```suggestion
    if(session->startup_key_state.key_state_low.private_key)
        _libssh2_ecdsa_free(
               session->startup_key_state.key_state_low.private_key);
#endif
    if(session->startup_key_state.key_state_low.public_key_oct)
```

src/session.c
```diff
@@ -982,6 +982,15 @@ session_free(LIBSSH2_SESSION *session)
         LIBSSH2_FREE(session, session->remote.lang_prefs);
     }
 
+    /* Free startup_key_state.key_state_low variables */
+#if LIBSSH2_ECDSA
+    if(session->startup_key_state.key_state_low.private_key)
+        _libssh2_ecdsa_free(
+               session->startup_key_state.key_state_low.private_key);
```
vszakats, line 989:
```suggestion
                         session->startup_key_state.key_state_low.private_key);
```

vszakats:
The issue seems valid, but the fix would need more work according to the thread. Closing for now.

Feel free to reopen if there is any new development or as an Issue.

PR #439, opened 2019-12-16
Fix use-after-free crash in openssl backend without memory leak

meierha:
In pull request 387 a use-after-free crash in openssl backend was fixed, see https://github.com/libssh2/libssh2/pull/387

I can confirm that this crash also happens in my scenario. 

Pull request 387 was merged into master on 12.Jul 2019. The fix from the pull request fixes the use-after-free bug but also introduces a memory leak as it already has been stated in pull request 413, see https://github.com/libssh2/libssh2/pull/413

I also can confirm this memory leak. Pull request 413 was closed without being merged. When I had a look into it, it basically was a step back to the previous situation, so not a suitable solution.

So, after carefully inspecting the source, I think I understood the original intention of the author of the "#ifdef HAVE_OPAQUE_STRUCTS" blocks and fixed the bug according to this intention. Basically the bug was to store the pointer returned by make_ctr_evp in a function scope static pointer instead of the already existing module global static pointer aes_128_ctr_cipher (192 and 256 all the same). This function scope static pointer wasn't (and certainly could not) be zeroed when the struct was feed in some other line of code in some other function.

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -596,7 +596,8 @@ _libssh2_EVP_aes_128_ctr(void)
 {
 #ifdef HAVE_OPAQUE_STRUCTS
     return !aes_128_ctr_cipher ?
-        make_ctr_evp(16, &aes_128_ctr_cipher, NID_aes_128_ctr) : aes_128_ctr_cipher;
+        make_ctr_evp(16, &aes_128_ctr_cipher, NID_aes_128_ctr) : 
+        aes_128_ctr_cipher;
 #else
     static EVP_CIPHER aes_ctr_cipher;
     static EVP_CIPHER * aes_ctr_cipher_ptr = &aes_ctr_cipher;
@@ -610,7 +611,8 @@ _libssh2_EVP_aes_192_ctr(void)
 {
 #ifdef HAVE_OPAQUE_STRUCTS
     return !aes_192_ctr_cipher ?
-        make_ctr_evp(24, &aes_192_ctr_cipher, NID_aes_192_ctr) : aes_192_ctr_cipher;
+        make_ctr_evp(24, &aes_192_ctr_cipher, NID_aes_192_ctr) : 
+        aes_192_ctr_cipher;
 #else
     static EVP_CIPHER aes_ctr_cipher;
     static EVP_CIPHER * aes_ctr_cipher_ptr = &aes_ctr_cipher;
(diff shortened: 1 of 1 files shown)


willco007:
Looks good, anyone else want to take a look at this solution?

src/openssl.c
```diff
@@ -589,49 +595,46 @@ const EVP_CIPHER *
...
-    EVP_CIPHER * aes_ctr_cipher;
-    return make_ctr_evp(16, &aes_ctr_cipher, NID_aes_128_ctr);
+    return !aes_128_ctr_cipher ?
+        make_ctr_evp(16, &aes_128_ctr_cipher, NID_aes_128_ctr) : aes_128_ctr_cipher;
 #else
     static EVP_CIPHER aes_ctr_cipher;
-    EVP_CIPHER * aes_ctr_cipher_ptr = &aes_ctr_cipher;
-    return make_ctr_evp(16, &aes_ctr_cipher_ptr, 0);
+    static EVP_CIPHER * aes_ctr_cipher_ptr = &aes_ctr_cipher;
+    return !aes_ctr_cipher.key_len ?
```
tsengjun: using uninitialized variable causes compiler warning line 601 "aes_ctr_cipher" not initialized, how about give a initial value to aes_ctr_cipher.len? or replace the check condition with " !aes_128_ctr_cipher". the aes 192/256 ctor has similar behavior.

Thanks!

meierha:
Hi,

... back in office today at the customer where I had this problem, so sorry for the delay!

So, I am afraid you are right. Here we depend on whether there is a constructor for  EVP_CIPHER (which comes from openssl, not from libssh) that initializes key_len to zero. Looking into the source of (some release of) openssl, it does not, it is just a plain struct without constructor (well, sure, it is C, not C++).

To your suggestions: Simply initializing key_len doesn't work because it may only be initialized while constructing the static object, so only once. So I decided for using aes_128_ctr_cipher being NULL or not to determine the initialization state.

By the way, your comment doesn't appear on https://github.com/libssh2/libssh2/pull/439 ... would have been better! How did you comment so that it doesn't appear there?

Best wishes

Hans

Am 04.02.20 um 03:05 schrieb Tseng Jun:

@tsengjun commented on this pull request.

________________________________

In src/openssl.c<https://github.com/libssh2/libssh2/pull/439#discussion_r374440824>:

>  #else
     static EVP_CIPHER aes_ctr_cipher;
-    EVP_CIPHER * aes_ctr_cipher_ptr = &aes_ctr_cipher;
-    return make_ctr_evp(16, &aes_ctr_cipher_ptr, 0);
+    static EVP_CIPHER * aes_ctr_cipher_ptr = &aes_ctr_cipher;
+    return !aes_ctr_cipher.key_len ?


using uninitialized variable causes compiler warning line 601 "aes_ctr_cipher" not initialized, how about give a initial value to aes_ctr_cipher.len? or replace the check condition with " !aes_128_ctr_cipher". the aes 192/256 ctor has similar behavior.

Thanks!

-

tsengjun:
Hi Hans,
It is because I saw C99 already considered the static variable and initialized it, so no problem, that’s why my comment not apply. Anyway, your&nbsp;new fix looks good too!


Thanks!





------------------ Original ------------------

PR #458, opened 2020-03-10
channel: Refactor closed tracking logic to be more robust.

augustin:
Previously, there was a "close" flag in both the local and remote
state structures, but only the local one was checked in
_libssh2_channel_close, meaning that if someone invoked "channel_close"
after the channel had already been closed by the server,
an indefinite hang would occur.

Now, there is only one "closed" field, which is set when the
server replies that the channel is in fact closed.

### DIFF
--- a/src/channel.c
+++ b/src/channel.c
@@ -2109,7 +2109,7 @@ ssize_t _libssh2_channel_read(LIBSSH2_CHANNEL *channel, int stream_id,
            that back. We may have gotten that info while draining the incoming
            transport layer until EAGAIN so we must not be fooled by that
            return code. */
-        if(channel->remote.eof || channel->remote.close)
+        if(channel->remote.eof || channel->closed)
             return 0;
         else if(rc != LIBSSH2_ERROR_EAGAIN)
             return 0;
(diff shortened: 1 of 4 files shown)


augustin:
Something is still not working quite right here. When single-stepping in a debugger, I always get to the top of `libssh2_channel_close` with `closed=1`, or if not the first time, then the packet processing loop inside it finds the message with the close in it. When I don't single-step, then the `SSH_MSG_CHANNEL_CLOSE` case is never even hit.

(I'm running on Windows with WinCNG, for whatever it's worth.)

augustin:
Actually, it seems the code I was using had a few race conditions, and this may not be strictly necessary after all to prevent hangs. Still, it's more correct to not send a close request after the channel has already been closed anyway.

augustin:
If you are uninterested as a project in actually reviewing pull requests and letting them become "stale" and then closed arbitrarily, then I'm not really interested.

mback2k:
Sorry, but this project has been in quite dormant state for some time now. I don't have much time to spend on this, but my plan is to go through the stale PRs from time to time and eventually merge them.

Issue #457, opened 2020-03-06
Add support for RFC8258 key exchanges


sldr:
Add support the following key exchanges:
diffie-hellman-group14-sha256
diffie-hellman-group15-sha512
diffie-hellman-group16-sha512
diffie-hellman-group17-sha512
diffie-hellman-group18-sha512

TDi-jonesds:
since more implementations are defaulting to it, at least please add diffie-hellman-group14-sha256

willco007:
I'm finishing up work on group16 and group18 to bring us to the defaults OpenSSH uses which seems adequate.

asvc:
@willco007 , any update on releasing a version with this and maybe https://github.com/libssh2/libssh2/pull/493?

willco007:
This work is sitting on PR #468. I've been busing with other things, but hopefully I can loop back around and land it before too long. As for a release, someone (again, me most likely) still needs to address the missing key test files which are skipped by the release script.

willco007:
Support landed with #468, closing.

PR #464, opened 2020-04-28
Add diffie-hellman-group14-sha256 Key Exchange Method

piru:
diffie-hellman-group14-sha256 is the transitional key exchange method from sha1 to sha256. It is often found on embedded systems as it is fairly cheap on resources while still considered secure.


willco007:
Tracking issue #457.

src/kex.c
```diff
@@ -1637,6 +1637,83 @@ kex_method_diffie_hellman_group14_sha1_key_exchange(LIBSSH2_SESSION *session,
...
 
+/* kex_method_diffie_hellman_group14_sha256_key_exchange
+ * Diffie-Hellman Group14 Key Exchange using SHA256
+ */
+static int
+kex_method_diffie_hellman_group14_sha256_key_exchange(LIBSSH2_SESSION *session,
+                                                      key_exchange_state_low_t
+                                                      * key_state)
+{
+    static const unsigned char p_value[256] = {
```
willco007, line 1648: Could we pull this value out group14 sha1 function to reuse it instead of redefining it twice?

piru: > Could we pull this value out group14 sha1 function to reuse it instead of defining it twice?
Sure. Added a generic kex_method_diffie_hellman_group14_key_exchange function that is passed the DH hash function, diffie_hellman_sha1 or diffie_hellman_sha256.

willco007:
Looks good to me, thanks for the PR!

PR #465, opened 2020-04-30
os400qc3.h: define sha512 macros

willco007:
file: os400qc3.h
notes: fix for modern building of libssh2


monnerat:
I still can't check it, but you probably also have to define `libssh2_sha512_ctx` as `Qc3_Format_ALGD0100_T`.

willco007:
@monnerat Ah yes, missed that. Flying blind here. :)

monnerat:
The commit seems good, but the comment is wrong !

PR #467, opened 2020-05-05
kex.c: group16-sha512 and group18-sha512 support #457

willco007:
kex.c

Notes:
Added key exchange group16-sha512 and group18-sha512. As a result did the following:
- Abstracted diffie_hellman_sha256() to diffie_hellman_sha_algo() which is now algorithm agnostic and takes the algorithm as a parameter since we needed sha512 support. Unfortunately is required some helper functions but they are simple.
- Deleted diffie_hellman_sha1()
- Deleted diffie_hellman_sha1 specific macro
- Cleaned up some formatting

Credit:
Will Cosgrove


willco007:
Ah shoot, this isn't going to work due to Windows using a struct for it's sha context.

PR #468, opened 2020-05-05
kex.c: group16-sha512 and group18-sha512 support #457

willco007:
kex.c

Notes:
Added key exchange group16-sha512 and group18-sha512. As a result did the following:

- Abstracted diffie_hellman_sha256() to diffie_hellman_sha_algo() which is now algorithm agnostic and takes the algorithm as a parameter since we needed sha512 support. Unfortunately it required some helper functions but they are simple.
- Deleted diffie_hellman_sha1()
- Deleted diffie_hellman_sha1 specific macro
- Cleaned up some formatting
- Defined sha384 in os400 and wincng backends
- Defined LIBSSH2_DH_MAX_MODULUS_BITS to abort the connection if we receive too large of p from the server doing sha1 key exchange.
- Reorder the default key exchange list to match OpenSSH and improve security

Credit:
Will Cosgrove


include/libssh2.h
```diff
@@ -239,7 +239,7 @@ typedef off_t libssh2_struct_stat_size;
 #define LIBSSH2_DH_GEX_OPTGROUP     1536
 #define LIBSSH2_DH_GEX_MAXGROUP     2048
 
-#define LIBSSH2_DH_MAX_MODULUS_BITS 10000
+#define LIBSSH2_DH_MAX_MODULUS_BITS 16384
```
kdudka, line 242: It is pretty obvious from the diff what the change does.  It would be good if the commit message explained why the change was needed or useful.

willco007: Sure, the value 10000 was pulled out of OpenSSH, which is the value I've been using for a while in my fork without issue. With this commit making into the official release, I double-checked OpenSSH and their max values change if using RSA or DSA keys so I bumped this value to the larger number. This is to prevent DOS with extremely large values being computed.

willco007:
I'm at a bit of a loss as to why the mbedTLS 64 bit builds are falling, but the 32 bit builds are working.

asvc:
@kdudka , any chance this can be reviewed?

kdudka:
@asvc Sorry if this was blocked on me.  I do not feel competent to review this pull request.

@willco007 Please feel free to remove me as a reviewer.  I can't see how to do it myself.

willco007:
I recently shipped this update in my app and I don't expect any issues. I'll rebase and commit if there are no objections.

mback2k:
@willco007 please check this PR again, it now contains 53 commits instead of being rebased.

willco007:
@mback2k The rebase choked and this was the result. I attempted to clean it up but alas. I could attempt to restart my changes from scratch but if this is squashed it will flatten all the merge commit messages into the final commit so is it worth starting over? I've already fought with this for far too long.

mback2k:
Ok, maybe you can just force-push a squashed commit to this PR branch?

willco007:
@mback2k let me give it a shot.

PR #470, opened 2020-05-16
Use libssh2_EXPORTS as an alternative to _WINDLL

xhochy:
`_WINDLL` is only defined when a Visual Studio CMake generator is used, `libssh2_EXPORTS` is used though for all CMake generator if a shared libssh2 library is being built.

The corresponding CMake issue is at https://gitlab.kitware.com/cmake/cmake/issues/16370


xhochy:
Would it be possible to get review instead of closing?

xhochy:
I would be still interested in contributing this!

xhochy:
I would be still interested in contributing this!

PR #473, opened 2020-05-24
Fix detailed _libssh2_error being overwritten

Zenju:
Hi,

_libssh2_pub_priv_keyfilememory in src\openssl.c already sets _libssh2_error() with detailed error messages, while its caller, src\userauth.c, userauth_publickey_frommemory() overwrites it with a generic "Unable to extract public key from private key."
The patch fixes this to facilitate troubleshooting userauth_publickey_frommemory issues.

Best


Zenju:
PS: I've checked all code paths to make sure that all subroutines do indeed set _libssh2_error with detailed info.

PR #475, opened 2020-05-25
Changed type of LIBSSH2_FX_* constants to unsigned long

### LINKED ISSUE #474: LIBSSH2_FX_* constants have mismatching types
Error constants `LIBSSH2_FX_*` defined in [`include/libssh2_sftp.h`:192](https://github.com/libssh2/libssh2/blob/4e75a59e294b7170428f028267460b1026c2634e/include/libssh2_sftp.h#L192) are only returned by the API `libssh2_sftp_last_error()` which is defined as:
```
unsigned long libssh2_sftp_last_error(LIBSSH2_SFTP *sftp);
```
Therefore, all error constants `LIBSSH2_FX_*` should have the type `unsigned long`. Their current definition gives them the type `int`. For example:
```
#define LIBSSH2_FX_NO_SUCH_FILE             2
```
Whereas, they should be defined as `unsigned long` literals instead. For example:
```
#define LIBSSH2_FX_NO_SUCH_FILE             2UL
```

**Additional context:**
This might not be an issue for the `C/C++` languages which perform implicit automatic conversions when comparison is performed between the return value of `libssh2_sftp_last_error()` and the  `LIBSSH2_FX_*` constants. However, bindings to `libssh2` in more strictly-typed languages, such as Rust, where bindings are automatically generated from the public headers of `libssh2`, and where the mentioned implicit type conversion does not happen, consider this type mismatch as an error.

koutheir:
Error constants `LIBSSH2_FX_*` are only returned by `libssh2_sftp_last_error()` which returns `unsigned long`. Therefore these constants should be defined as `unsigned long` literals, instead of `int` literals.

Fixes #474

### DIFF
--- a/include/libssh2_sftp.h
+++ b/include/libssh2_sftp.h
@@ -189,30 +189,30 @@ struct _LIBSSH2_SFTP_STATVFS {
 #define LIBSSH2_FXF_EXCL                        0x00000020
 
 /* SFTP Status Codes (returned by libssh2_sftp_last_error() ) */
-#define LIBSSH2_FX_OK                       0
-#define LIBSSH2_FX_EOF                      1
-#define LIBSSH2_FX_NO_SUCH_FILE             2
-#define LIBSSH2_FX_PERMISSION_DENIED        3
-#define LIBSSH2_FX_FAILURE                  4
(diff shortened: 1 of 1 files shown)


willco007:
Thanks for the PR!

PR #476, opened 2020-05-28
fix use-of-uninitialized-value

ltx2018:
malloc(0) may not always return NULL, return error if 'total_num = 0' in _libssh2_transport_read in case of 'payload' will be used later


willco007:
Looks good, thanks!

samueloph:
CVE-2020-22218 was assigned to this, and NVD scored it as "high" severity.

I had not involvement in the assignment. I'm commenting here for reference.

PR #481, opened 2020-06-01
Update INSTALL_CMAKE.md

suryakalpo:
Although the original text would be immediately clear to seasoned users of CMAKE and/or Unix shell, the lack of newlines may cause some confusion for newcomers. Hence, wrapping the texts in a code-block such that the newlines appear as intended.


willco007:
Thanks for the PR.

PR #484, opened 2020-06-08
Change PR#397 to a feature-testing-based fallback-approach

mback2k:
_Description of the most relevant commit in this PR:_

Avoid the use of RtlGetVersion or similar Win32 functions,
since these depend on version information from manifests.

This commit makes the WinCNG backend first try to use the
new DH algorithm API with the raw secret derivation feature.
In case this feature is not available the WinCNG backend
will fallback to the classic approach of using RSA-encrypt
to perform the required modular exponentiation of BigNums.

The feature availability test is done during the first handshake
and the result is stored in the crypto backends global state.

Follow up to #397
Closes #484

### DIFF
--- a/src/wincng.c
+++ b/src/wincng.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2013-2015 Marc Hoersken <info@marc-hoersken.de>
+ * Copyright (C) 2013-2020 Marc Hoersken <info@marc-hoersken.de>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms,
@@ -58,6 +58,7 @@
 
 #include <windows.h>
 #include <bcrypt.h>
+#include <ntstatus.h>
 #include <math.h>
 #include "misc.h"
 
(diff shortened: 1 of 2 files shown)


mback2k:
**Attention:** this PR is not indented to be merged, this is just a CI vehicle for my current WIP changes.

mback2k:
This PR now also temporarily caries things from #485 to have the fixes in for testing.

mback2k:
> But, most likely the pre-existing BCrypt calls already bumped this requirement, in which case there is no regression here.

Yes, this has been the case from the beginning of the WinCNG backend, because that is only available starting with Vista. It makes no sense to compile libssh2 with that and then try to run it on an earlier version. Unless libssh2 at some point gets MultiCrypto backend support (like curl with MultiSSL), I don't see any use in complicating the code to use `GetProcAddress()`.

mback2k:
Although CI already passed, I just rebased this to current master to strop duplicate commits.

mback2k:
@wez @jonathanturcotte would you mind taking a look at this proposal? :rocket:

jonathanturcotte:
Can confirm this branch worked for me on my Windows 10, v1909 machine. Thanks for this @mback2k !

wez:
Confirmed that this works for me on 1909 in a rust application too

mback2k:
Thanks, I plan merging this some time this week.

mback2k:
Sorry for the delay, I was quite busy the last week. I just merged this into master and plan to do some follow ups via separate PRs.

PR #485, opened 2020-06-14
Various fixes to the BigNum and DH handling in WinCNG

mback2k:
- wincng: add and improve checks in bit counting function
- wincng: align bits to bytes calculation in all functions
- wincng: do not disable key validation that can be enabled
- wincng: fix return value in _libssh2_dh_secret


mback2k:
Thanks!

PR #499, opened 2020-07-08
openssl: clean up curve25519 code

davidben:
This cleans up a few things in the curve25519 implementation:

- There is no need to create X509_PUBKEYs or PKCS8_PRIV_KEY_INFOs to
  extract key material. EVP_PKEY_get_raw_private_key and
  EVP_PKEY_get_raw_public_key work fine.

- libssh2_x25519_ctx was never used (and occasionally mis-typedefed to
  libssh2_ed25519_ctx). Remove it. The _libssh2_curve25519_new and
  _libssh2_curve25519_gen_k interfaces use the bytes. Note, if it needs
  to be added back, there is no need to roundtrip through
  EVP_PKEY_new_raw_private_key. EVP_PKEY_keygen already generated an
  EVP_PKEY.

- Add some missing error checks.

NB: I was able to check this compiled, but it looks like the tests requires some docker setup. So probably wait for the CI to clear here.

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -1465,75 +1465,53 @@ _libssh2_ecdsa_new_private_frommemory(libssh2_ecdsa_ctx ** ec_ctx,
 #if LIBSSH2_ED25519
 
 int
-_libssh2_curve25519_new(LIBSSH2_SESSION *session, libssh2_x25519_ctx **out_ctx,
+_libssh2_curve25519_new(LIBSSH2_SESSION *session,
                         unsigned char **out_public_key,
                         unsigned char **out_private_key)
 {
     EVP_PKEY *key = NULL;
     EVP_PKEY_CTX *pctx = NULL;
-    PKCS8_PRIV_KEY_INFO *info = NULL;
-    ASN1_OCTET_STRING *oct = NULL;
-    X509_PUBKEY *pubkey = NULL;
-    libssh2_ed25519_ctx *ctx = NULL;
-    const unsigned char *pkcs, *priv, *pub;
-    int privLen, pubLen, pkcsLen;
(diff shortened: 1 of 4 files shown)


davidben:
> NB: I was able to check this compiled, but it looks like the tests requires some docker setup. So probably wait for the CI to clear here.

Ah great. Looks like CI is clean.

src/openssl.c
```diff
@@ -1465,92 +1465,68 @@ _libssh2_ecdsa_new_private_frommemory(libssh2_ecdsa_ctx ** ec_ctx,
...
 {
     EVP_PKEY *key = NULL;
     EVP_PKEY_CTX *pctx = NULL;
-    PKCS8_PRIV_KEY_INFO *info = NULL;
-    ASN1_OCTET_STRING *oct = NULL;
-    X509_PUBKEY *pubkey = NULL;
-    libssh2_ed25519_ctx *ctx = NULL;
-    const unsigned char *pkcs, *priv, *pub;
-    int privLen, pubLen, pkcsLen;
+    unsigned char *priv, *pub;
```
willco007, line 1474: priv, pub can be uninitialized on cleanExit if EVP_PKEY_keygen_* fails. They should be initialized to NULL here.

willco007:
I've added one small comment, otherwise this looks good to me.

src/openssl.c
davidben: Ah oops. Fixed.

dfandrich:
On Wed, Jul 08, 2020 at 01:10:30PM -0700, David Benjamin wrote:
>     NB: I was able to check this compiled, but it looks like the tests requires
>     some docker setup. So probably wait for the CI to clear here.

NB, running the tests locally is as easy as "make check".

PR #506, opened 2020-08-04
alternative functions added to fill LIBSSH2_SFTP_ATTRIBUTES on sftp_open

vajdaakos:
Currently, libssh2 sends hardcoded LIBSSH2_SFTP_ATTRIBUTES struct on handle opening. This can be problematic on some special OS, where the file size should be known on new file creation. I added some function variations to resolve this issue.


doorsdown:
This looks sane to me

PR #510, opened 2020-08-29
Recv and send all bytes via network

cerevra:
**Describe the bug**
libssh2 functions return error [LIBSSH2_ERROR_AGENT_PROTOCOL](https://github.com/libssh2/libssh2/blob/master/include/libssh2.h#L499) for valid client-server protocol.
Reason: posix functions `recv()`/`send()` accept buffer_start_pointer and buffer_len. Theese functions return succeed len which could be less then buffer_len. libssh2 doesn't cover this case ( [1](https://github.com/libssh2/libssh2/blob/master/src/agent.c#L187), [2](https://github.com/libssh2/libssh2/blob/master/src/libssh2_priv.h#L198) ): so only part of message could be finished.

It leads to operations with garbage at the end of buffer on `recv()`.
For example in [agent_list_identities()](https://github.com/libssh2/libssh2/blob/master/src/agent.c#L498) in my project:
```
=== agent_list_identities got len: 689
=== agent_list_identities got *s: 12
=== agent_list_identities got num_identities: 2
=== agent_list_identities === next identity
=== agent_list_identities got blob_len: 277
=== agent_list_identities got comment_len: 35
=== agent_list_identities === next identity
=== agent_list_identities got blob_len: 279
=== agent_list_identities got comment_len: 3486502863
=== agent_list_identities goto error #7: -42
```

**To Reproduce**
I can't provide real steps in client: it depends on network quality and server side.

**Expected behavior**
Functions should return success code if client-server protocol is correct.

 - OS: linux
 - libssh2 version 1.8.0

**Additional context**
Duplicate: https://github.com/libssh2/libssh2/pull/511

### DIFF
--- a/src/agent.c
+++ b/src/agent.c
@@ -175,6 +175,40 @@ agent_connect_unix(LIBSSH2_AGENT *agent)
     return LIBSSH2_ERROR_NONE;
 }
 
+#define RECV_SEND_ALL(func, socket, buffer, length, flags, abstract) \
+    int rc; \
+    size_t finished; \
+    \
+    finished = 0; \
+    \
+    while(finished < length) { \
+        rc = func(socket, \
+                  (char *)buffer + finished, length - finished, \
+                  flags, abstract); \
+        if(rc < 0) \
+            return rc; \
+        \
+        finished += rc; \
+    } \
+    \
+    return finished;
+
+static ssize_t _send_all(LIBSSH2_SEND_FUNC(func), libssh2_socket_t socket,
+                         const void *buffer, size_t length,
+                         int flags, void **abstract)
+{
+    RECV_SEND_ALL(func, socket, buffer, length, flags, abstract);
+}
+
+static ssize_t _recv_all(LIBSSH2_RECV_FUNC(func), libssh2_socket_t socket,
+                         void *buffer, size_t length,
+                         int flags, void **abstract)
+{
+    RECV_SEND_ALL(func, socket, buffer, length, flags, abstract);
+}
+
+#undef RECV_SEND_ALL
+
 static int
 agent_transact_unix(LIBSSH2_AGENT *agent, agent_transaction_ctx_t transctx)
 {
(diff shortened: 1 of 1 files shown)


cerevra:
Sorry for several issues: I'm not very experienced in github

cerevra:
Test in broken in master: https://github.com/libssh2/libssh2/issues/513

cerevra:
@bagder 
@mback2k
Can I ask you to pay attention to this issue? How can I move it forward?

mback2k:
I like the proposed solution, but I am wondering why this issue is coming up now. Isn't it the callers responsibility to call the API until all data is available? @bagder @willco007

willco007:
@mback2k I was just thinking the same thing. The solution seems good, but it seems odd this would be coming up now.

pkittenis:
Indeed clients are currently handling this, and this code is also [in examples](https://github.com/libssh2/libssh2/blob/master/example/scp.c#L152-L168).

willco007:
Looking at the code further, I think @pkittenis is right. All the places in sftp.c I looked at all take into account the possibility of a partial reply.

cerevra:
OK, let's fix only definitely broken function =). `agent_transact_unix()` encupsulates operation with network and doesn't check length  -  i.e. result of the first recv/send call. `agent_transact_unix()` is used only in internal functions `agent_list_identities()` and `agent_sign()`.
The only solution I see is to handle case with incomplete buffer inside of `agent_transact_unix()`.

As far as I see length is checked for other usages of `LIBSSH2_SEND()`.

Could I ask you to review next iteration of this patch?
@bagder @mback2k @willco007 @pkittenis

cerevra:
Are there any comments for patch? Could I provide any help here?

cerevra:
Could you please take a look at this issue?
@bagder @mback2k @willco007 @pkittenis

cerevra:
Could I provide any help to move forward this issue?

torkve:
Hi all!

I have client based on libssh2, that suffers from this very issue since my ssh-agent yields chunked data.
This patch appears to fix it. Could you merge it or mark if any problems exist, please?

PR #517, opened 2020-09-14
agent: Add support for Windows OpenSSH agent

yodaldevoid:
This is still WIP. I threw this together in a day loosely based on the corresponding code in OpenSSH. Currently it does not use async IO.

The main reason I am opening this PR is to  maybe get some direction on how to maybe support testing this via the testing framework. That, and I need a good way to test this myself.

Things left to clean up:
[ ] test it
[ ] Figure out license shenanigans
[ ] Support async IO


yodaldevoid:
I think this is on a technical level ready to merge now. Please let me know if you want me to clean up anything either code wise or commit wise.

yodaldevoid:
Rebased on master and added support for partial transfers following #510.

yodaldevoid:
I hate to be that guy, but I would like some forward movement with this sooner rather than later.

@willco007 @bagder @mback2k @vszakats

bagder:
Wouldn't these commits be better squashed into a single one for readability and easier review? It makes it hard to review when the 5 subsequent commits change things in previous commits.

I don't know Windows specific things so I cannot comment on those specifics. In general this seems fine and @yodaldevoid says he's tested it. There should be very little risk.

yodaldevoid:
I've squashed all commits as requested other than the one disabling overlapped IO. I feel that one commit should be left separate.

src/agent.c
```diff
@@ -396,6 +406,295 @@ struct agent_ops agent_ops_pageant = {
...
+    HANDLE event = NULL;
+
+    path = agent->identity_agent_path;
+    if(!path) {
+        path = getenv("SSH_AUTH_SOCK");
+        if(!path)
+            path = WIN32_OPENSSH_AGENT_SOCK;
+    }
+
+    while(1) {
```
vszakats, line 491: Suggesting `for(;;)` here to prevent some complilers warning about "condition is always true".

yodaldevoid: Good point. I've pushed up another set of patches with this corrected.

yodaldevoid:
Once again being that guy.

I have resolved the issues brought up so far and there have been no further issues brought up within the past ~2 weeks. Would you folks be ok with merging this as-is or would you rather have a third party who knows Windows to take a look at it?

@willco007 @bagder @mback2k @vszakats

willco007:
It looks OK to me at first glance, but I don't have a Windows box to test it with. It would be great if more Windows users would take a look at it.

yodaldevoid:
I'm going to have to throw in the towel on finding someone else to test or review these changes. I've poked around off and on for the past month or so on various communities where I think I might find people who would be interested to help out, but I haven't gotten any bites. I'd rather not leave this PR to languish, but if we want to get more testing before it is merged, I don't see any good way forward.

pkittenis:
I'd like to test this, and have an [existing project](https://github.com/ParallelSSH/parallel-ssh) that uses SSH agent and builds binary packages on Windows to test with.

I lack a Windows 10 machine to do the testing on at the moment, but should have some time this weekend to setup a VM for it.

yodaldevoid:
@pkittenis Do you need any help marking the changes required to test this or with setting up the test environment?

gim913:
Hi, I have similar, possibly cleaner change that I've written from scratch.
I tested it by firing docker with ssh deamon, adding authorized_keys there and running example-ssh2_agent.exe against is.

https://github.com/libssh2/libssh2/compare/master...gim913:openssh-windows

yodaldevoid:
@gim913 Would it be possible for you to test my set of patches using your docker method? Did you use any specific docker container?

I would prefer to get my set of patches merged, partially because they are more complete and partially because of how long I've waited to get these approved. That said, I don't see anything wrong with your patches, and as long as some sort of support for the Windows OpenSSH agent is added, I don't really care which set of patches are merged.

gim913:
@yodaldevoid sure, will try to do this either tomorrow or this week 👍

gim913:
@yodaldevoid I've rebased your branch on top of master and I've made following test:

I'm using image that is generated for test purposes via Dockerfile in `tests/openssh_server`
First I generated public keys for the keys that are present there:

    ssh-keygen.exe -y -f ssh_host_ed25519_key  > ssh_host_ed25519_key.pub
    ssh-keygen.exe -y -f ssh_host_rsa_key > ssh_host_rsa_key.pub
    ssh-keygen.exe -y -f ssh_host_ecdsa_key > ssh_host_ecdsa_key.pub
    
It just so happens, that `ssh_host_ed25519_key.pub` is also inside `authorized_keys`, how convenient!

Fire up the image in detatched mode:
    docker run -d --name ssh-server -p 22:22 libssh2/openssh_server
    
Add ssh-key to running agent:
    ssh-add tests\openssh_server\ssh_host_ed25519_key
    
Fire up example:
    _build\example\Debug\example-ssh2_agent.exe 127.0.0.1 libssh2

    Fingerprint: F3 CD 59 E2 91 3F 44 22 B8 0F 7B 0A 82 B2 B8 9E AE 44 93 87
    Authentication methods: publickey,password,keyboard-interactive
            Authentication with username libssh2 and public key ***** ***
            Authentication with username libssh2 and public key ***** ***
            Authentication with username libssh2 and public key will@iCube.local succeeded!
all done!

image

yodaldevoid:
Thank you @gim913! I've gone through your instructions and can confirm they worked for me, though this should be no surprise to anyone. I hope your instructions can help anyone who wants to help test this. I suppose

@pkittenis Do the instructions above help you help us test these changes?

@willco007 How, if at all, do you want to move forward with this?

yodaldevoid:
@willco007 @bagder @mback2k @vszakats

I hate to be that guy (again), but how do we want to move forward with this? Do we want to wait for more people to test this or is the testing done by @gim913 and myself enough?

On a personal note, I've been trying to get this through for months now and it is increasingly frustrating pushing on this. I completely understand the concern with Windows code when none of the maintainers have much experience with such code and no way to test it, but I feel this is mitigated due to the contained nature of the code and the testing done my myself and now @gim913 from a number of different angles. If you folks feel that further mitigation is required, I could place these changes behind a compile-time feature flag. I really want to see this support merged within the month.

vszakats:
@yodaldevoid I understand your frustration, and your patience with this is much appreciated. By the looks of it, your patch seems fine, clean and fairly isolated. As you noted, I also have no system to test this on though. One question came to mind when browsing the code, is the two new licences covering the added code snippet. Are they compatible with the libssh2 license, or pose any other concern?

Anyhow, if others have no objections, my vote for merging.

yodaldevoid:
@vszakats As noted in the top comment, I believe the licenses are compatible.

libssh2 uses a three clause BSD license while Microsoft and Markus Friedl used the two-clause BSD license. I believe these are compatible as libssh2 the license contains the same restrictions as the derived code.

The license Tatu Ylonen's code is both simpler and more complicated. It simply asks for attribution, for derivative work to be marked as such, and for implementations not compatible with the SSH RFC to not use that name. I believe how it is currently handled meets all three of these requirements, especially since this is how it is handled in the project the code is based off of.

That said, I am no lawyer, and license compatibility is a hairy subject.

gim913:
I see there's an appveyor file in the repository, are there VS-based build being made? are build results publicly avail? if so, maybe we could try to add some tests there?

yodaldevoid:
@gim913 The AppVeyor builds are publicly available, just hit "Show all checks" at the bottom of this PR and then click "details" next to the AppVeyor build. It looks like the AppVeyor builds are expecting to run testing on some DigitalOcean instance that isn't set up, so testing on Windows isn't run. I'm guessing this was done because AppVeyor does not/did not support running docker in the build instance, but I wouldn't know. This might be reason to look into moving CI to GitHub Actions or Azure.

While futzing around @gim913's tests, I found that you can spin up a separate ssh-agent in a Windows container to avoid contaminating a developer's ssh-agent setup. The problem then was getting that Windows container networked with another container running a ssh server to run the existing tests in addition to any agent tests that we could add. Networking the existing Linux based ssh-server container is probably possible, but running Windows and Linux containers side-by-side is experimental at best and unsupported at worst. I did try getting a Windows based ssh server container built, but I ran into so many issues that I eventually decided it was not worth it at the time. I could continue looking into that, but if there are any alternatives I would rather go with that.

We could add tests that don't attempt to keep the ssh-agent used for testing separate from the developer's ssh-agent with the expectation that they would mainly only be run in clean CI environments. I don't like this, but I expect it would be a stop-gap while we figure out something better.

yodaldevoid:
I have put some time into creating a test ensuring that we can authenticate using a key stored in an agent. I have found that it is more complicated than it looks at first blush (at least for me). Long story short, I don't think it is currently feasible to create a test that would work across platforms for authenticating using an agent. I could add a test that just attempts to connect to an agent, but I'm not sure of the value of that.

The first problem is adding (and removing) a key to the agent. We can call commands from C easily enough so adding keys should be easy. All OpenSSH implementations work more or less the same and generally use a program named `ssh-add` so we should be fine there. The problem comes if we would like to support `pageant` for the test as it allows adding a key through the CLI, but not removing it.

The second problem then revolves around key file permissions. At least on Windows, there is a good chance that when the repo is checked out, the SSH private keys will be readable by too many users, thus causing `ssh-add` to refuse it. In Unix land this is easy enough to fix, but in Windows land the script to fix the permissions needs to be run in an admin command prompt. This prevents us from running this script in CI. There are possibly some work-arounds for this, but I have not found any yet. There is also the possibility that this is a moot point in CI as the repo could be checked out with the right permissions, though the possibility of this is slim.

Other than those two issues, the test runs without a problem. I have not pushed the code up due to the issues I have just listed and because the Windows CI seems to not be running the tests anyway so that needs to be resolved first. I will keep the code around for a future PR.

@willco007 @bagder @mback2k @vszakats
With the above information, how would you like to move forward?

yodaldevoid:
@willco007 @bagder @mback2k @vszakats
I hate to be a broken record, but how do you folks want to move forward on this?

I see two options: block this on adding an ssh-agent test that can run on all platforms or merge it as is. As I think can be understood, I am partial to merging as is. That said if you, the maintainers, want to block this on some sort of automated test, I'd like to know that now so I can work towards that. To be clear, adding an automated test for this includes not only adding a cross-platform test (or a set of tests that are selected based on platform) but also fixing Windows tests in the CI runner as right now tests are not being run there.

Please provide some sort of direction forward. I have seen all but @mback2k voice some level of support for the code as is, and seeing as they are marked as "busy" I  can understand why they have not chimed in.

willco007:
At this point, short of writing Windows unit tests (and there are a lack of Windows devs on here), I think we should just take it. Any objections?

yodaldevoid:
I've rebased on top of the current master. Not that it was needed since there were not merge conflicts, I just felt like making this merge as smoothly as possible.

yodaldevoid:
It looks like a bunch of Linux tests failed due to Docker Hub rate-limiting. Not sure what to do with that.

mback2k:
I will try to test this during this week.

src/agent.c
```diff
@@ -396,6 +406,295 @@ struct agent_ops agent_ops_pageant = {
     agent_transact_pageant,
     agent_disconnect_pageant
 };
+
+/* Code to talk to OpenSSH was taken and modified from the Win32 port of
```
mback2k, line 410: I would like to suggest to move the copied Windows specific code to it's own dedicated source file, e.g. "agent_win.c" maybe?

mback2k (changes requested):
Except the structural code change suggested above (to make it easier to identify the code taken from another project) this looks good to me and I was successfully able to test this with the libssh2 example ssh2_agent. Thanks a lot!

src/agent.c
yodaldevoid: Should the same be done for the Putty-specific code as that is in a similar situation?

mback2k: Good idea for another PR. For now focus on what was taken from OpenSSH for Windows and please also show path and commit revision which was copied

yodaldevoid: I've noted the commit SHA, git repo, and the main files that the implementation was based on. Does this fit with your expectations?

yodaldevoid: @mback2k Do the changes made address your concerns in a way you are happy with?

src/agent.h
```diff
@@ -0,0 +1,110 @@
...
+    char *identity_agent_path; /* Path to a custom identity agent socket */
+
+#ifdef WIN32
+    OVERLAPPED overlapped;
+    HANDLE pipe;
+    BOOL pending_io;
+#endif
+};
+
+extern struct agent_ops agent_ops_openssh;
```
mback2k, line 108: Would this also need to be wrapped in `#ifdef WIN32`?

src/agent_win.c
```diff
@@ -0,0 +1,354 @@
...
+#ifdef WIN32
+/* Code to talk to OpenSSH was taken and modified from the Win32 port of
+ * Portable OpenSSH by the PowerShell team. Commit
+ * 8ab565c53f3619d6a1f5ac229e212cad8a52852c of
+ * https://github.com/PowerShell/openssh-portable.git was used as the base,
+ * specificaly the following files:
+ *
+ * - contrib\win32\win32compat\wmain_common.c
+ * - contrib\win32\win32compat\fileio.c
+ * - authfd.c
```
mback2k, line 67: So is this copied code a 1-to-1 copy or an inspiration? If individual functions were takes from these files, it could make sense which function came from which file.

src/agent.c
mback2k: Thanks again, see new comments below.

src/agent.h
yodaldevoid: It doesn't need to be, but doing so would change any use of `agent_ops_openssh` to be a compilation error rather than a linking error, which would probably be more readable.

mback2k (approved):
Yes, looks good to me. Thanks a lot! I also verified that it still works after the structural changes.

BrianInglis:
Breaks Cygwin build in snapshots since 20210512 see issue #597. 
Cygwin is a POSIX (was Solaris now Linux like) distro, and subsytem under Windows, supporting its own BSD sockets, network and ssh utilities and daemons running as Windows services, and X Window system display servers and remote clients.

mback2k:
I guess we need to add some Cygwin CI builds then.

yodaldevoid:
I am away for the next week and (probably) won't be able to work on this until then.

mback2k:
I will try to work on it this weekend.

BrianInglis:
Cygwin may be set up automatically by downloading the installers <https://cygwin.com/setup-x86{,_64}.exe> and running them against a specified nearby mirror (no smarts here) to install the base packages into specified Cygwin "root" directories for 32/64 installs. The default cross-built "native" Windows terminal is `mintty` which runs `bash` under Cygwin `pty`s, although MS `cmd` may also be used with limitations. 
Cygwin and all the distro packages (including the base) are normally built using a package called cygport (based on Gentoo portage with some Fedora package build influences from the RedHat developers who are also current volunteers), which are a library of shell scripts with a lot of builtin defaults and info about how Cygwin packages and build processes work, that checks assumptions and just "does the right thing". 
Site/install/maintainer defaults may be customized using /etc/cygport.conf  and ~/.cygport.conf. 
Each package is built using the cygport package, which is installed with all its dependencies which make up the default toolchain, and each package's own build dependencies specified in cygport (shell) variables DEPEND/BUILD_REQUIRES in each packages's $NAME.cygport shell source script. In that script, build functions as well as variables may be overriden, for subpackages as well as main packages, used when building packages with libraries, where -devel and -doc subpackages may be built from the same .cygport script, and -debuginfo subpackages are automatically generated for binary library and executable packages. 
Building and testing each package is as simple as running (under both Cygwin 64 and 32 bit install bash shells): 
```
$ cygport libssh.cygport download all check
```
where `download` downloads upstream sources into a cache directory the first time it is used, `all` runs prep (create working directory $NAME-$VERSION.$ARCH, unpack sources and apply patches), compile (run all compilation steps including autoreconfig if autotools files are present, configure/make if those are present, make if a makefile is present), install (install into a DESTDIR, and run post-installation steps), and package (create binary and source packages as compressed tar files) steps. 
For example, under my ~/src/cygwin/libssh2/ directory, I have:
```
$ l . libssh2-1.9.0-1.*86*
.:
cygwin-configure-ac.patch libssh2.cygport libssh2-1.9.0-1.i686/ libssh2-1.9.0-1.x86_64/

libssh2-1.9.0-1.i686:
build/ config/ CYGWIN-PATCHES@ dist/ inst/ log/ origsrc/ patch/ spkg/ src/ temp/

libssh2-1.9.0-1.x86_64:
build/ config/ CYGWIN-PATCHES@ dist/ inst/ log/ origsrc/ patch/ spkg/ src/ temp/
```
I've hacked up the production Cygwin [libssh2.cygport](https://github.com/libssh2/libssh2/files/6487670/libssh2.cygport.txt) to allow me to manually specify snapshots to check those builds, but the date expression should normally be used, as it compensates for time zones to download the currently available, commented out to download release tarballs, or further tweaked to use git repos as sources. 
See [Cygport docs](https://cygwin.github.io/cygport/cygport_in.html)

yodaldevoid:
We should probably keep further discussion of the cygwin issue in #597 or a new issue.

PR #520, opened 2020-09-30
Fix _libssh2_random silently discarding errors

### LINKED ISSUE #519: _libssh2_random() silently discards any error
Dear `libssh2` developers,

# The problem

`_libssh2_random()` makes it appear like it cannot fail. But it delegates its job (filling a buffer with random bytes) to the crypto backend in use, and most of them explicitely _can_ fail to provide randomness.

`_libssh2_random()` is advertized as returning `void` in `docs/HACKING.CRYPTO`. That’s a lie, actually; it generally is a macro expanding to something having an integer return code. There is no normalization/consistency on that return code. Anyway, that return is never checked, and that’s a potential issue.

Let’s review the situation for the different crypto lib backends:

## `libgcrypt`

`gcry_randomize` returns `void` and implicitely cannot fail. Good. `_libssh2_random`  in `libgcrypt.h` wraps it as:
```C
#define _libssh2_random(buf, len)                \
  (gcry_randomize ((buf), (len), GCRY_STRONG_RANDOM), 1)
```
That macro expansion “returns” as `1`. Having `1` be the code for a success is inconsistent with `libssh2` general use of `0` for that, and with what `_libssh2_random` evals to with some other backends (as we’ll see next).

`gcry_randomize` doc: https://www.gnupg.org/documentation/manuals/gcrypt/Retrieving-random-numbers.html

## `OpenSSL`

`RAND_bytes` _can_ fail. `_libssh2_random`  in `openssl.h` simply wraps it as:
```C
#define _libssh2_random(buf, len) RAND_bytes ((buf), (len))
```
But beware! `RAND_bytes` returns `0` _or  `-1`_ in case of an error (and `1` if successful). It would be better to check & normalize its return code properly.

`RAND_bytes` doc: https://www.openssl.org/docs/man1.1.1/man3/RAND_bytes.html

## `MbedTLS`

`mbedtls_ctr_drbg_random` _can_ fail. `_libssh2_mbedtls_random` in `mbedtls.c` wraps it with:
```C
int
_libssh2_mbedtls_random(unsigned char *buf, int len)
{
    int ret;
    ret = mbedtls_ctr_drbg_random(&_libssh2_mbedtls_ctr_drbg, buf, len);
    return ret == 0 ? 0 : -1;
}
```
So properly checking its return code, _but_ this time using `0` for success (not `1`!) That’s contrary to the GnuCrypt & OpenSSL cases!

Then `_libssh2_random` in `mbedtls.h` is defined as:
```C
#define _libssh2_random(buf, len) \
  _libssh2_mbedtls_random(buf, len)
```
 This macro does not reverse the return code, so the inconsistency problem is still here.

`mbedtls_ctr_drbg_random` doc : https://tls.mbed.org/api/ctr__drbg_8h.html#af6e4dd295ae790a33128562dd01c79ab

## `WinCNG`

`BCryptGenRandom` _can_ fail. `_libssh2_wincng_random` in `wincng.c` wraps it as:
```C
int
_libssh2_wincng_random(void *buf, int len)
{
    int ret;

    ret = BCryptGenRandom(_libssh2_wincng.hAlgRNG, buf, len, 0);

    return BCRYPT_SUCCESS(ret) ? 0 : -1;
}
```
Same than for `mbedtls`: it does check the result, returns `0` on success, `-1` on error.

Then `wincng.h` defines:
```C
#define _libssh2_random(buf, len) \
  _libssh2_wincng_random(buf, len)
```
End result is again that here, a `0` return denotes success, contrary to the OpenSSL & libgcrypt cases.

`BCryptGenRandom` doc: https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom

## OS400 QC3

`Qc3GenPRNs` _can_ fail. The call is wrapped in `os400qc3.c` as:
```C
void
_libssh2_random(unsigned char *buf, int len)
{
    Qc3GenPRNs(buf, len,
        Qc3PRN_TYPE_NORMAL, Qc3PRN_NO_PARITY, (char *) &ecnull);
}
```
And `os400qc3.h` simply declares:
```C
extern void     _libssh2_random(unsigned char *buf, int len);
```
This time, any error is silently discarded (`(char *) &ecnull`). Note that dixit IBM: “CPFBAF3 E : The system seed digest is not ready.” is a potential thing (i.e. `Qc3GenPRNs` does _not_ only fail in case of incorrect input parameters!)

(And, btw, there is inconsistency on the way to wrap things between the different crypto backends; here `_libssh2_random` is directly declared as a function.)

`Qc3GenPRNs` doc: https://www.ibm.com/support/knowledgecenter/ssw_ibm_i_73/apis/qc3genprns.htm

# A solution

This is what I intend to do & pull-request in the following hours/day:

- Advertise, in `HACKING.crypto`, `_libssh2_random` as returning an `int`, not `void`. That’s actually the case already, _except_ in the OS400 QC3 case.
- Normalize its return code to be **`0` on success**. On failure, I’ll see; probably make it explicitely `-1` (only).
- Properly check that return code everywhere `_libssh2_random` is called.

What I will **_not_** do:
- Deal with the error discarding in OS400 QC3. Doing so seems to warrant a bigger refactoring where QC3 errors in general (not just for `Qc3GenPRNs`) would be checked.
- Fix the naming/macro-vs-function inconsistency noted above (it has no functional impact.)

Please let me know if anything said here sounds wrong or unreasonable. Otherwise, wait for the PR! :-)

PaulCapron:
This fixes #519. Please see the issue & commit messages for more info.

### DIFF
--- a/docs/HACKING.CRYPTO
+++ b/docs/HACKING.CRYPTO
@@ -897,5 +897,6 @@ In example, this is needed to preset unused structure slacks on platforms
 requiring it.
 If this is not needed, it should be defined as an empty macro.
 
-void _libssh2_random(unsigned char *buf, int len);
+int _libssh2_random(unsigned char *buf, int len);
 Store len random bytes at buf.
+Returns 0 if OK, else -1.
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -505,6 +505,7 @@ typedef struct _LIBSSH2_POLLFD {
 #define LIBSSH2_ERROR_KNOWN_HOSTS               -46
 #define LIBSSH2_ERROR_CHANNEL_WINDOW_FULL       -47
 #define LIBSSH2_ERROR_KEYFILE_AUTH_FAILED       -48
+#define LIBSSH2_ERROR_RANDGEN                   -49
 
 /* this is a define to provide the old (<= 1.2.7) name */
 #define LIBSSH2_ERROR_BANNER_NONE LIBSSH2_ERROR_BANNER_RECV
--- a/src/channel.c
+++ b/src/channel.c
@@ -1338,7 +1338,11 @@ channel_x11_req(LIBSSH2_CHANNEL *channel, int single_connection,
                border */
             unsigned char buffer[(LIBSSH2_X11_RANDOM_COOKIE_LEN / 2) + 1];
 
-            _libssh2_random(buffer, LIBSSH2_X11_RANDOM_COOKIE_LEN / 2);
+            if(_libssh2_random(buffer, LIBSSH2_X11_RANDOM_COOKIE_LEN / 2)) {
+                return _libssh2_error(session, LIBSSH2_ERROR_RANDGEN,
+                                      "Unable to get random bytes "
+                                      "for x11-req cookie");
+            }
             for(i = 0; i < (LIBSSH2_X11_RANDOM_COOKIE_LEN / 2); i++) {
                 snprintf((char *)&s[i*2], 3, "%02X", buffer[i]);
             }
--- a/src/kex.c
+++ b/src/kex.c
@@ -3177,7 +3177,11 @@ static int kexinit(LIBSSH2_SESSION * session)
 
         *(s++) = SSH_MSG_KEXINIT;
 
-        _libssh2_random(s, 16);
+        if(_libssh2_random(s, 16)) {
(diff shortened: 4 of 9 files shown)


doorsdown:
This looks good

Issue #527, opened 2020-10-26
libssh2_session_handshake() creates latent OpenSSL error in FIPS mode


MarcT512:
**Describe the bug**
In OpenSSL FIPS certified cryptography mode, calling libssh2_session_handshake() creates a latent OpenSSL error of the form:
`139663755794080:error:060A80A3:digital envelope routines:FIPS_DIGESTINIT:disabled for fips:fips_md.c:180:`

This error is picked up by the next call to OpenSSL ERR_print_errors(), etc.

**To Reproduce**

1. Build libssh2 against a FIPS enabled OpenSSL 1.0.2 release.
NB: There is no FIPS in OpenSSL 1.1.x, but there should be a FIPS provider in the forthcoming OpenSSL 3.0 release.
2. call libssh2_session_handshake()
3. call ERR_print_errors()

Result is:
`139663755794080:error:060A80A3:digital envelope routines:FIPS_DIGESTINIT:disabled for fips:fips_md.c:180`

**Expected behavior**
No latent OpenSSL error should be created.

**Version (please complete the following information):**
 - OS: Linux (any)
 - libssh2 version: latest git release (commit: 6c7769dcc422250d14af1b06fce378b6ee009440)
 - OpenSSL v1.0.2o & openssl-fips-2.0.16

**Additional context**
This happens because the MD5 digest is not permitted in FIPS certified cryptography mode. libssh2 tries to init MD5 in _libssh2_md5_init().

I have a "works for me" patch that avoids initting the MD5 digest if FIPS is enabled. I don't know what the implications of this are to SSH in general, but I've seen no ill effects on the systems I've tested against.

PR #528, opened 2020-10-26
Avoid OpenSSL latent error in FIPS mode

MarcT512:
Avoid initing MD5 digest, which is not permitted in OpenSSL FIPS certified cryptography mode.
Fixes: https://github.com/libssh2/libssh2/issues/527


MarcT512:
Done.

willco007:
Looking at where `libssh2_md5_init` is called, there could be some unintended side effects of disabling MD5 resulting in some errors that you may not expect. I suppose if you're using OpenSSL with FIPS mode enabled you should already be aware of these limitations? My other thought is abstracting the FIPS check out into a generic libssh2 function you can call from other places in the library to provide a better error message to the user rather than simply MD5 init failed.

PR #547, opened 2021-01-13
Option to build both static and shared libraries

berney:
This PR updates CMake to have a new option `BUILD_BOTH_LIBS` that if enabled both a static and a shared library will be built.
CMake is also refactored to use an object library, so that source files are only compiled once when building both targets.

When building both static and shared libraries, `POSITION_INDEPENDENT_CODE` is enabled as it is required. This means the static library will also be PIC in this case. If both `BUILD_BOTH_LIBS` and `BUILD_SHARED_LIB` are off, then the static library will be built without PIC.


berney:
I think this PR is ready. The tests are passing now. Please review to ensure the changes are correct.

willco007:
Any comments on this? Seems OK but I'm not a cmake expert by any means.

stac47:
Hello,
Thanks for this great PR. It is definitely an excellent feature which should be activated by default to align with the autotools build.
I just test it on linux system and it worked like a charm.
My only concern is that the archive will contain PIC code which could (will) be less efficient than before.
Regards,

berney:
> Hello,
> Thanks for this great PR. It is definitely an excellent feature which should be activated by default to align with the autotools build.

You're welcome.

> I just test it on linux system and it worked like a charm.

Thanks for confirming with your own testing.

> My only concern is that the archive will contain PIC code which could (will) be less efficient than before.

Yes, I had noted this as well. This will only be the case when building both static and dynamic, if just building static it will not be PIC.

| `BUILD_SHARED_LIB` | `BUILD_BOTH_LIBS` | Result                           |
| ----------------------- | --------------------- | ------------------------ |
| false                           |  false                       | static (non-PIC)           |
| true                            | false                       | dynamic                       |
| false                           | true                        | static (PIC) + dynamic  |
| true                            | true                        | static (PIC) + dynamic |

If non-PIC static is required (or desired) it will need to be compiled alone from any dynamic target.

berney:
Any decision on this PR?

willco007:
@berney Unfortunately, as you probably were notified, this no longer builds cleanly with the new keyboard interactive code. I do apologize for not getting this merged in sooner. I'd be happy to expedite a fixed PR.

Issue #463, opened 2020-04-27
libssh2_session_handshake returns -44  with OpenSSL crypto backend


SX-80:
The function libssh2_session_handshake returns -44 "Unable to ask for ssh-userauth service" in code using OpenSSL crypto backend.

```
    // ... initialize socket ...

    LIBSSH2_SESSION* session = libssh2_session_init();
    int res = libssh2_session_handshake( session, socket );

    char const* str = nullptr;
    int len = 0;
    libssh2_session_last_error( session, &str, &len, false );

    std::cout << "res: " << res << ", error: " << str << std::endl;
```

Error is found in function ''_libssh2_cipher_crypt" (openssl.c: 416)
The problem appears after commit in OpenSSL:
https://github.com/openssl/openssl/commit/f7397f0d58ce7ddf4c5366cd1846f16b341fbe43

**Version**
 - OS: Windows 7
 - libssh2 version [1.9.0_DEV]              - master on 2020-04-27
 - openssl version [3.0.0-alpha2-dev]  - master on 2020-04-27

willco007:
Have you tried a shipping version of OpenSSL? Seems like a non-zero chance this is a bug in the alpha release of OpenSSL you're using.

SX-80:
Corrected version
```
int
_libssh2_cipher_crypt(_libssh2_cipher_ctx * ctx,
                      _libssh2_cipher_type(algo),
                      int encrypt, unsigned char *block, size_t blocksize)
{
    unsigned char buf[EVP_MAX_BLOCK_LENGTH];
    int ret;
    (void) algo;
    (void) encrypt;

#ifdef HAVE_OPAQUE_STRUCTS
    ret = EVP_Cipher(*ctx, buf, block, blocksize);
#else
    ret = EVP_Cipher(ctx, buf, block, blocksize);
#endif
    if(0 < ret) {
        memcpy(block, buf, blocksize);
    }
    return 0 < ret ? 0 : 1;
}

SX-80:
The changes in OpenSSL library
![изображение](https://user-images.githubusercontent.com/64437815/80421138-a9264500-88e4-11ea-9b74-177eeaeee56e.png)


Now returns 'outl'  or '-1'

SX-80:
![изображение](https://user-images.githubusercontent.com/64437815/80421378-0cb07280-88e5-11ea-92fb-ad9e735db51f.png)

willco007:
[...]

PR #556, opened 2021-02-23
Fix EVP_Cipher interface change in openssl 3

stac47:
Hello,
We discovered that bug when we linked curl with openssl version 3. It has been reported in #463. 
I propose this fix.
Regards,


willco007:
This is discussed in issue #463.

willco007:
The define `OPENSSL_VERSION_MAJOR` is not defined on older versions of OpenSSL. This fix isn't going to work as written.

stac47:
Ok I will check if this macro is defined first.
Thanks for your feedback.
Regards,

PR #557, opened 2021-02-25
[tests] Allow the tests to run inside a container

stac47:
Hello,

The current tests suite starts SSH server as OCI container. This commit
add the possibility to run the tests in a container provided that:

* the docker client is installed builder container
* the host docker daemon unix socket has been mounted in the builder
  container (with, if needed, the DOCKER_HOST environment variable
  accordingly set, and the permission to write on this socker)
* the builder container is run on the default bridge network, or the
  host network. This PR does not handle the case where the builder
  container is on another network.

Notes about the implementation:
* I used `getline` function which has been standardized in 2008 (POSIX). It maybe does not reach the portability criteria but as it aims at being built in a docker container it should be ok. Tell me if you prefer `#ifdef __linux__` or if you prefer a more standard implementation.
* Tell me if anything breaks the coding rules of this project, I will fix.

Regards,

### DIFF
--- a/tests/openssh_fixture.c
+++ b/tests/openssh_fixture.c
@@ -178,6 +178,34 @@ static const char *docker_machine_name(void)
     return getenv("DOCKER_MACHINE_NAME");
 }
 
+static int is_running_inside_a_container()
+{
+#ifdef WIN32
+    return 0;
+#else
+    const char *cgroup_filename = "/proc/self/cgroup";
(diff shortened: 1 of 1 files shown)


mback2k:
@stac47 can I kindly ask you to rebase this PR on current master and force-push? Thanks in advance!

mback2k (approved):

PR #566, opened 2021-03-10
Add some comments on the command basis

tsengjun:
ssh2_exec.c
Note: Add some comment on handling command with spaces; and how to deal with stderr.


tsengjun:
Close as it is not constructive. Better to add these comments into the help/API document.

Thanks!

Issue #576, opened 2021-04-15
Timeout during keyboard-interactive login


teottin:
When I attempt keyboard interactive login (Azure AD 2FA login) and use more than 60 seconds to complete the login, the connection fails.
There seems to be a bug in the _libssh2_packet_requirev function (see my comment in the code below):
  if(strchr((char *) packet_types, ret)) {
      /* Be lazy, let packet_ask pull it out of the brigade */
      /* I think the timeout should be reset before returning here (i.e. state->start = 0;)
      return _libssh2_packet_askv(session, packet_types, data,
                                  data_len, match_ofs, match_buf,
                                  match_len);
  }

I think the timer is started on the first call to this function, and then when the users answers the interactive login, this function is called again but the timer is already elapsed.

teottin:
The _libssh2_packet_require function does almost the same as _libssh2_packet_requirev but this function sets state->start = 0 before returning.

willco007:
Fixed with #586, thanks!

PR #577, opened 2021-04-15
packet.c: Reset read timeout when successfully received a packet (#576)

teottin:


teottin:
Created a new pull request based on 1.8.x

PR #581, opened 2021-04-27
ci: Swap from Travis to Github Actions

yodaldevoid:
I saw in the mailing list that there was some desire to move from Travis (and maybe Appveyor) to Github Actions. I also wanted to do this in the hope that it will cut down CI times and spurious CI failures that I have seen over and over.

To have the workflows I've added run for this PR, I believe a maintainer will need to approve the workflow for this PR because I am not yet marked as a contributor.
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

In testing I've seen Linux builds ~8 minutes in length.

Notes:
- `ubuntu-16.04` was selected as the virtual environment to run on as this matches what Travis has been using.
- For 32 builds, `zlib1g-dev:i386` needed to be installed for whatever reason. I'm guessing this was installed by the `chrome` addon on Travis, but I honestly don't know.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,4 +99,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Run Fuzzer
-        run: GIT_BRANCH=${GITHUB_REF#refs/heads/} ./ci/ossfuzz.sh
+        run: GIT_REF=$GITHUB_REF ./ci/ossfuzz.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,4 +99,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Run Fuzzer
-        run: TRAVIS_BRANCH=${GITHUB_REF#refs/heads/} ./tests/ossfuzz/travisoss.sh
+        run: GIT_BRANCH=${GITHUB_REF#refs/heads/} ./ci/ossfuzz.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,102 @@
+name: CI
+
+on:
+  push:
+    branches: [ $default-branch ]
+  pull_request:
+    branches: [ $default-branch ]
+
+jobs:
+  style-check:
+    runs-on: ubuntu-16.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check Style
+        run: ./ci/checksrc.sh
+  build:
+    runs-on: ubuntu-16.04
+    strategy:
+      fail-fast: false
+      matrix:
+        compiler: [gcc, clang]
+        address_size: [32, 64]
+        crypto_backend: [OpenSSL, Libgcrypt, mbedTLS]
+        build_shared_libs: [OFF, ON]
+        enable_zlib_compression: [OFF, ON]
+        b: [cmake]
+        include:
+          - compiler: gcc
+            address_size: 64
+            crypto_backend: OpenSSL
+            build_shared_libs: OFF
+            enable_zlib_compression: OFF
+            b: configure
+          - compiler: clang
+            address_size: 64
+            crypto_backend: OpenSSL
+            build_shared_libs: OFF
+            enable_zlib_compression: OFF
+            b: configure
+    env:
+      CC: ${{ matrix.compiler }}
+      CC_FOR_BUILD: ${{ matrix.compiler }}
+      CRYPTO_BACKEND: ${{ matrix.crypto_backend }}
+      BUILD_SHARED_LIBS: ${{ matrix.build_shared_libs }}
+      ENABLE_ZLIB_COMPRESSION: ${{ matrix.enable_zlib_compression }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install 32 Bit Dependencies
+        if: ${{ matrix.address_size == 32 }}
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt-get update -qq
+          sudo apt-get install -y gcc-multilib
+          sudo apt-get install -y libssl-dev:i386 libgcrypt20-dev:i386 zlib1g-dev:i386 build-essential gcc-multilib
+          sudo dpkg --purge --force-depends gcc-multilib
+          sudo dpkg --purge --force-depends libssl-dev
+          echo "TOOLCHAIN_OPTION=-DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchain-Linux-32.cmake" >> $GITHUB_ENV
+      - name: Install 64 Bit Dependencies
+        if: ${{ matrix.address_size == 64 }}
+        run: |
+          sudo apt-get install -y libssl-dev
+          sudo apt-get install -y libgcrypt11-dev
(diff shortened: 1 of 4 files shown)


yodaldevoid:
This is ready from a Linux CI standpoint, but no work has been done for Windows.

Windows is a little complicated because the current Appveyor run tests Visual Studio 12 2013 and Visual Studio 14 2015, both of which are not available on GitHub Actions. Only Visual Studio 15 2017 and Visual Studio 16 2019 are available. What should be done here?

It would definitely be nice to move the Windows CI to GitHub Actions as it would remove the need to rely on Digital Ocean to run the docker containers.

mback2k:
I have other plans for the AppVeyor CI builds and would like to keep them in addition to more modern Windows builds. I am waiting on @alamaison to turn the AppVeyor project into an org.

mback2k:
￼￼Generally speaking I would like to follow the way the curl project does CI, e.g.:
- AppVeyor for legacy Windows
- GitHub Actions for Linux and macOS
- Azure Pipelines for modern Windows

yodaldevoid:
@mback2k What is the reasoning behind using Azure Pipelines for modern Windows builds rather than also using GitHub Actions for that?

mback2k:
Primarily balancing the load to get lower response times.

yodaldevoid:
I'm not sure response times would be a problem if the modern Windows builds were on GitHub Actions. The 50 odd builds generated for Linux only take ~8 minutes right now and adding a handful of Windows builds should not effect that much. While load balancing might be a concern if the build and test suite starts taking an order of magnitude longer, I don't think the downside of having to maintain yet another CI system right now is outweighed by load balancing concerns in the near to medium future.

That said, it's your decision and I won't argue the point further

yodaldevoid:
At any rate, if I'm only futzing with formerly Travis builds in this PR, I don't have any further changes to make. I believe the only remaining step is to approve the actions to run on this PR right now (assuming that is what has caused them not to run so far) in order to test them on this repo.

mback2k:
Okay, I will check this PR once I am back at a browser. Currently on mobile there approval is not possible.

willco007:
Thanks so much for doing this! Some-what ironically Appveyor builds failed due to Docker being down. If we're voting, I'd vote to move the Windows tests to Github while continuing the legacy tests on Appveyor. My thoughts are, the fewer services we need to integrate with the easier it is to maintain in the future. That being said, I don't feel strongly about it and I'm not a Windows guy so I'll defer to our Windows users.

.github/workflows/test-builds-linux.yml
```diff
@@ -0,0 +1,102 @@
+name: Linux Test Builds
+
+on:
+  push:
+    branches: [ $default-branch ]
+  pull_request:
+    branches: [ $default-branch ]
```
mback2k, line 7: It seems like this is not working, at least I have not way of "approving" this workflow.

.github/workflows/test-builds-linux.yml
```diff
@@ -0,0 +1,102 @@
+name: Linux Test Builds
```
mback2k, line 1: I would shorten this to maybe just "CI", because the individual job names in the file will be appended.

.github/workflows/test-builds-linux.yml
```diff
@@ -0,0 +1,102 @@
...
+            b: configure
+          - compiler: clang
+            address_size: 64
+            crypto_backend: OpenSSL
+            build_shared_libs: OFF
+            enable_zlib_compression: OFF
+            b: configure
+    env:
+      CC: ${{ matrix.compiler }}
+      CC_FOR_BUILD: ${{ matrix.compiler }}
```
mback2k, line 42: What is the purpose of this variable?

mback2k:
I guess we can give running all modern tests on GitHub and keem Azure in mind if we run into CI congestion issues.

I hope to be able to fix the AppVeyor builds soon, because those are very important to keep testing the legacy WinCNG support available between Windows Vista and Windows 10 1903+.

.github/workflows/test-builds-linux.yml
yodaldevoid: I was matching Travis's `compiler` option behavior. They set both `CC` and `CC_FOR_BUILD` with the compiler selected. I don't think `CC_FOR_BUILD` is actually used anywhere, but I didn't want to take any chances.

https://docs.travis-ci.com/user/languages/c/#choosing-compilers-to-test-against

.github/workflows/test-builds-linux.yml
yodaldevoid: After doing some testing on my fork, it looks like the workflow will not be run if the workflow is not defined on the target branch of a PR. Once the workflow is defined in the target branch, any changes made by a future PR will effect the workflow during those future PRs.

e.g. I have the branches `$default-branch` (`master` in this case) and `github_ci_test` in the allowed list of branches for the `pull_request` event. If I open a PR of `github_ci_test` onto `master`, the workflow will not be run. If I open a PR of `github_ci_test_test` onto `github_ci_test` with some random code change, the workflow is run. If I open a PR of `github_ci_test_test` onto `github_ci_test` with a change removing `github_ci_test` from the allow list, the workflow does not run.

Once again I seem to have run into a case where I don't have a good way for you folks to test my changes at the moment. I have found https://github.com/nektos/act which allows you to run a workflow locally, but this is not ideal. I will continue looking into this.

A slight side note, I can't say I'm very happy with this "feature". I get that it is a security feature to prevent random PRs form untrusted people from leaking secrets, but it makes it a pain to switch over to. I can't say I have ever run into anything like this on another platform.

yodaldevoid: Looking at other projects, I think the path forward here is to add GitHub Actions as a CI path without removing Travis. Afterwords, if the workflow is working as expected the old Travis configuration can be removed.

.github/workflows/test-builds-linux.yml
yodaldevoid: https://www.linuxtopia.org/online_books/linux_development_tools/libtool_autoconf_automake_book/autobook_270.html

This explains what `CC_FOR_BUILD` is doing. I see no reason to keep it as we are not cross compiling, but at the same time it's doing no harm, so I'll leave the choice up to you.

mback2k: Okay, thanks for explaining this.

.github/workflows/test-builds-linux.yml
mback2k: Good idea. Would it be possible for you to run the latest version of these builds on your forked repo for us to see the results?

yodaldevoid: Here is an example PR to show the results.

https://github.com/yodaldevoid/libssh2/pull/5

yodaldevoid:
I noticed that the fuzzer has so far been running on what I assume was an unintended branch. How it was set up, the fuzzer was always set to run on the target branch of a PR, not the branch getting merged in.

I've been able to fix and test this for the GitHub Actions workflow already and we'll see if my fix also works for Travis.

mback2k (approved):
Looks good to me, thank you very much!

mback2k:
@willco007 if you agree on the approach, would you be able to merge this please?

PR #585, opened 2021-05-02
Explicitly state the default branch it GitHub Action CI

yodaldevoid:
It looks like the $default-branch macro only works in templates, not
workflows. This is not explicitly stated anywhere except the linked PR
comment.

https://github.com/actions/starter-workflows/pull/590#issuecomment-672360634

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ $default-branch ]
+    branches: [ master ]
   pull_request:
-    branches: [ $default-branch ]
+    branches: [ master ]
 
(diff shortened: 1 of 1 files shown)


willco007:
@yodaldevoid After this commit landing PRs today are all failing with a can't build docker image. Any chance you'd be willing to give one of them a peak and see if anything odd is going on?

yodaldevoid:
Yeah, I'll take a look

yodaldevoid:
@willco007 It looks like the tests are probably failing because the docker image takes too long to build when the servers are under a significant load. This is causing the tests to reach their time limit. I'll look into what can be done for caching the docker image layers.

PR #587, opened 2021-05-04
ci: Speed up docker builds for tests

yodaldevoid:
The OpenSSH server docker image used for tests is pre-built to prevent
wasting time building it during a test, and unneeded rebuilds are
prevented by caching the image layers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
       ENABLE_ZLIB_COMPRESSION: ${{ matrix.enable_zlib_compression }}
     steps:
       - uses: actions/checkout@v2
+      - uses: satackey/action-docker-layer-caching@v0.0.11
       - name: Install 32 Bit Dependencies
         if: ${{ matrix.address_size == 32 }}
         run: |
@@ -85,6 +86,9 @@ jobs:
           cd bin
           cmake $TOOLCHAIN_OPTION -DCRYPTO_BACKEND=$CRYPTO_BACKEND -DBUILD_SHARED_LIBS=$BUILD_SHARED_LIBS -DENABLE_ZLIB_COMPRESSION=$ENABLE_ZLIB_COMPRESSION ..
           cmake --build .
+          pushd ../tests
+          docker build -t libssh2/openssh_server openssh_server
+          popd
           CTEST_OUTPUT_ON_FAILURE=1 cmake --build . --target test
           cmake --build . --target package
   fuzzer:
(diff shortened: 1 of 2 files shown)


yodaldevoid:
@willco007 This should fix the build issues.

mback2k:
@yodaldevoid AppVeyor seems to be failing, because it does not have Docker daemon support on the CI itself, so @alamaison decided to use DigitalOcean for that which seems no longer to be accessible from the AppVeyor CI. My personal plan is to replace DigitalOcean with the free tier of fly.io to host the SSH daemon on demand for use during AppVeyor CI builds. Still waiting on @alamaison to turn his AppVeyor libssh2 project into an org and/or invite me to manage it.

yodaldevoid:
I noticed that the remote docker-machine support was failing on Appveyor due to an unrelated reason to what this PR aims to solve. What I meant was that if this caching thing is an issue for the Appveyor builds once whatever else is causing failures in resolved, I (or someone else) would need to look into solutions as I put no work into caching or pre-building for the Appveyor builds.

For what it's worth, I understand why DigitalOcean is used for Appveyor builds. I have actually been looking at the possibility of adding Windows builds under GitHub Actions and have slowly come to realize that if we want to test against a fully-featured OpenSSH server for Windows builds, something like connecting to a remote DigitalOcean server is going to be needed, no matter where the CI runs. At any rate, that is a discussion for another PR.

willco007:
Thanks again for doing all this work; it's a huge help!

willco007:
Bah, still getting a transient 429 error (too many requests) from docker.

yodaldevoid:
I think the only ways to not get rate limited by Docker Hub at this point is either to store the test image in a private docker hub or to use a docker login so we are not lumped in with everyone else using CI without a login. If we use a login, there should be no reasonable way for us to hit the rate limit. I am surprised the caching is hitting the Docker Hub, though.

mback2k:
We could also use GitHub Container Registry instead.

yodaldevoid:
Looking at it closer, it actually looks like what we are hitting the rate limit on is not Docker Hub but the caching service. We could try to reduce concurrency to limit the number of requests made to the cache service.

mback2k:
So what about instead of caching layers, using a multi-stage CI to first build and push the container to GutHub Container Registry and then just consume that during the actual test run?

yodaldevoid:
> So what about instead of caching layers, using a multi-stage CI to first build and push the container to GutHub Container Registry and then just consume that during the actual test run?

That definitely seems like a reasonable solution. I'll see about throwing something together later today.

PR #588, opened 2021-05-05
ci: Use the GitHub Docker registry rather than caching

yodaldevoid:
This uses the docker registry rather than the newer container registry because the container registry needs to be manually enabled for a user/organization while it is still in beta. I obviously don't have the permissions to do that for libssh2, but if a maintainer would like to do that, I can try to move this over to the container registry.

https://docs.github.com/en/packages/working-with-a-github-packages-registry/enabling-improved-container-support-with-the-container-registry

Utilizing the "cached" image from Travis builds would require a personal access token when using the docker registry, which I don't have the permissions to make for the libssh2 organization. This would not be a problem for the container registry.

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-docker-registry

Pulling images won't work on the first pass for obvious reasons. Pushing images won't work in the PR as $GITHUB_TOKEN doesn't have the required permissions for pushing for forked repos, which applies to PRs from forked repos as well. If we wish to test pulling from the repository in this PR, a maintainer will need to upload the openssh_server image to the docker registry and then the CI will need to be re-run.

As a side note, I hate how much of this CI stuff I've been working on cannot be fully tested before merging, and it is mostly because of the security measures that GitHub forces on us.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,6 @@ jobs:
       ENABLE_ZLIB_COMPRESSION: ${{ matrix.enable_zlib_compression }}
     steps:
       - uses: actions/checkout@v2
-      - uses: satackey/action-docker-layer-caching@v0.0.11
       - name: Install 32 Bit Dependencies
         if: ${{ matrix.address_size == 32 }}
         run: |
@@ -72,6 +71,22 @@ jobs:
           cd ..
           echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/usr/lib" >> $GITHUB_ENV
           echo "TOOLCHAIN_OPTION=$TOOLCHAIN_OPTION -DCMAKE_PREFIX_PATH=$PWD/usr" >> $GITHUB_ENV
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Prepare OpenSSH Server Container for Testing
+        uses: docker/build-push-action@v2
+        with:
+          pull: true
+          load: true
+          push: ${{ github.event_name == 'push' }}
+          context: ./tests/openssh_server
+          tags: |
+            libssh2/openssh_server
+            ghcr.io/libssh2/ci_tests_openssh_server
       - name: Build with Configure
         if: ${{ matrix.b == 'configure' }}
         run: |
@@ -86,10 +101,7 @@ jobs:
           cd bin
           cmake $TOOLCHAIN_OPTION -DCRYPTO_BACKEND=$CRYPTO_BACKEND -DBUILD_SHARED_LIBS=$BUILD_SHARED_LIBS -DENABLE_ZLIB_COMPRESSION=$ENABLE_ZLIB_COMPRESSION ..
           cmake --build .
-          pushd ../tests
-          docker build -t libssh2/openssh_server openssh_server
-          popd
-          CTEST_OUTPUT_ON_FAILURE=1 cmake --build . --target test
+          DISABLE_DOCKER_BUILD=1 CTEST_OUTPUT_ON_FAILURE=1 cmake --build . --target test
           cmake --build . --target package
   fuzzer:
     runs-on: ubuntu-16.04
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,6 +113,7 @@ script:
       if [ "$B" = "cmake" ]; then
         mkdir bin
         cd tests
+        (docker pull ghcr.io/libssh2/ci_tests_openssh_server && docker tag ghcr.io/libssh2/ci_tests_openssh_server libssh2/openssh_server) || true
         docker build -t libssh2/openssh_server openssh_server
         cd ../bin
         cmake $TOOLCHAIN_OPTION -DCRYPTO_BACKEND=$CRYPTO_BACKEND -DBUILD_SHARED_LIBS=$BUILD_SHARED_LIBS -DENABLE_ZLIB_COMPRESSION=$ENABLE_ZLIB_COMPRESSION .. && cmake --build . && CTEST_OUTPUT_ON_FAILURE=1 cmake --build . --target test && cmake --build . --target package
--- a/tests/openssh_fixture.c
+++ b/tests/openssh_fixture.c
@@ -145,8 +145,14 @@ static int run_command(char **output, const char *command, ...)
 
 static int build_openssh_server_docker_image(void)
 {
-    return run_command(NULL, "docker build -t libssh2/openssh_server "
-                             "openssh_server");
+    int ret = 0;
+
+    if(NULL == getenv("DISABLE_DOCKER_BUILD")) {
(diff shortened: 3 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -72,6 +71,17 @@ jobs:
           cd ..
           echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/usr/lib" >> $GITHUB_ENV
           echo "TOOLCHAIN_OPTION=$TOOLCHAIN_OPTION -DCMAKE_PREFIX_PATH=$PWD/usr" >> $GITHUB_ENV
+      - name: Prepare OpenSSH Server Container for Testing
```
mback2k, line 74: If this there it's own task, it could be done just once for the whole matrix.

.github/workflows/ci.yml
```diff
@@ -72,6 +71,17 @@ jobs:
...
+        run: |
+          cd tests
+          set +eo pipefail
+          docker login https://docker.pkg.github.com -u libssh2 -p ${{ secrets.GITHUB_TOKEN }}
```
mback2k, line 78: I think docker provides ready to use actions for this.

mback2k (changes requested):
I will enable the registry functionality later today.

.github/workflows/ci.yml
yodaldevoid: For the container to actually be available during the build run we would still need to perform a pull/build action for each build run. We could perform a separate task to build and cache it once to prevent the first few builds from duplicating effort, but this case will only come up when we have an empty cache (this PR) or if anything ever changes the Dockerfile or it's build context, which should be fairly rare.

.github/workflows/ci.yml
yodaldevoid: It seems there is not one by docker, but there is one made by another user.

https://github.com/marketplace/actions/build-docker-images-using-cache

mback2k: Did you see this? https://www.docker.com/blog/docker-github-actions/

.github/workflows/ci.yml
mback2k: You are right, it would only actually do something the first time or if the Dockerfile changes, but the matrix builds then only need a pull, not a build. ;-)

mback2k:
I just enabled improved Docker registry support for the libssh2 org on GitHub. Since the package name "openssh_server" could lead people to think that this something to be used for production purposes, I would also recommend to change the Docker image name from "openssh_server" to maybe something like "ci_tests_openssh_server".

.github/workflows/ci.yml
yodaldevoid: The caching they do in that step is more-or-less identical to the caching action that this PR replaces, just done manually. They also do not perform a pull as this PR does as they rely on the cache.

.github/workflows/ci.yml
yodaldevoid: It would still need the build for at the very least this PR as due to the GitHub security policies. As mentioned in the top comment, PRs from forked repos (such as this PR) will not be able to push a container to the repo. This means that for this PR where we don't have a container already cached the separate task would not be able push the container it built and the build tasks would have nothing to pull, necessitating a build.

https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

.github/workflows/ci.yml
mback2k: I meant using `uses: docker/login-action@v1` and `uses: docker/build-push-action@v2` instead of the `docker` commands.

.travis.yml
```diff
@@ -113,6 +113,7 @@ script:
       if [ "$B" = "cmake" ]; then
         mkdir bin
         cd tests
+        docker pull ghcr.io/libssh2/ci_tests_openssh_server || true
```
mback2k, line 116: The next line needs to be adjusted to use the same image name, otherwise it will create a second image. You could also re-tag the pulled image locally to the existing internal image name with `docker tag`.

yodaldevoid: It will create a new image, but that should not be a problem because the new image will use the pulled image's cached layers.

.github/workflows/ci.yml
```diff
@@ -72,6 +71,28 @@ jobs:
           cd ..
           echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/usr/lib" >> $GITHUB_ENV
           echo "TOOLCHAIN_OPTION=$TOOLCHAIN_OPTION -DCMAKE_PREFIX_PATH=$PWD/usr" >> $GITHUB_ENV
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Docker Buildx
```
mback2k, line 80: Does Buildx really help us somehow? Just asking, because I have not used/switched to it (yet).

.github/workflows/ci.yml
```diff
@@ -72,6 +71,28 @@ jobs:
...
+        with:
+          driver: docker
+          id: builder
+      - name: Prepare OpenSSH Server Container for Testing
+        uses: docker/build-push-action@v2
+        with:
+          builder: builder
+          pull: true
+          load: true
+          push: ${{ github.event_name == 'push' }}
```
mback2k, line 91: Do I understand it correctly that this makes it only push the image for commits on master branch?

yodaldevoid: Correct

.github/workflows/ci.yml
yodaldevoid: I was trying `docker/setup-buildx-action@v1` as it seems the `docker/build-push-action@v2` works in a container meaning that the image is not available to the following steps and using `docker/setup-buildx-action@v1` is listed as the way to get it to not run in a container. `load: true` should have handled that already, but it didn't seem to be working.

.github/workflows/ci.yml
yodaldevoid: I swapped over to `docker/login-action@v1` and `docker/build-push-action@v2` from `whoan/docker-build-with-cache-action@v5`. It required adding an option to skip `docker build` in the tests themselves for whatever reason, but it works now.

.github/workflows/ci.yml
```diff
@@ -72,6 +71,22 @@ jobs:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Prepare OpenSSH Server Container for Testing
+        uses: docker/build-push-action@v2
+        with:
+          pull: true
+          load: true
+          push: ${{ github.event_name == 'push' }}
+          context: ./tests/openssh_server
+          tags: |
+            libssh2/openssh_server
```
mback2k, line 88: Just a heads up: it could be that this will cause the push to fail, because only the 2nd tag can be pushed.

yodaldevoid:
This is going to sound a little crazy after how much work I've already put into this, but does it even make sense to cache the docker image? The docker build does not take long (it can be short enough to fit in the `ctest` timeout) and caching in any manner has show itself to be difficult and possibly flaky. Basically, is caching more trouble than it is worth?

I propose we just remove the caching and leave the docker image pre-building.

PR #589, opened 2021-05-07
ci: Remove caching of docker image layers

yodaldevoid:
This PR removes the caching of docker image layers entirely to avoid CI failures due to the caching action failing. Honestly, the OpenSSH test server docker image built takes so little time to build that caching the layers doesn't get us much.

An alternative to the work in #588.


mback2k (approved):

willco007:
I'll defer to what you both feel like is the better solution here and get it landed.

yodaldevoid:
@willco007 I think this is the better solution in the short term. I doubt we will ever *need* to cache the docker layers, but if the CI times start getting too long and the docker build step is the easiest solution, we can revisit caching.

PR #594, opened 2021-05-13
configure.ac: don't undefine scoped variable

willco007:
To get this script to run with Autoconf 2.71 on macOS I had to remove the undefine of the backend for loop variable. It seems scoped to the for loop and also isn't referenced later in the script so it seems OK to remove it.


willco007:
Tested with Autoconf 2.69 on Fedora as well, worked without issue.

PR #599, opened 2021-05-15
dist: add src/agent.h

bagder:
Needed by (some) builds.


yodaldevoid:
Looks like Appveyor had a one-off DNS resolution issue. This looks good to me, but I would want to double check if there any any other source/header file lists that I forgot on my first pass.

mback2k:
> Looks like Appveyor had a one-off DNS resolution issue.

I just triggered it to retry the failed builds.

bagder:
If there are more mistakes, I'm sure we can deal with them as well!

PR #609, opened 2021-06-02
Add libssh2_userauth_keyboard_interactive2_ex

kkoenig:
Adds a new function to the public API that allows
  the caller to work around two issues with the previous
  version:
  1) Allow the callback to return EAGAIN so that it can
  be used in an asychronous context.  This now works
  in the same was as LIBSSH2_USERAUTH_PUBLICKEY_SIGN_FUNC
  2) Allow the caller to pass in their own state.  Again
  similar to libssh2_userauth_publickey, the caller
  can pass in their own abstract pointer.

  Extend the internal function userauth_keyboard_interactive
  to preserve the existing legacy behavior to avoid
  duplicating the core functionality

### DIFF
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -295,6 +295,13 @@ typedef struct _LIBSSH2_USERAUTH_KBDINT_RESPONSE
             const LIBSSH2_USERAUTH_KBDINT_PROMPT *prompts,              \
             LIBSSH2_USERAUTH_KBDINT_RESPONSE *responses, void **abstract)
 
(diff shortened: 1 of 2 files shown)


kkoenig:
Tried to do this in as few changes as possible without changing existing behavior.   Tried to model after userauth_publickey but happy to change the approach, please let me know!

willco007:
This looks OK to me, it just needs to rebased on the current source.

willco007:
Still waiting on a rebase with this.

Issue #614, opened 2021-08-13
Wrong description of libssh2_channel_flush_ex() return value in docs


ashegoulding:
**Describe the bug**
In [/docs/libssh2_channel_flush_ex.3](https://github.com/libssh2/libssh2/blob/a88a727c2a1840f979b34f12bcce3d55dcd7ea6e/docs/libssh2_channel_flush_ex.3), it says

> Return 0 on success or negative on failure.  It returns
> LIBSSH2_ERROR_EAGAIN when it would otherwise block. While
> LIBSSH2_ERROR_EAGAIN is a negative number, it isn't really a failure per se.

But in [/src/channel.h](https://github.com/libssh2/libssh2/blob/5627b82be66135dbda6423ae10bcad1c9ac86f9d/src/channel.h), it says

> /*
>  * _libssh2_channel_flush
>  *
>  * Flush data from one (or all) stream
>  * Returns number of bytes flushed, or negative on failure
>  */

The document should say "Returns number of bytes flushed", same as the comment.

willco007:
Fixed with #728.

PR #615, opened 2021-08-13
Fixing description of  libssh2_channel_flush_ex() in docs

doorsdown:
In #614 it was identified the docs do not accurately show how libssh2_channel_flush_ex() return value is set. I have updated the doc's to correctly show what the function is returining.


doorsdown:
Need to fix branch issue first.

Issue #536, opened 2020-11-17
Add support for RFC8332 (rsa-sha2-512 and rsa-sha2-256)


WSLUser:
**Describe the bug**
An application using this library is unable to connect to a couple hosts due to switching to ssh-rsa-sha2 and dropping ssh-rsa (which is considered insecure).
**To Reproduce**
Steps to reproduce the behavior.
Use an application that makes use of this library on a host that supports OpenSSH 7.2 or newer. The hosts I used are on Fedora 31, which uses OpenSSH 8.1.
A recipe or example code that reproduces the problem? A stack trace from a crash

**Expected behavior**
I expect the ability to harden my ssh configuration without losing ability to connect to hosts with an application due to 
 the underlying ssh library.
**Version (please complete the following information):**
 - OS: All OS that support OpenSSH 7.2 or newer
 - libssh2 version [e.g. 1.8.0]: Any

**Additional context**
See https://tools.ietf.org/html/rfc8332 as linked in https://www.openssh.com/specs.html

Also see https://github.com/jtesta/ssh-audit to see what should be supported from a hardening perspective.

WSLUser:
@willco007 Can you look into this? I verified both the java ssh library and the original libssh support this but couldn't do so for this library.

willco007:
@WSLUser My time is very limited at the moment, but I do have an outstanding issue on our own internal tracker about supporting these key types so there is hope. :) I would recommend ec keys for the time being.

WSLUser:
It's the same keys, just different algorithm. ssh-rsa uses sha-1, which is now considered in-secure. The RFC introduced a sha2 variant that works with existing keys. Problem with this library is that it doesn't permit using the newer algorithm with the existing key. Based on the implementation I saw in the java ssh library, it shouldn't be terribly difficult to add support for but is beyond my personal experience to implement.

willco007:
I have an implementation working on my private fork. We'll give a once over and land it soon-ish. Thanks for the report!

WSLUser:
@willco007 Now the holidays are finishing up, can you pick this back up with the once over? Love to see these get added before the next release of libssh2!

willco007:
This patch just went through my QA department and looks good to go. I will try and get it merged in next week time permitting.

WSLUser:
Friendly reminder. I see it hasn't been added yet but you're throwing up other PRs. Want to be sure this isn't missed.

willco007:
[...]

PR #619, opened 2021-08-30
Host Key RSA 256/512 support libssh2 #536

willco007:
Files: openssl.h, openssl.c, crypto.h, hostkey.c, HACKING-CRYPTO

Notes:
Added `rsa-sha2-256` and `rsa-sha2-512` as described in RFC 8332 to OpenSSL backend. To implement in other backends implementers need to define `LIBSSH2_RSA_SHA2` and implement `_libssh2_rsa_sha2_sign` and `_libssh2_rsa_sha2_verify`.

Credit:
Will Cosgrove


willco007:
Going to re-submit to trigger new CI

Issue #616, opened 2021-08-30
[1.10.0] openssh_fixture fails to build openssh server docker image


stac47:
Hello,

I am back with a unit test issue in the new release 1.10.0.
When the run the test (via CMake), all the tests are failing because the OpenSSH server image cannot be build:
```
Start  2: test_hostkey_hash
2/14 Test  #2: test_hostkey_hash ...............................................***Failed  118.13 sec
Error running command 'docker build -t libssh2/openssh_server openssh_server' (exit 36096): Sending build context to Docker daemon  15.36kB
Step 1/32 : FROM dockerhub.rnd.amadeus.net:5005/debian:jessie
---> 3aaeab7a4777
Step 2/32 : RUN apt-get update  && apt-get install -y openssh-server  && apt-get clean  && rm -rf /var/lib/apt/lists/*
---> Running in 34745de55fb5
Get:1 http://security.debian.org jessie/updates InRelease [44.9 kB]
Ign http://deb.debian.org jessie InRelease
[...]
```
The reason of this is the #580 which does not take care of the case where the output of the command exceeds the BUFSIZ (which is 8Kbytes on my platform). Normally, `pclose` should wait for the `docker build` to terminate but somehow it is not the case. On Docker daemon side I can see that it is notified that it should stop probably when the test process kills children:
```
 level=debug msg="Build cancelled, killing and removing container: d84bead224d9af58da4f1bfdcf85bcd200a31be4f5a6c718c5ebc6c663ec452"
level=debug msg="Sending kill signal 9 to container d84bead224d9af58da4f1bfdcf85bcd200a31be4f5a6c718c5ebc6c663ec452"
```

`run_command` is used only to issue a few docker commands. If we need an output, we take care of having only one line like in `docker inspect` (--format option), `docker run` (-d option), `docker-machine ip`. For the rest of the calls, we do not care about the output like in `docker build`.

With the current implementation, we get the beginning of the output while generally we are interested in the end (where the results or the error are written). The previous implementation only kept the last line which is enough in case we needed the output or not.

We have several solutions here:
* revert the commit and we document that `run_command` will only return the last line of an output. I would be in favour of this.
* we let the commit and document that the output is limited to `BUFSIZ` and to "fix" the build of the openssh image with something like `docker build --quiet`.
* we augment the size of `buf`

Regards,

willco007:
How large is the output on your system? Seems the simplest solution is increase `BUFSIZ`.

willco007:
[...]

PR #620, opened 2021-08-31
Fix openssh_server build not working (#616)

stac47:
Ok following #616, instead of increasing the buffer for an output we don't care (output of the docker build), let's just stop logging the whole image construction. We will still get the error message if any.

Don't hesitate to tell me if you prefer something else.


willco007:
Looks good to me. Odd the agent forward test failed on Windows; any ideas on that one?

stac47:
It is unlikely that it is linked to this PR. Is it possible for you to restart that job ?

PR #628, opened 2021-09-30
windows: fix clang and WinCNG warnings

vszakats:
Fix these categories of warning:

- in `wincng.c` disagreement in signed/unsigned char when passing around
  the passphrase string:
  `warning: pointer targets in passing argument [...] differ in signedness [-Wpointer-sign]`
  Fixed by using `const unsigned char *` in all static functions and
  applying/updating casts as necessary.

- in each use of `libssh2_*_init()` macros where the result is not used:
  `warning: value computed is not used [-Wunused-value]`
  Fixed by using `(void)` casts.

- `channel.c:1171:7: warning: 'rc' may be used uninitialized in this function [-Wmaybe-uninitialized]`
  Fixed by initializing this variable with `LIBSSH2_ERROR_CHANNEL_UNKNOWN`.
  While there I replaced a few 0 literals with `LIBSSH2_ERROR_NONE`.

- in `sftp.c`, several of these two warnings:
  `warning: 'data' may be used uninitialized in this function [-Wmaybe-uninitialized]`
  `warning: 'data_len' may be used uninitialized in this function [-Wmaybe-uninitialized]`
  Fixed by initializing these variables with NULL and 0 respectively.

- Also removed the exec attribute from `wincng.h`.

Notes:
- There are many pre-existing checksrc issues.
- The `sftp.c` and `channel.c` warnings may apply to other platforms as well.


vszakats:
The CI error seems unrelated (the same error can be seen a few weeks back.)

mback2k:
The AppVeyor tests require a Docker host which is provided by a GitHub Action. Sometimes that may be delayed too much.

vszakats:
@mback2k: Oh great, thanks for the info. Indeed after much waiting the tasks finished in green.

[ It leaves me wondering why reversing the order of OpenSSL and WinCNG tests made it work. Originally the errors were different, and they happened in actual tests, and in random jobs. ]

PR #629, opened 2021-10-01
Add support for a wolfSSL crypto backend.

haydenroche5:
It uses wolfSSL's OpenSSL compatibility layer, so rather than introduce new
wolfssl.h/c files, the new backend just reuses openssl.h/c. Additionally,
replace EVP_Cipher() calls with EVP_CipherUpdate(), since EVP_Cipher() is not
recommended.

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -40,7 +40,8 @@
 
 #include "libssh2_priv.h"
 
-#ifdef LIBSSH2_OPENSSL /* compile only if we build with openssl */
+/* compile only if we build with openssl or wolfSSL */
+#if defined(LIBSSH2_OPENSSL) || defined(LIBSSH2_WOLFSSL)
 
 #include <string.h>
 #include "misc.h"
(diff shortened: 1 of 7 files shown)


src/openssl.c
```diff
@@ -419,27 +420,20 @@ _libssh2_cipher_crypt(_libssh2_cipher_ctx * ctx,
 {
     unsigned char buf[EVP_MAX_BLOCK_LENGTH];
     int ret;
+    int outlen;
     (void) algo;
     (void) encrypt;
 
 #ifdef HAVE_OPAQUE_STRUCTS
-    ret = EVP_Cipher(*ctx, buf, block, blocksize);
+    ret = EVP_CipherUpdate(*ctx, buf, &outlen, block, blocksize);
```
haydenroche5, line 428: I _think_ simply switching to using `EVP_CipherUpdate` here clears up the return code complexity.

bagder (approved):

willco007:
This is great! I would like to see this backend added to the ci if possible, is that something you might be able to tackle?

haydenroche5:
> This is great! I would like to see this backend added to the ci if possible, is that something you might be able to tackle?

Sure, I can look into that. Got any examples/docs that I can use to get set up with your CI system?

willco007:
The file that needs updating is `.github/workflows/ci.yml`. Basically you'd need to add wolf to the matrix and then install it; you can model the code off the mbedtls backend which does something similar.

willco007:
I'm going to land this to make sure it gets into the next release. I'd still like to see it in CI when you have some time @haydenroche5. Thanks again!

PR #632, opened 2021-10-15
Support for direct-streamlocal@openssh.com UNIX socket connection

pa-sowa:
This is the #216 PR updated with CMakeLists.txt change that was blocking it.

This patch allow to use direct-streamlocal service from OpenSSH 6.7, that allows UNIX socket connections.

### DIFF
--- a/src/channel.c
+++ b/src/channel.c
@@ -434,6 +434,81 @@ libssh2_channel_direct_tcpip_ex(LIBSSH2_SESSION *session, const char *host,
 }
 
 /*
+ * libssh2_channel_direct_streamlocal_ex
+ *
+ * Tunnel TCP/IP connect through the SSH session to direct UNIX socket
+ */
+static LIBSSH2_CHANNEL *
+channel_direct_streamlocal(LIBSSH2_SESSION * session, const char *socket_path, const char *shost, int sport)
+{
+    LIBSSH2_CHANNEL *channel;
+    unsigned char *s;
+
+    if (session->direct_state == libssh2_NB_state_idle) {
+        session->direct_host_len = strlen(socket_path);
+        session->direct_shost_len = strlen(shost);
+        session->direct_message_len =
+            session->direct_host_len + session->direct_shost_len + 12;
+
+        _libssh2_debug(session, LIBSSH2_TRACE_CONN,
+                       "Requesting direct-streamlocal session to %s",
+                       socket_path);
+
+        s = session->direct_message =
+            LIBSSH2_ALLOC(session, session->direct_message_len);
+        if (!session->direct_message) {
+            _libssh2_error(session, LIBSSH2_ERROR_ALLOC,
+                           "Unable to allocate memory for direct-streamlocal connection");
+            return NULL;
+        }
+        _libssh2_store_str(&s, socket_path, session->direct_host_len);
+        _libssh2_store_str(&s, shost, session->direct_shost_len);
+        _libssh2_store_u32(&s, sport);
+    }
+
+    channel =
+        _libssh2_channel_open(session, "direct-streamlocal@openssh.com",
+                              sizeof("direct-streamlocal@openssh.com") - 1,
+                              LIBSSH2_CHANNEL_WINDOW_DEFAULT,
+                              LIBSSH2_CHANNEL_PACKET_DEFAULT,
+                              session->direct_message,
+                              session->direct_message_len);
+
+    if (!channel &&
+        libssh2_session_last_errno(session) == LIBSSH2_ERROR_EAGAIN) {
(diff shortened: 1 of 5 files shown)


kdart-brt:
FYI, I found it's possible to use the generic `libssh2_channel_open_ex` in your own code with a constructed message type of `direct-streamlocal@openssh.com` to achieve this.

willco007:
Would a sample app be more appropriate than a feature API if this can be accomplished with the existing API?

will:
> FYI, I found it's possible to use the generic `libssh2_channel_open_ex` in your own code with a constructed message type of `direct-streamlocal@openssh.com` to achieve this.

@kdart-brt could you share an example of this please? I imagine it's much like the code in the patch, but it'd be helpful to compare any small differences.

kdart-brt:
@will Well, I actually use this through a Python wrapper. I don't know how useful that will be to you, but here's the relevant part.
```python
    async def open_unix_tunnel(self, path: str, local_port: int) -> AsyncChannel:
        # We encode message body for generic channel open message.
        msgtype = b"direct-streamlocal@openssh.com"
        host = b"localhost"
        path = path.encode("utf8")
        body = len(path).to_bytes(4, byteorder="big") + path
        body += len(host).to_bytes(4, byteorder="big") + host
        body += local_port.to_bytes(4, byteorder="big")
        # open the generic channel with our carefully crafted message body.
        chan = self._session.open_channel(msgtype, body)
        while chan == LIBSSH2_ERROR_EAGAIN:
            await streams.trap_read_wait(self._sock)
            chan = self._session.open_channel(msgtype, body)
        assert isinstance(chan, Channel)
        return AsyncChannel(chan, self._sock)

```

will:
@kdart-brt hah, I was just typing up a message to tell you to never mind. I appreciate it though!

I was misunderstanding `libssh2_channel_forward_listen_ex` and though that it would actually open up a local port, and that this patch was needed to open a local socket. I now know that opening a local port or socket is your responsibility, and actually for my use case I don’t even need to open any local port or socket, I can give this channel straight to the thing I need it for.

Just for the record if other people happens to search, I was trying to tunnel a postgres connection through ssh with crystal, and it ends up looking something like this:

```crystal
socket = TCPSocket.new(host, 22, connect_timeout: 1)
ssh = SSH2::Session.new(socket)
ssh.login_with_data("username", priv_key, pub_key)
ssh_ch = sssh.direct_tcpip("localhost", 5432, "localhost", 5432)
conninfo = PQ::ConnInfo.new(user:"user", password:"secret", database:"postgres")
module PQ
  class Connection
    def initialize(@conninfo : ConnInfo, @soc : SSH2::Channel)
      @mutex = Mutex.new
      @server_parameters = Hash(String, String).new
      @established = false
      @notice_handler = Proc(Notice, Void).new { }
      @notification_handler = Proc(Notification, Void).new { }
    end
  end
end


db = PQ::Connection.new(conninfo, ssh_ch)
db.connect
q = PQ::ExtendedQuery.new(db, "select 'hahaha yes'", [] of String)
q.get_data {|row| p row.map{|c| String.new(c) if c }  }
```

The postgres driver needs that monkey patch right now in order to give it the socket that comes from libss2. However, I'm also the author of the crystal postgres driver, so I'll probably make it support taking an already established socket instead of always making a new one.

jakob:
I think this PR is a valuable addition!

It is already possible to open channels to a UNIX socket with libssh2_channel_open_ex, but it requires reading about internals of the SSH protocol to construct the channel open message.

libssh2 already provides a convenience method for opening  **TCP/IP channels**:  libssh2_channel_direct_tcpip_ex()

So it would definitely make sense to also add a convenience method for opening **Unix socket channels**: libssh2_channel_direct_streamlocal_ex()

I think it absolutely makes sense to offer both of these methods. If `libssh2_channel_direct_streamlocal_ex()` is not needed, why is `libssh2_channel_direct_tcpip_ex()` necessary?

src/channel.c
```diff
@@ -433,6 +433,84 @@ libssh2_channel_direct_tcpip_ex(LIBSSH2_SESSION *session, const char *host,
...
+        s = session->direct_message =
+            LIBSSH2_ALLOC(session, session->direct_message_len);
+        if(!session->direct_message) {
+            _libssh2_error(session, LIBSSH2_ERROR_ALLOC,
+                "Unable to allocate memory for direct-streamlocal connection");
+            return NULL;
+        }
+        _libssh2_store_str(&s, socket_path, session->direct_host_len);
+        _libssh2_store_str(&s, shost, session->direct_shost_len);
+        _libssh2_store_u32(&s, sport);
```
jakob, line 467: How did you find out that these parameters are source host / source port? The [PROTOCOL file](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL) just lists these fields as `reserved`.

PR #635, opened 2021-10-27
ci: Add CIFuzz integration

DavidKorczynski:
Add CIFuzz Github action to build and run fuzzers on pull requests.

This will run the fuzzers for 600 seconds (split between all the fuzzers) in the CI when a pull request is made, and is used to catch bugs early and also catch regressions. CIFuzz is a service run in parallel with OSS-Fuzz, which libssh2 is already integrated into.

Documentation for CIFuzz can be found here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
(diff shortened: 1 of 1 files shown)


WSLUser:
OneFuzz from Microsoft would be a better option than Google's.

bagder:
@WSLUser you're most welcome to provide integration into other services

bagder:
Thanks!

PR #637, opened 2021-11-02
Clean up a variable redefinition warning

tsengjun:
Clean up a variable redefinition warning to satisfy the compiler.


tsengjun:
Trigger the checker by reopening it.

tsengjun:
Reopen to trigger the checkers

tsengjun:
There are many of these kinds of issues in the code. They do not impact functionality; so far, everything has gone well. I think we can keep it as is. Close this PR as there is no side effect - only some compiler warnings.

PR #643, opened 2021-11-22
Runtime engine detection

BastienDurel:
Proposition of an API to let the programmer know which crypto engine is used in libssh2

### DIFF
--- a/src/version.c
+++ b/src/version.c
@@ -52,3 +52,19 @@ const char *libssh2_version(int req_version_num)
         return LIBSSH2_VERSION;
     return NULL; /* this is not a suitable library! */
 }
+
+LIBSSH2_API
+libssh2_crypto_engine_t libssh2_crypto_engine()
+{
+#if defined LIBSSH2_OPENSSL
+    return libssh2_openssl;
+#elif defined LIBSSH2_LIBGCRYPT
+    return libssh2_gcrypt;
+#elif defined LIBSSH2_MBEDTLS
+    return libssh2_mbedtls;
+#elif defined LIBSSH2_WINCNG
+    return libssh2_wincng;
+#else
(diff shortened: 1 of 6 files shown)


src/version.c
```diff
@@ -52,3 +52,19 @@ const char *libssh2_version(int req_version_num)
...
+#if defined LIBSSH2_OPENSSL
+    return libssh2_openssl;
+#elif defined LIBSSH2_LIBGCRYPT
+    return libssh2_gcrypt;
+#elif defined LIBSSH2_MBEDTLS
+    return libssh2_mbedtls;
+#elif defined LIBSSH2_WINCNG
+    return libssh2_wincng;
+#else
+    #error "Unhandled crypto engine"
```
willco007, line 68: You can build libssh2 with no crypto backend. It should return a `none` option instead of erroring here.

willco007 (changes requested):
Looks good, just needs to return a none option with no backend.

BastienDurel:
Hello, is this good now ?

willco007 (changes requested):
The next version will be 1.11, changing the version strings here would be great. My only other thought is moving the `return libssh2_no_crypto` outside of the #if/elif block, but that's my personal preference more than anything. :)

BastienDurel:
I prefer keeping `return libssh2_no_crypto` inside the `#else` to prevent unreachable code. Some compiler may complain about it.

BastienDurel:
Hello,

any news ?

willco007 (changes requested):
libssh2_crypto_engine.3 file needs to be added to the docs/CmakeLists.txt file as well as the Makefile.am.

BastienDurel:
Done

Issue #659, opened 2022-01-12
ssh-agent libssh2_agent_userauth error, works on 1.10.0


ianhattendorf:
**Describe the bug**
Hi there,

We're excited about the `rsa-sha2-256/512` support in libssh2, however we encountered a bug with ssh-agent support reproduced with the examples.

Everything works fine when using keys from disk, however when trying to auth to a server using `libssh2_agent_userauth` we get `LIBSSH2_ERROR_PUBLICKEY_UNVERIFIED`. The same exact code works with 1.10.0 (when accessing a server that accepts `ssh-rsa`).

See below for output from a few different scenarios, along with a dockerfile and patch used for testing.

Potentially related: #536 and #626

**To Reproduce**
`./example/example-ssh2 <server-ip> git`: success
`./example/example-ssh2_agent <server-ip> git`: error

See test patch below used to provide output, but reproducible without any patches.

**Expected behavior**
`libssh2_agent_userauth` should auth successfully.

**Version (please complete the following information):**
 - OS: Fedora 35
 - libssh2 version: master (13ad7b2f5cd67e0dc843098ce19ce8b208368c29)

**Additional context**
**1.10.0 example-ssh2 PubkeyAcceptedKeyTypes=ssh-ed25519,rsa-sha2-256,ssh-rsa (Success expected):**

```
Connecting to git@172.17.0.2. pass: password
[libssh2] 0.572899 Key Ex: Sent KEX: curve25519-sha256,curve25519-sha256@libssh.org,ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521,diffie-hellman-group-exchange-sha256,diffie-hellman-group16-sha512,diffie-hellman-group18-sha512,diffie-hellman-group14-sha256,diffie-hellman-group14-sha1,diffie-hellman-group1-sha1,diffie-hellman-group-exchange-sha1
[libssh2] 0.572909 Key Ex: Sent HOSTKEY: ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ecdsa-sha2-nistp256-cert-v01@openssh.com,ecdsa-sha2-nistp384-cert-v01@openssh.com,ecdsa-sha2-nistp521-cert-v01@openssh.com,ssh-ed25519,ssh-rsa,ssh-dss
[...]

PR #662, opened 2022-01-14
Legacy Agent support for rsa2 key upgrading/downgrading #659

willco007:
Files: libssh2.h, agent.c, userauth.c

Notes:
Part 2 of the fix for #659. This adds rsa key downgrading for agents that don't support sha2 upgrading. It also adds better trace output for debugging/logging around key upgrading.

Credit:
Will Cosgrove (signed off by Michael Buckley)


ianhattendorf:
This looks good to me, our test suite is now passing (our Windows pageant version currently doesn't support SHA2).
Before: https://github.com/nodegit/nodegit/runs/4814700972?check_suite_focus=true
After: https://github.com/nodegit/nodegit/actions/runs/1698528629

Thanks for following up with this.

PR #669, opened 2022-02-03
Fix heap buffer overflow in _libssh2_key_sign_algorithm

tihmstar:
When allocating `session->server_sign_algorithms` which is a `char*` is is important to also allocate space for the string-terminating null byte at the end and make sure the string is actually null terminated.

Without this fix, the `strchr()` call inside the `_libssh2_key_sign_algorithm` (line 1219) function will try to parse the string and go out of buffer on the last invocation.

EDIT:
Dunno why i didn't directly just link to the bug lol.
Heap buffer overflow (without this fix) happens here https://github.com/libssh2/libssh2/blob/de7a74aff24c47b2f2e9815f0a98598195d602e4/src/userauth.c#L1212-L1219

### DIFF
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -642,7 +642,6 @@ struct _LIBSSH2_SESSION
 
     /* public key algorithms accepted as comma separated list */
     char *server_sign_algorithms;
-    size_t server_sign_algorithms_len;
 
     /* key signing algorithm preferences -- NULL yields server order */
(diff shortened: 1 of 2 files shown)


willco007:
Thanks for the fix! With this change, we don't need to keep `server_sign_algorithms_len` around. Would you mind updating your PR to remove that ivar?

willco007:
I went ahead and updated your PR, thanks again @tihmstar.

PR #674, opened 2022-02-17
feat: userauth.c: Calculating internally the length of the given user name in libssh2_userauth_list()

anaktas:
I've tried to simplify the usage of the libssh2_userauth_list() function, by moving the calculation of the username length internally instead of passing it as function argument.

### DIFF
--- a/docs/libssh2_userauth_list_ex.3
+++ b/docs/libssh2_userauth_list_ex.3
@@ -1,0 +1,36 @@
+.TH libssh2_userauth_list_ex 3 "21 Feb 2022" "libssh2 1.10.0" "libssh2 manual"
+.SH NAME
+libssh2_userauth_list_ex - list supported authentication methods
+.SH SYNOPSIS
+.nf
+#include <libssh2.h>
+
+char *
+libssh2_userauth_list_ex(LIBSSH2_SESSION *session, const char *username);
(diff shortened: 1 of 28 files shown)


anaktas:
Making a second thought about this, this could probably cause some serious compatibility issues. i will try to make either a macro, or a similar wrapper function.

willco007:
This is failing the style check and would need to be documented. What is the up-side to this API other than not calling strlen() yourself?

anaktas:
Hi @willco007 

The only upside of this API is just the benefit of usability. While using first time the libssh2_userauth_list(), I was confused about the username_len parameter (aka not reading the documentation and not interpreting the _len with length...) and instead of using strlen(username) I used sizeof(username), which leaded me to a problematic behavior after that. Since this is a necessary parameter for listing the available authentication parameters, I thought that it would be better to hide it from the API and calculate the username length internally.

willco007:
Hiding the `username_len` parameter has a potential downside if the client data is not null terminated this API will crash. In this case being explicit seems to have no tangible draw back. While I appreciate any and all PRs, I'm just not sure this has enough value to include in the library.

anaktas:
Thanks for reviewing this @willco007 !

I understand your considerations. Please feel free to close this PR without merging it.

PR #691, opened 2022-04-18
tests: fix ssh2.sh for OpenSSH 8.8 and newer

vangyzen:
OpenSSH on FreeBSD does not accept ssh-rsa keys by default.
Specify the key type for host and user keys so the test passes.
FreeBSD has a separate PubkeyAcceptedAlgorithms config option
for user keys.  The test still passes on openSUSE 15.3.


willco007:
All versions of OpenSSH 8.8 or greater disable sha1 rsa keys. Adding back the rsa key option should probably look at the version of OpenSSH running and not the OS for greater flexibility when running the tests.

vangyzen:
Adding the ssh-rsa key type unconditionally keeps the test simple.

Now I see that `PubkeyAcceptedAlgorithms` was simply renamed from `PubkeyAcceptedKeyTypes`, and the old name is retained as an alias, so this can be simplified quite a bit.

willco007:
Hum, the `configure` tests failed on this commit, would you mind investigating?

vszakats:
Would it be feasible to upgrade to modern (SHA2) algo instead?

vszakats:
Would this also make sense for the `sshd_config` inside `Dockerfile`?

We use Debian bullseye there, with OpenSSH 8.4, but the next release will have 9.2 or newer:
https://packages.debian.org/bullseye/openssh-server
https://packages.debian.org/testing/openssh-server

```diff
diff --git a/tests/openssh_server/Dockerfile b/tests/openssh_server/Dockerfile
index 39921c2c..3aed7cb7 100644
--- a/tests/openssh_server/Dockerfile
+++ b/tests/openssh_server/Dockerfile
@@ -80,7 +80,8 @@ RUN echo 'libssh2:my test password' | chpasswd
 
 RUN sed -i 's/ChallengeResponseAuthentication no/ChallengeResponseAuthentication yes/' /etc/ssh/sshd_config
 RUN echo "TrustedUserCAKeys /etc/ssh/ca_main.pub" >> /etc/ssh/sshd_config
-RUN echo "PubkeyAcceptedKeyTypes=+ssh-dss" >> /etc/ssh/sshd_config
+RUN echo "HostKeyAlgorithms=+ssh-rsa" >> /etc/ssh/sshd_config
+RUN echo "PubkeyAcceptedKeyTypes=+ssh-rsa,ssh-dss" >> /etc/ssh/sshd_config
 RUN echo "MACs +hmac-sha1,hmac-sha1-96,hmac-sha2-256,hmac-sha2-512,hmac-md5,hmac-md5-96,umac-64@openssh.com,umac-128@openssh.com,hmac-sha1-etm@openssh.com,hmac-sha1-96-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com,hmac-md5-etm@openssh.com,hmac-md5-96-etm@openssh.com,umac-64-etm@openssh.com,umac-128-etm@openssh.com" >> /etc/ssh/sshd_config
 RUN echo "Ciphers +3des-cbc,aes128-cbc,aes192-cbc,aes256-cbc,rijndael-cbc@lysator.liu.se,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com,chacha20-poly1305@openssh.com" >> /etc/ssh/sshd_config
```

PR #698, opened 2022-05-13
Add support for sk-ecdsa-sha2-nistp256@openssh.com and…

MichaelBuckley:
… sk-ssh-ed25519@openssh.com key exchange. These are the FIDO keys. There are also certificate variants that are not covered as part of this pull request. I hope to add them in a future PR. They are described in https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.u2f

This change adds a new public function, `libssh2_userauth_publickey_sk`. This function takes a callback function that's responsible for performing fido signing and  filling out a new `LIBSSH2_SK_SIG_INFO` struct. The flags field passed into this callback uses two new #defines from libssh2.h:

- `LIBSSH2_SK_PRESENCE_REQUIRED` indicates whether the user must touch the security key to authenticate.
- `LIBSSH2_SK_VERIFICATION_REQUIRED` indicates whether the user must supply a PIN to the security key. The PIN is not passed to the callback function. It is up to the callback function to prompt the user for the PIN, or the client can pass it as part of the `abstract` parameter. This was done to prevent the need to parse the key multiple times to determine if a PIN is required.

I am posting this PR to solicit feedback on this APi, as well as to ask for more eyes on the implementation to check for mistakes.

The only implementation provided is for the openssl backend, as that's the backend I'm using. Stubs are provided for the other backends.

I'm not sure how to best update documentation for the new public function. There a bunch of man pages in the docs directory, but they appear to be outdated, and some of them are incomplete. I'm not sure if they're actively maintained.

### DIFF
--- a/docs/libssh2_sign_sk.3
+++ b/docs/libssh2_sign_sk.3
@@ -1,0 +1,85 @@
+.TH libssh2_sign_sk 3 "1 Jun 2022" "libssh2 1.10.0" "libssh2 manual"
+.SH NAME
+libssh2_sign_sk - Create a signature from a FIDO2 authenticator.
+.SH SYNOPSIS
+#include <libssh2.h>
+.nf
+int libssh2_sign_sk(LIBSSH2_SESSION *session,
+                    unsigned char **sig,
+                    size_t *sig_len,
+                    const unsigned char *data,
+                    size_t data_len,
+                    void **abstract);
+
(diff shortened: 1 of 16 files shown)


MichaelBuckley:
I'll fix the CI errors and push fixes tomorrow.

MichaelBuckley:
All CI tests now passing.

PR #682, opened 2022-03-03
Opaque structs in LibreSSL 3.5

vishwin:
This fixes the build with LibreSSL 3.5. See [release notes](https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.5.0-relnotes.txt) for the large amount of API additions.

PR #700, opened 2022-05-19
openssl: add support for LibreSSL 3.5.x

vszakats:
LibreSSL 3.5.0 made more structures opaque, so let's enable existing
support for that when building against these LibreSSL versions.

Ref: https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.5.0-relnotes.txt

### DIFF
--- a/src/openssl.h
+++ b/src/openssl.h
@@ -95,7 +95,8 @@
 #include <openssl/rand.h>
 
 #if (OPENSSL_VERSION_NUMBER >= 0x10100000L && \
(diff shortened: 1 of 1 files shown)


vszakats:
Oops, just found an existing PR for this. Though this implementation remains within line length limits and is a little bit simpler (but should work just as well.)

Ref: https://github.com/libssh2/libssh2/pull/682

willco007 (approved):

PR #732, opened 2022-08-04
Meson build scripts added

### LINKED ISSUE #725: add meson build system
everything is in the title. Advantages : 
 * simpler syntax than cmake
 * a bit faster than cmake
 * it just works, no hacks like one can find in cmake
 * cross-compilation is easier to set up
 * options are clearly available in a separate file, no need to find them in all the CMakeLists.txt or in cmake/
 * you can you wrap feature to get dependencies like glig2 or gstreamer projects are doing

ilsme:
Fixes #725

### DIFF
--- a/example/meson.build
+++ b/example/meson.build
@@ -1,0 +1,67 @@
+examples_c = [
+    'direct_tcpip.c',
+    'scp.c',
+    'scp_nonblock.c',
+    'scp_write.c',
+    'scp_write_nonblock.c',
+    'sftp.c',
+    'sftp_RW_nonblock.c',
+    'sftp_append.c',
+    'sftp_mkdir.c',
+    'sftp_mkdir_nonblock.c',
+    'sftp_nonblock.c',
+    'sftp_write.c',
+    'sftp_write_nonblock.c',
+    'sftp_write_sliding.c',
+    'sftpdir.c',
+    'sftpdir_nonblock.c',
+    'ssh2.c',
+    'ssh2_agent.c',
+    'ssh2_echo.c',
+    'ssh2_exec.c',
+    'subsystem_netconf.c',
+    'tcpip-forward.c'
+]
+
+example_libssh2_config = configuration_data()
+
+check_example_headers = [
+    'unistd.h',
+    'inttypes.h',
+    'stdlib.h',
+    'sys/select.h',
+    'sys/socket.h',
+    'netinet/in.h',
+    'arpa/inet.h',
+    'sys/time.h',
+    'winsock2.h',
+]
+
+foreach h : check_example_headers
+    example_libssh2_config.set('HAVE_@0@'.format(h.underscorify().to_upper()), cc.has_header(h))
+endforeach
+
+foreach f : ['strcasecmp', '_stricmp', 'snprintf', '_snprintf']
+    example_libssh2_config.set('HAVE_@0@'.format(f.underscorify().to_upper()), cc.has_function(f))
+endforeach
+
+example_libssh2_config_h = configure_file(
+    input: 'libssh2_config_cmake.h.in',
+    output: 'libssh2_config.h',
+    configuration: example_libssh2_config,
+    format: 'cmake',
+)
+
+x11_dep = dependency('x11', required: false)
+if x11_dep.found()
+    examples_c += 'x11.c'
+endif
+
+foreach example : examples_c
+    executable(
+        example.replace('.c', ''),
+        [ example, example_libssh2_config_h ],
+        include_directories: include_directories('.'),
+        dependencies: libssh2_dep
+    )
+endforeach
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,155 @@
+project('libssh2', 'c',
+  version : '1.10.1-dev',
+  license : 'BSD-3',
+  meson_version : '>=0.58.0')
+
+cc = meson.get_compiler('c')
+pkgconfig = import('pkgconfig')
+
+libssh2_sources = [
+    'src/misc.c',
+    'src/scp.c',
+    'src/crypt.c',
+    'src/hostkey.c',
+    'src/version.c',
+    'src/wincng.c',
+    'src/blowfish.c',
+    'src/mac.c',
+    'src/os400qc3.c',
+    'src/transport.c',
+    'src/agent.c',
+    'src/pem.c',
+    'src/kex.c',
+    'src/sftp.c',
+    'src/knownhost.c',
+    'src/publickey.c',
+    'src/channel.c',
+    'src/global.c',
+    'src/packet.c',
+    'src/bcrypt_pbkdf.c',
+    'src/userauth.c',
+    'src/keepalive.c',
+    'src/comp.c',
+    'src/session.c',
+    'src/userauth_kbd_packet.c'
+]
+
+libssh2_deps = []
+extra_flags = []
+
+if get_option('crypto_backend') == 'openssl'
+    libssh2_sources += ['src/openssl.c']
+    libssh2_deps += dependency('openssl', version : '>= 1.0.0')
+    extra_flags += ['-DLIBSSH2_OPENSSL']
+elif get_option('crypto_backend') == 'mbedtls'
+    libssh2_sources += ['src/mbedtls.c']
+
+    mbedtls_dep = dependency('mbedtls', required: false)
+    if mbedtls_dep.found()
+        libssh2_deps += mbedtls_dep
+    else
+        libmbedtls = cc.find_library('mbedtls', has_headers: ['mbedtls/version.h'])
+        libmbedx509 = cc.find_library('mbedx509')
+        libmbedcrypto = cc.find_library('mbedcrypto')
+        libssh2_deps += [libmbedtls, libmbedx509, libmbedcrypto]
+    endif
+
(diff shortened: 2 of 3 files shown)


ilsme:
#725

vtorri:
very nice!

vtorri:
any review of this PR ?

vtorri:
@ilsme maybe there should be more options: look at all the AC_ARG_* commands in configure.ac

vtorri:
add wolfSSL and WinCNG crypto backend

ilsme:
@vtorri you're welcome to add it by yourself

vtorri:
@ilsme i'll be happy to provide a PR if your PR will be applied.

ilsme:
@vtorri my PR was applied to wrapdb, you can add your PR there

neheb:
@vtorri WrapDB has WinCNG, openssl, gnutls, and mbedtls but no wolfssl.

ilsme:
@neheb wdym "have" mbedtls? I though someone added it as a separate wrap but I don't see it.

vszakats:
After making an updated version of this PR locally, the build fails in an odd way (on macOS), see below. Also there have been a bunch of improvements with existing CMake and autotools builds, so those might work better now than earlier. They are also fairly well tested. I'm personally not very keen on adding more build tools, even though Meson looks much leaner than the classic ones.

@libssh2/dev-team What do you think?

```
[1/24] Compiling C object libssh2.1.dylib.p/src_global.c.o
FAILED: libssh2.1.dylib.p/src_global.c.o 
cc -Ilibssh2.1.dylib.p -I. -I.. -I../include -I/usr/local/include/ -fcolor-diagnostics -Wall -Winvalid-pch -O0 -g -DLIBSSH2_MBEDTLS -MD -MQ libssh2.1.dylib.p/src_global.c.o -MF libssh2.1.dylib.p/src_global.c.o.d -o libssh2.1.dylib.p/src_global.c.o -c ../src/global.c
In file included from ../src/global.c:39:
../src/libssh2_priv.h:966:32: warning: declaration of 'struct iovec' will not be visible outside of this function [-Wvisibility]
                  const struct iovec datavec[], void **abstract);
                               ^
../src/libssh2_priv.h:966:45: error: array has incomplete element type 'const struct iovec'
                  const struct iovec datavec[], void **abstract);
                                            ^
../src/libssh2_priv.h:966:32: note: forward declaration of 'struct iovec'
                  const struct iovec datavec[], void **abstract);
                               ^
1 warning and 1 error generated.
```

dfandrich:
I think we already support one build system too many.

neheb:
@vszakats wrapdb has a newer version that works. This PR needs updating.

bagder:
There are many cool and good build systems out there but I don't think libssh2 needs to add support for more than it already does. It complicates the lives of the maintainers for very little gain.

vtorri:
@bagder about gain, get a computer with Windows, install MSYS2 with mingw-w64 in it. Compile curl with autotools, then with cmake. And you will see that there is a gain... meson is a bit faster than cmake. And muon (a C99 port of meson) is a quite faster than meson.

bagder:
I am aware. I still think it is a net *cost* added to the project and its maintainers, only to give a few people slightly faster builds the few times they build it.

vszakats:
@vtorri: speaking of speed, CMake should be faster now on Windows, due the efforts made to eliminate as much detection steps as possible. We also still have `Makefile.mk` (formerly known as `win32/GNUmakefile`), which doesn't have a detection phase at all on Windows, making it the fastest by a good margin.

jas4711:
+1 to focus on autoconf which is the most flexible/portable.  I agree multiple build systems costs more than it gives back to the project.

neheb:
The issue with autoconf is its buggyness and slow compilation speed. I'd at least get rid of it and keep only CMake.

vszakats:
CMake provides MSVC support (autotools doesn't), which seems to justify supporting it next to autotools. After recent updates, both should work equally well for most cases and have fairly good feature parity. Both are here to stay I believe. Also, all build tools now share common input source lists, so there is no need anymore to maintain parallel ones. (Meson would re-add a second source list, and it doesn't support reusing `Makefile.inc` as far as I could tell.)

(Disclaimer: not a fan of either of autotools/CMake and also not an MSVC user.)

vszakats:
Thanks for the input to everyone. The conclusion is that we are not planning to add a new build system. Having said that it was interesing to hear about Meson, Muon and ninja a giving them a shot.

Also opened PR #1025 deleting one more detection step to make CMake init a bit faster.

vszakats:
In my experience for projects of sizes comparable to libssh2 (= smaller ones), CMake and autotools are slow on Windows because of their often inflated detection phase. On Windows, file ops are very slow and detections are effectively compiling a whole .exe each. Since it's easy to add more detections, projects tend to accumulate a lot of them, making them slow. Also Windows is a fairly static environment, so most of these detections have predictable results so mostly avoidable with little downside.

The compilation/linking phase is bound to the toolchain's speed, and doesn't depend much on the build tool. (again for smaller projects like libssh2). This phase could be sped up by invoking the compiler a single time for all the sources, but interestingly none of the build tools do that (not even Meson), even though all compilers support it.

One more thing we can do to speed up CMake for libssh2 on Windows, is to avoid detections completely and rely on the recently built-in and streamlined internal Win32 configuration. I.e. skip all detections, and avoid defining `HAVE_CONFIG_H`, then tackle fallouts. Note that I've done that for quite a bit of detections already.

neheb:
Oh you mean __has_include? That’s a C++17 feature AFAIK.

vszakats:
No, I meant to use this with CMake (though on a second thought, I'm not sure it's a good idea, as it would introduce surprising/unexpected behaviour): https://github.com/libssh2/libssh2/blob/272a17ce266b4428c95c5337925aababc4acfa53/src/libssh2_setup.h#L22-L44

With the patch linked above we'll be down to [6 detections](https://ci.appveyor.com/project/libssh2org/libssh2/builds/46958280) on Windows with MSVC, 8 with clang/mingw64, 11 with gcc/mingw64. This is down from 32 for MSVC [two months ago](https://ci.appveyor.com/project/libssh2org/libssh2/builds/46958280). All this while adding picky warnings, which used to require many dozens of extra detections, but that also got almost completely eliminated (to 3 with gcc).

Compilation should also be faster after recent changes, e.g. on Windows `WINDOWS_LEAN_AND_MEAN` is universally enabled, with more similar options added and eliminated most duplicate or unnecessary includes, amongst them many Windows ones. We also have less compiler invocations due to the side-effect of compiling some sources as one unit (e.g. `blowfish.c` or `agent_win.c`). I've also added `TrackFileAccess=false` eliminating some VS wrapper we don't need with MSVC (perhaps this should be limited to our CI, I couldn't find much background info on its purpose). Enabling parallel builds can also reduce build times quite noticeably (our CI now uses `make -j3` and `cmake --build ... --parallel 3`).

Unless we expect `poll()` support to appear on Windows anytime soon, we might exclude that one from CMake detection as well.

Have you tried compiling a recent master branch?

bagder:
> I’m in favor of getting rid of autoconf.

I would be strongly against that. Call me conservative and old-fashioned but I have played this game (Open source) for a long time, and my experience have so far been that autotools is well-known, survives and is pretty much the only system that generates build files that can run on systems without them first having to install a new build toolset. This often makes autotools the only system that that still works on legacy things from the past. And I value stability, portability and accessibility before a minor increase in build-time. This project is tiny and builds fast anyway. IMHO.

neheb:
And in my experience with OpenWrt autotools is slow and buggy (read: compilation failure).

I've spent a considerable amount of time migrating as many packages as possible from autotools to CMake/meson and at this point, the whole compilation process is faster and less likely to break.

bagder:
> And in my experience with OpenWrt autotools is slow and buggy (read: compilation failure).

Slow, yes. Buggy: not in general. That's mostly not because of autotools, but because of wrong/bad use of it.

You win *nothing* by removing support for autotools in libssh2, but you lose something.

vszakats:
Predefining certain things is perfectly safe on Windows, as the environment is fairly static. The hand-crafted config was (as `win32/libssh2_config.h`) and is used with build tools without detection logic, like `Makefile.mk`, `NMakefile` and anyone building libssh2 via Visual Studio's own build facilities, aka anyone without a `config.h`.

We have ~40 GitHub Actions, but there is no job there that uses CMake on Windows. Maybe I should add one for clang and gcc. Or what do you mean by regular GitHub Actions?

neheb:
Looking again, it seems it's currently done: https://github.com/libssh2/libssh2/blob/master/.github/workflows/ci.yml#L164

clang-cl is not tested though. Might get some useful warnings if compiling with /Wanalyzer

vszakats:
That's testing CMake with MSVC. What's missing is CMake + mingw-w64 + clang/gcc.

vszakats:
After some updates done yesterday, CMake detections are down to zero for MSVC and MinGW + clang, and down to 3 for MinGW + gcc (that can be zeroes by disabling picky warnings). Configuration takes just a few seconds now.

vszakats:
Also added "unity" builds for CMake: #1034. This should boost build times to a few seconds per lib flavour. E.g. on macOS it's below 1 second. For Windows clang cross-builds it's below 5 seconds (after 1.5s configuration), on old hardware.

PR #734, opened 2022-08-11
Fix compile for Windows UWP/WinRT platform.

dmitrykos:
These changes are required to compile for Windows UWP/WinRT because  it does not support full set of WIN32 API.

### DIFF
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -50,6 +50,11 @@
 #undef WIN32_LEAN_AND_MEAN
 #endif
 
+/* Windows UWP/WinRT platform. */
+#if defined(WINAPI_PARTITION_APP) && WINAPI_PARTITION_APP
+#define WINAPP
+#endif
+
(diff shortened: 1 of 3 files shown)


src/libssh2_priv.h
```diff
@@ -50,6 +50,11 @@
 #undef WIN32_LEAN_AND_MEAN
 #endif
 
+/* Windows UWP/WinRT platform. */
+#if defined(WINAPI_PARTITION_APP) && WINAPI_PARTITION_APP
+#define WINAPP
```
mback2k, line 55: If this is a custom define, we should probably prefix it with `LIBSSH2_`

dmitrykos: @mback2k, sorry noticed your reply only now when receive a e-mail notification about PR closure. It seems corrections were incorporated anyway here https://github.com/libssh2/libssh2/blob/730c606b64f86b5899183655e8f07578a54b8f9e/src/agent_win.c#L58

PR #742, opened 2022-09-12
Add CI for mingw-w64 via msys2

jeroen:


mback2k:
Nice one! What happens if you add `--enable-werror` to the `./configure`?

jeroen:
> Nice one! What happens if you add --enable-werror to the ./configure?

It errors due to printf formatting warnings

mback2k (approved):

PR #743, opened 2022-09-12
Add test to connect to git server

jeroen:


willco007:
What's the upside of this test verses a 'standard' ssh connection test?

PR #745, opened 2022-09-16
Add missing winsock2 link library

robUx4:
It's needed for the `WSAGetLastError()` call in misc.c for example.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -215,6 +215,8 @@ set(SOURCES
 
 if(WIN32 AND BUILD_SHARED_LIBS)
   list(APPEND SOURCES ${PROJECT_SOURCE_DIR}/win32/libssh2.rc)
+  list(APPEND LIBRARIES ws2_32)
+  list(APPEND PC_LIBS -lws2_32)
 endif()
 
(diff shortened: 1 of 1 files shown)


willco007:
Hum, how does this build and pass the Windows tests if this is required?

willco007:
Looking at the CMake build logs, it does find winsock2 and builds against it; but it's not linking it (?).  Seems like this would have come up before now.

vszakats:
Please see #811 for a fix proposal.

robUx4:
> Replacing `cmake/SocketLibraries.cmake` content with this, should fix it:
> 
> ```cmake
> macro(append_needed_socket_libraries LIBRARIES_LIST)
>   if(WIN32)
>     set(HAVE_SELECT 1)
>     list(APPEND ${LIBRARIES_LIST} ws2_32)
>   endif()
> endmacro()
> ```
> 
> Related PR: #811

It does not add `-lws2_32` to the `libssh2.pc` file.

vszakats:
Does that fix it?: https://github.com/vszakats/libssh2/commit/972d6220b309d79f713672e24b43d86fe764f869

PR #746, opened 2022-09-19
sftp: Prevent files from being skipped if the output buffer is too small

### LINKED ISSUE #714: libssh2_sftp_opendir does not return all files if LIBSSH2_ERROR_BUFFER_TOO_SMALL returned
**Describe the bug**

When reading a directory using libssh2_sftp_readdir_ex, LIBSSH2_ERROR_BUFFER_TOO_SMALL is returned if the buffer is too small to contain a returned directory entry. An application would then allocate a larger buffer and try again, however libssh2 then returns one less file than the directory actually contains.

**To Reproduce**

A simple reproduction can be found here: https://gist.github.com/harmic/459af38e7c0717c22e59a49c88109c08

Example - create a test directory 'tst' with a couple of files in it:

```
mkdir tst
touch tst/12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678
touch tst/foo
```

Run the test program on this directory, the output is:

```
1: .
2: ..
128: 12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678
```
Note, that the file `foo` is not listed.

**Expected behavior**

All directory entries should be returned, even if libssh2_sftp_readdir_ex returns LIBSSH2_ERROR_BUFFER_TOO_SMALL and the application calls again with a larger buffer.

 - OS and version: Rocky Linux 8
 - libssh2 version: 1.10.0
 - crypto backend: OpenSSL

**Additional context**

Looking at the code, there are a large number of places where a buffer too small condition is detected and it goes to the label `end`.

Then at the label end:

```
end:

            if((--handle->u.dir.names_left) == 0)
                LIBSSH2_FREE(session, handle->u.dir.names_packet);

```
ie. even though there was an error reading this entry, it still decrements the number of names left.

Perhaps the `end` label should be below this statement so that when the function returns with an error, it does not update the `names_left`.

yodaldevoid:
LIBSSH2_ERROR_BUFFER_TOO_SMALL is returned if the buffer is too small to contain a returned directory entry. On this condition we jump to the label `end`. At this point the number of names left is decremented despite no name being returned.

As suggested in #714, this commit moves the error label after the decrement of `names_left`.

Fixes #714

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -1852,11 +1852,11 @@ static ssize_t sftp_readdir(LIBSSH2_SFTP_HANDLE *handle, char *buffer,
 
             handle->u.dir.next_name = (char *) s;
             handle->u.dir.names_packet_len = names_packet_len;
-          end:
 
             if((--handle->u.dir.names_left) == 0)
                 LIBSSH2_FREE(session, handle->u.dir.names_packet);
 
+          end:
             _libssh2_debug(session, LIBSSH2_TRACE_SFTP,
                            "libssh2_sftp_readdir_ex() return %d",
                            filename_len);


willco007:
Seems nice a reasonable fix to me.

PR #757, opened 2022-10-18
Continue the authentication according to RFC 4252 section-5.1

lampmanyao:
According to RFC 4252 section-5.1, we should continue the authentication.

### DIFF
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -1005,6 +1005,8 @@ _libssh2_debug(LIBSSH2_SESSION * session, int context, const char *format, ...)
 /* Error message is allocated on the heap */
 #define LIBSSH2_ERR_FLAG_DUP                     1
 
+#define SSH_USERAUTH_PARTIAL_SUCCESS             1
+
 /* SSH Packet Types -- Defined by internet draft */
 /* Transport Layer */
 #define SSH_MSG_DISCONNECT                          1
--- a/src/userauth.c
+++ b/src/userauth.c
@@ -1408,6 +1408,281 @@ _libssh2_key_sign_algorithm(LIBSSH2_SESSION *session,
     return rc;
 }
 
+/*
+ * userauth_keyboard_interactive
+ *
+ * Authenticate using a challenge-response authentication
+ */
+static int
+userauth_keyboard_interactive(LIBSSH2_SESSION * session,
+                              const char *username,
+                              unsigned int username_len,
+                              LIBSSH2_USERAUTH_KBDINT_RESPONSE_FUNC
+                              ((*response_callback)))
+{
+    unsigned char *s;
+
+    int rc;
+
+    static const unsigned char reply_codes[4] = {
+        SSH_MSG_USERAUTH_SUCCESS,
+        SSH_MSG_USERAUTH_FAILURE, SSH_MSG_USERAUTH_INFO_REQUEST, 0
+    };
+    unsigned int i;
+
+    if(session->userauth_kybd_state == libssh2_NB_state_idle) {
+        session->userauth_kybd_auth_name = NULL;
+        session->userauth_kybd_auth_instruction = NULL;
+        session->userauth_kybd_num_prompts = 0;
+        session->userauth_kybd_auth_failure = 1;
+        session->userauth_kybd_prompts = NULL;
+        session->userauth_kybd_responses = NULL;
+
+        /* Zero the whole thing out */
+        memset(&session->userauth_kybd_packet_requirev_state, 0,
+               sizeof(session->userauth_kybd_packet_requirev_state));
+
+        session->userauth_kybd_packet_len =
+            1                   /* byte    SSH_MSG_USERAUTH_REQUEST */
+            + 4 + username_len  /* string  user name (ISO-10646 UTF-8, as
+                                   defined in [RFC-3629]) */
+            + 4 + 14            /* string  service name (US-ASCII) */
+            + 4 + 20            /* string  "keyboard-interactive" (US-ASCII) */
+            + 4 + 0             /* string  language tag (as defined in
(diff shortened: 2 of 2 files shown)


MichaelBuckley:
After re-reading RFC 4252, Section 5.1, I'm not sure this is correct. It's a bit difficult to tell for sure, since the diff shows a lot of code just being moved around. However, if I'm reading it correctly, the main change is in _libssh2_userauth_publickey(). Upon recieving SSH_MSG_USERAUTH_FAILURE, this function now:

- Checks the partial success flag, and if it is false, fails auth
- Checks the namelist, and if it's empty or does not begin with "keyboard-interactive" fails auth
- Attempts keyboard interactive auth

The problems with this are, as I read Section 5.1:
- The value of the partial success flag is informational only. If it is false, auth should not fail
- All methods in the list should be supported, not just keyboard-interactive
- Section 5.1 applies to all authentication types, not only public key authentication
- _libssh2_userauth_publickey() should not attempt keyboard interactive auth directly. Rather, it should be up to the client to decide how to proceed

In pseudocode, here's how a cient could currently write its authentication flow to to work with Section 5.1.

```
namelist = libssh2_userauth_list(session, username, username_len);
rc = LIBSSH2_ERROR_AUTHENTICATION_FAILED
while namelist is not empty && rc != LIBSSH2_ERROR_NONE {
    method = namelist[0];
    if method == "password" {
        rc = libssh2_userauth_password(session, username, password);
    } else if method == "keyboard-interactive" {
        rc = libssh2_userauth_keyboard_interactive(session, username, callback);
    } else if method == "publickey" {
        for key in keys {
            rc = libssh2_userauth_publickey_*(session, ...);
            if rc == LIBSSH2_ERROR_NONE {
                break;
            }
        }
    }

    if rc != LIBSSH2_ERROR_NONE {
        namelist = libssh2_userauth_list(session, username, username_len);
    }
}
```

The second call to libssh2_userauth_list(), at the bottom of the loop, fetches the new namelist that's returned with the SSH_MSG_USERAUTH_FAILURE packet. This allows auth to proceed, even ignoring Section 5.1. However, this does require an extra roundtrip to the server to fetch the namelist.

A better way to support Section 5.1 would be to always extract the namelist from SSH_MSG_USERAUTH_FAILURE, regardless of the partial success flag, and copy it to session->userauth_list_data. Clients could then write code like this:

```
namelist = libssh2_userauth_list(session, username, username_len);
rc = LIBSSH2_ERROR_AUTHENTICATION_FAILED
while namelist is not empty && rc != LIBSSH2_ERROR_NONE {
    method = namelist[0];
    if method == "password" {
        rc = libssh2_userauth_password(session, username, password);
    } else if method == "keyboard-interactive" {
        rc = libssh2_userauth_keyboard_interactive(session, username, callback);
    } else if method == "publickey" {
        for key in keys {
            rc = libssh2_userauth_publickey_*(session, ...);
            if rc == LIBSSH2_ERROR_NONE {
                break;
            }
        }
    }

    if rc != LIBSSH2_ERROR_NONE {
        namelist = session-> userauth_list_data;
    }
}
```

MichaelBuckley:
On closer inspection, It appears most of the diff appears to be code moving around because of automatic formatting by an editor. Unfortunately, this violates libssh2's style guide, and the CI's style-check failed.

lampmanyao:
i use libssh2 to login a server which has OpenSSH_7.2 installed, however i can't login the server, 
the error message is 'Invalid signature for supplied public key, or bad username/public key combination'.
The username and the private key are both correct, because i can login the server via 'ssh -i private_key username@host'.
I found that the server respond a '51,"keyboard-interactive"1' packet when i enable debug and trace.

I'v try the pseudocode you post, it works. Thanks.

You are right about section 5.1 and this PR.

But i am curious,
0, why OpenSSH7.2 respond a '51,"keyboard-interactive"1' packet even though the username and the private key are both correct?
1, OpenSSH seems do the authentications at the core code instead of let the user do it at the client side, the reason for libssh2 implement like this is?

IMO, the simple way is just invoke an API and the lib do the internal job for me unless some error occurs.

MichaelBuckley:
@lampmanyao 

#1 is probably better answered by someone who works on OpenSSH. I can only guess, but it sounds like it might be caused by the server configuration.

For #2, libssh2 and OpenSSH have different goals. libssh2 is a flexable, general-purpose library for SSH that's used across a lot of different device types. OpenSSH is not a library, but a full client and server that are designed to be run on desktops and server hardware. So, for example OpenSSH can look for key files in ~/.ssh, but an iOS app using libssh2 doesn't have access to that directory. Instead, the key files might be stored in the iOS keychain, and the iOS app needs to call libssh2_userauth_publickey_frommemory() instead of libssh2_userauth_publickey_fromfile().

The point is, different apps using libssh2 will need to authenticate differently, so that's why libssh2 can't do everything automatically like OpenSSH does.

It would probably be possible for libssh2 to contain the loop I described in pseudocode, and have apps register callbacks. The apps could then register callbacks with libssh2 and call a single function. Something like this:

```namelist = libssh2_userauth_list(session, username, username_len);
rc = LIBSSH2_ERROR_AUTHENTICATION_FAILED
while namelist is not empty && rc != LIBSSH2_ERROR_NONE {
    method = namelist[0];
    if method == "password" {
        rc = tryPasswordCallback(session);
    } else if method == "keyboard-interactive" {
        rc = tryKeyboardInteractiveCallback(session);
    } else if method == "publickey" {
        tryPublicKeyCallback(session);
    }

    if rc != LIBSSH2_ERROR_NONE {
        namelist = session->userauth_list_data;
    }
}```

I don't know why libssh2 wasn't designed this way. I suspect it may not be flexible enough for some use cases.

lampmanyao:
@MichaelBuckley 

thanks your reply again.

Issue #739, opened 2022-09-06
[libssh 1.10.1_dev] Error "Invalid MAC received" on Windows when cloning git+ssh


jeroen:
Using libssh2 + libgit2 + openssl compiled with mingw-w64. After upgrading from 1.10.0 to 1.10.1_dev (master branch) cloning git+ssh repos from GitHub fails with:

```
Error in libgit2::git_clone : failed to start SSH session: Invalid MAC received
```

Reverting to libssh2 1.10.1 fixes the problem, so there seems to be a regression. What could be causing this?

jeroen:
Did some debugging and it indeed fails here with 2 completely different mac addresses: 

https://github.com/libssh2/libssh2/blob/6c59eea5a9ea77127ec0fa3d6815c8adc743dba3/src/transport.c#L191-L194

One thing I noticed comparing the (working) 1.10.0 build with the dev version is that the value of `fullpacket_payload_len` is different. In the stable branch it is always 27, but in the dev version it is much larger, I have seen 59 or 539 depending on which server I use.

jeroen:
Bisected to https://github.com/libssh2/libssh2/commit/17c9c1fcdf48ea033a307ce77c432f039ead4584 (pr https://github.com/libssh2/libssh2/pull/629).
cc @haydenroche5 @bagder perhaps the change to EVP_CipherUpdate was not without side effects?

willco007:
@jeroen if you revert the `EVP_CipherUpdate` change does it work for you? We do run tests on Windows + OpenSSL which should catch a regression in transport like this unless it's some odd interplay with mingw.

jeroen:
Yes after reverting [the `EVP_CipherUpdate` change](https://github.com/libssh2/libssh2/commit/17c9c1fcdf48ea033a307ce77c432f039ead4584) it works again. Do the tests actually connect to an ssh server? I am also surprised myself to see this problem only on Windows; we specifically build with gcc and openssl to get the same behaviour as on mac/linux.

willco007:
If I'm reading this correctly, the API only appears to used with `arcfour128` cipher which is disabled in modern OpenSSH and isn't hit with the unit tests.

jeroen:
GitHub and other git servers seem to be running: `OpenSSH_8.6p1, LibreSSL 3.3.6`. I wasn't able to connect to any of them using the master branch.

willco007:
[...]

PR #766, opened 2022-10-31
Test CI build without HAVE_EVP_AES_128_CTR

jeroen:
This PR is not to be merged, but only used to trigger a CI run to test a configuration without `HAVE_EVP_AES_128_CTR`.

It shows that the implementation is broken for a configuration without `HAVE_EVP_AES_128_CTR`. This is not tested in regular CI runs. See also: https://github.com/libssh2/libssh2/issues/739

The best solution is to remove this test from autoconf/cmake, and always assume/require `EVP_aes_128_ctr` is available, and remove the fallback implementation. `EVP_aes_128_ctr` has been available for a very long time, but sometimes cmake/autoconf fails to properly detect it.


willco007:
Could you re-run this with this PR merged? https://github.com/libssh2/libssh2/commit/3fdcff2485a29544e974bf1d73dd5cfd275c6a47

jeroen:
Yeah that fixes it. Still though I think it would be better to remove the obsolete `HAVE_EVP_AES_128_CTR` stuff to prevent this from re-appearing in the future.

willco007:
That PR fixes the ciphers used when AES_128 is disabled. While I agree removing this probably is fine, it also doesn't hurt having it around and removing it won't fix the underlying issue (which was a bug fixed in https://github.com/libssh2/libssh2/commit/3fdcff2485a29544e974bf1d73dd5cfd275c6a47).

willco007:
I guess that does leave the issue with autoconf not picking it up which is still a bit mysterious.

jeroen:
> I guess that does leave the issue with autoconf not picking it up which is still a bit mysterious.

This happened because it was a windows environment with a static openssl, that needs additional linker flags to build a test program.

PR #785, opened 2023-01-05
userauth: new error code for 'continue authentication'

waylybaye:
I'm implementing 2FA support and found this PR https://github.com/libssh2/libssh2/pull/757

I think a simpler approach would be return a special error code instead of using LIBSSH2_ERROR_PUBLICKEY_UNVERIFIED

### DIFF
--- a/src/userauth.c
+++ b/src/userauth.c
@@ -1787,7 +1787,7 @@ _libssh2_userauth_publickey(LIBSSH2_SESSION *session,
         return 0;
     }
 
-    if (session->userauth_pblc_data[session->userauth_pblc_data_len - 1] == 1) {
+    if(session->userauth_pblc_data[session->userauth_pblc_data_len - 1] == 1) {
         /* This public key is accepted, but require more authentication */
         LIBSSH2_FREE(session, session->userauth_pblc_data);
         session->userauth_pblc_data = NULL;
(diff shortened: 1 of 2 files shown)


willco007:
How do you feel this better than checking the auth list for available auth methods after a failure?

src/userauth.c
```diff
@@ -1787,6 +1787,15 @@ _libssh2_userauth_publickey(LIBSSH2_SESSION *session,
         return 0;
     }
 
+    if (session->userauth_pblc_data[session->userauth_pblc_data_len - 1] == 1) {
+        /* This public key is accepted, but require more authentication */
+        LIBSSH2_FREE(session, session->userauth_pblc_data);
+        session->userauth_pblc_data = NULL;
+        session->userauth_pblc_state = libssh2_NB_state_idle;
+        return _libssh2_error(session, LIBSSH2_ERROR_PUBLICKEY_PARTIAL,
+                            "Publickey authentication partial successful");
```
vszakats, line 1796:
```suggestion
                              "Publickey authentication partially successful");
```

willco007:
My concern with this MR is it is a breaking change for existing clients. They won't be expecting this new error code.

src/userauth.c
```diff
@@ -1787,6 +1787,15 @@ _libssh2_userauth_publickey(LIBSSH2_SESSION *session,
         return 0;
     }
 
+    if (session->userauth_pblc_data[session->userauth_pblc_data_len - 1] == 1) {
```
vszakats, line 1790:
```suggestion
    if(session->userauth_pblc_data[session->userauth_pblc_data_len - 1] == 1) {
```

vszakats:
Thanks Will. Fair enough. We'd need a brand new function with this change of behaviour. This isn't likely to happen in the foreseeable future.

PR #787, opened 2023-01-08
Always clear SFTP protocol error

Zenju:
Hi,

libssh2_sftp_last_error() is documented to return a sensible error code [if libssh2 returned LIBSSH2_ERROR_SFTP_PROTOCOL in a previous call](https://www.libssh2.org/libssh2_sftp_last_error.html)

Unfortunately there are a lot of LIBSSH2_ERROR_SFTP_PROTOCOL errors in sftp.c which do *not* set LIBSSH2_SFTP::last_errno.
This means if a LIBSSH2_SFTP handle is reused, it can contain a misleading/stale value from a previous SFTP call.

Regards, Zenju


willco007:
Seems reasonable to me.

Issue #608, opened 2021-05-29
Can't distinguish wrong key from wrong passphrase


gnachman:
**Describe the bug**
When you authenticate using `libssh2_userauth_publickey_fromfile` it fails with `LIBSSH2_ERROR_FILE` when the passphrase provided for an encrypted private key is wrong, and also for many other reasons. This makes it impossible to build a UI that tells the user what to do. If the file is malformed, for example, there's no point prompting them to enter a passphrase.

**To Reproduce**
Call `libssh2_userauth_publickey_fromfile` giving it the path to an OpenSSH-formatted private key that is encrypted with a passphrase along with an empty passphrase. Note that it returns `LIBSSH2_ERROR_FILE`.

Then call it with the path to a malformed private key file (like `/etc/passwd`). Note that it returns `LIBSSH2_ERROR_FILE`.

**Expected behavior**
There should be a distinct error code for when the passphrase is wrong so that the application can prompt the user to enter a correct one.

For an OpenSSH-format private key, there is no simple way to tell before authenticating whether the private key is encrypted. Therefore it is undecidable whether to prompt the user for a passphrase.

**Version (please complete the following information):**
 - OS: macOS 11.4
 - libssh2 version 1.9.0

**Additional context**
`_libssh2_pub_priv_openssh_keyfile` returns `LIBSSH2_ERROR_KEYFILE_AUTH_FAILED`but `_libssh2_pub_priv_keyfile` changes the return code to `LIBSSH2_ERROR_FILE` which is used for many other purposes, making it impossible to tell whether the user should re-enter a passphrase.

Note that this is especially a problem specifically for OpenSSH-format private keys. Other kinds of private key indicate that they are encrypted in easier-to-parse ways.

willco007:
IIRC this is because OpenSSL returns a pretty generic error code when decoding OpenSSL style keys in a variety of situations. So to keep the API consistent it returns `LIBSSH2_ERROR_FILE`. In my client application, we do a bit of key pre-processing to determine if the key can be decoded, is encrypted, etc before passing it into the libssh2 API.

gnachman:
Clients should not be asked to parse a key file, base64-decode it, ASN.1-decode it, and so on to determine if a key is encrypted. This code already exists in libssh2 and is unlikely to be reimplemented correctly by its users.

[...]

PR #789, opened 2023-01-16
openssl: fix misleading error message if wrong passphrase

Zenju:
Hi,

if a wrong passphrase is entered by the user, libssh2_userauth_publickey_frommemory fails with:

> LIBSSH2_ERROR_PROTO: Error parsing PEM: offset out of bounds

Since this is a common error case, and errors are ultimately propagated to the end user, I believe an explicit error mentioning "wrong passphrase" should be returned instead.

Best, Zenju

---------------------
PS:
1. Perhaps error handling should be even more detailed, and the full OpenSSL message should be returned?

> ERR_error_string_n(ERR_get_error(), errorBuf, sizeof(errorBuf));

But it seems, _libssh2_error() cannot handle a non-static error message at this time.

2. Because of the OpenSSH format fallback, currently either the OpenSSH, or the OpenSSL error code is dismissed:
=> Ideally, libssh2 would search for OPENSSH_HEADER_BEGIN in "privatekeydata", and *only* then try the OpenSSH format. For a quick check, this would require something like strnstr() because "privatekeydata" is not null-terminated. e.g. https://opensource.apple.com/source/Libc/Libc-1158.30.7/string/FreeBSD/strnstr.c.auto.html


src/openssl.c
```diff
@@ -3943,9 +3945,22 @@ _libssh2_pub_priv_keyfilememory(LIBSSH2_SESSION *session,
                                                      privatekeydata,
                                                      privatekeydata_len,
                                            (unsigned const char *)passphrase);
-        if(st != 0)
-            return st;
-        return 0;
+        if(st == 0)
+            return 0;
+
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L /*OpenSSL 3.0.0*/
```
vszakats, line 3951:
```suggestion
#if OPENSSL_VERSION_NUMBER >= 0x30000000L /* OpenSSL 3.0.0 */
```

vszakats:
@libssh2/dev-team This is technically ready to merge. Any objection, opinion or review if we are good to go along with it?

vszakats:
Fixes #608

PR #791, opened 2023-01-23
sftp: remove packet limit for directory reading

Zenju:
Hi,

currently libssh2 cannot read huge directory listings when the package size of LIBSSH2_SFTP_PACKET_MAXLEN (256KB) is hit.
For example AWS always sends a single package with all files of a directory, no matter how big it is:
https://freefilesync.org/forum/viewtopic.php?t=10020
Package size is probably around 7 MB in this case!

LIBSSH2_SFTP_PACKET_MAXLEN is a good idea in general, but there doesn't seem to be a one size fits all.
While almost all(?) SFTP responses come in very small packages, I believe the SSH_FXP_READDIR request should be exempted.

The proposed patch, enhances the package size reading to include parsing the full SFTP packet header. And in case a package is of type SSH_FXP_NAME and matches an expected readdir_request_id, it does not fail if LIBSSH2_SFTP_PACKET_MAXLEN is hit.
The chances of accidentally hiding data-corruption are pretty non-existent, because both SFTP request_id and packet type must match. No change in behavior otherwise.

Best, Zenju

PS: Previous discussion: https://github.com/libssh2/libssh2/pull/268 https://github.com/libssh2/libssh2/pull/269

With the above changes, the LIBSSH2_SFTP_PACKET_MAXLEN value could (and should?) probably be set back to a small number again.

### DIFF
--- a/src/sftp.h
+++ b/src/sftp.h
@@ -153,9 +153,10 @@ struct _LIBSSH2_SFTP
     uint32_t last_errno;
 
     /* Holder for partial packet, use in libssh2_sftp_packet_read() */
-    unsigned char partial_size[4];      /* buffer for size field   */
-    size_t partial_size_len;            /* size field length       */
-    unsigned char *partial_packet;      /* The data                */
+    unsigned char packet_header[9];
+    /*packet size (4) packet type (1) request id (4) */
+    size_t packet_header_len;            /* packet_header length     */
+    unsigned char *partial_packet;      /* The data, including header */
     uint32_t partial_len;               /* Desired number of bytes */
     size_t partial_received;            /* Bytes received so far   */
 
--- a/src/sftp.h
+++ b/src/sftp.h
@@ -154,7 +154,7 @@ struct _LIBSSH2_SFTP
 
     /* Holder for partial packet, use in libssh2_sftp_packet_read() */
     unsigned char packet_header[9];
-    /*packet size (4) packet type (1) request id (4) */
+    /* packet size (4) packet type (1) request id (4) */
     size_t packet_header_len;            /* packet_header length     */
     unsigned char *partial_packet;      /* The data, including header */
     uint32_t partial_len;               /* Desired number of bytes */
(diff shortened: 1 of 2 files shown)


src/sftp.h
```diff
@@ -153,9 +153,10 @@ struct _LIBSSH2_SFTP
     uint32_t last_errno;
 
     /* Holder for partial packet, use in libssh2_sftp_packet_read() */
-    unsigned char partial_size[4];      /* buffer for size field   */
-    size_t partial_size_len;            /* size field length       */
-    unsigned char *partial_packet;      /* The data                */
+    unsigned char packet_header[9];
+    /*packet size (4) packet type (1) request id (4) */
```
vszakats, line 157:
```suggestion
    /* packet size (4) packet type (1) request id (4) */
```
make it pass checksrc.

src/sftp.h
```diff
@@ -153,9 +153,10 @@ struct _LIBSSH2_SFTP
     uint32_t last_errno;
 
     /* Holder for partial packet, use in libssh2_sftp_packet_read() */
-    unsigned char partial_size[4];      /* buffer for size field   */
-    size_t partial_size_len;            /* size field length       */
-    unsigned char *partial_packet;      /* The data                */
+    unsigned char packet_header[9];
+    /* packet size (4) packet type (1) request id (4) */
+    size_t packet_header_len;            /* packet_header length     */
```
vszakats, line 158:
```suggestion
    size_t packet_header_len;           /* packet_header length    */
```

src/sftp.h
```diff
@@ -153,9 +153,10 @@ struct _LIBSSH2_SFTP
     uint32_t last_errno;
 
     /* Holder for partial packet, use in libssh2_sftp_packet_read() */
-    unsigned char partial_size[4];      /* buffer for size field   */
-    size_t partial_size_len;            /* size field length       */
-    unsigned char *partial_packet;      /* The data                */
+    unsigned char packet_header[9];
+    /* packet size (4) packet type (1) request id (4) */
+    size_t packet_header_len;           /* packet_header length    */
+    unsigned char *partial_packet;      /* The data, including header */
```
vszakats, line 159:
```suggestion
    unsigned char *partial_packet;      /* The data, with header   */
```

vszakats:
@libssh2/dev-team This is technically ready to merge. Any objection, opinion or review if we are good to go along with it?

bagder (approved):
Looks good to me!

PR #803, opened 2023-02-24
Option to not quote scp paths

jorgensigvardsson:
A new flag named `LIBSSH2_FLAG_QUOTE_PATHS` has been added, to make libssh2 not quote file paths sent to the remote's scp sub system. Some custom ssh daemons cannot handle quoted paths, and this makes this flag useful.

### DIFF
--- a/src/scp.c
+++ b/src/scp.c
@@ -299,9 +299,21 @@ scp_recv(LIBSSH2_SESSION * session, const char *path, libssh2_struct_stat * sb)
                  "scp -%sf ", sb?"p":"");
 
         cmd_len = strlen((char *)session->scpRecv_command);
-        cmd_len += shell_quotearg(path,
-                                  &session->scpRecv_command[cmd_len],
-                                  session->scpRecv_command_len - cmd_len);
+
+        if(!session->flag.quote_paths) {
+            size_t path_len;
+
+            path_len = strlen(path);
+
+            /* no NUL-termination neeed, so memcpy will do */
+            memcpy(&session->scpRecv_command[cmd_len], path, path_len);
+            cmd_len += path_len;
+        }
+        else {
+            cmd_len += shell_quotearg(path,
+                                    &session->scpRecv_command[cmd_len],
+                                    session->scpRecv_command_len - cmd_len);
+        }
 
         /* the command to exec should _not_ be NUL-terminated */
         session->scpRecv_command_len = cmd_len;
@@ -860,9 +872,22 @@ scp_send(LIBSSH2_SESSION * session, const char *path, int mode,
                  "scp -%st ", (mtime || atime)?"p":"");
 
         cmd_len = strlen((char *)session->scpSend_command);
-        cmd_len += shell_quotearg(path,
-                                  &session->scpSend_command[cmd_len],
-                                  session->scpSend_command_len - cmd_len);
+
+        if(!session->flag.quote_paths) {
+            size_t path_len;
+
+            path_len = strlen(path);
+
+            /* no NUL-termination neeed, so memcpy will do */
+            memcpy(&session->scpSend_command[cmd_len], path, path_len);
+            cmd_len += path_len;
+
+        }
+        else {
+            cmd_len += shell_quotearg(path,
+                                    &session->scpSend_command[cmd_len],
+                                    session->scpSend_command_len - cmd_len);
+        }
 
         /* the command to exec should _not_ be NUL-terminated */
         session->scpSend_command_len = cmd_len;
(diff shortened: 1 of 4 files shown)


jorgensigvardsson:
Fixed the formatting issues, and the use of `strcat` (I know it's bad form, but it was justified in this case).

jorgensigvardsson:
Argh. This is kind of silly, the use of `strncat` in this patch is totally legit. The string buffer needed is smaller or equal to the allocated memory, because the patch allows the caller to opt out of quoting paths.

I could homeroll a similar function, but it'd be more work validating that function than the semantics of this style-check. :man_shrugging:

willco007:
My first thought is use `memcpy`, should work the same, just make sure you grab the null terminator if you need it.

jorgensigvardsson:
That's one way to get around the style check issue, sure. Using `memcpy` is more complicated than `strncat`, because not only do I have to offset the destination pointer, I also, as you point out, not forget the null terminator.

Oh well, if `memcpy` is what it takes, then OK. :)

By the way, do you have any interest of this PR? It's not a big contribution, nor would it be used a lot, but it has helped us a lot when communicating with an embedded system over which we (my team) have very little influence. I was inspired by another C#/.NET library that implemented something called "path transforms". Not sure how big of a problem paths really are, but if it's a common misery, then maybe an option to make the entire shell quote function replaceable would be beneficial?

jorgensigvardsson:
Ok, after thinking about it, `memcpy` is the way to go. When reading the rest of the code and comments in `session.c` regarding scp functionality, the function `_libssh2_channel_process_startup` does not expect a NUL-terminated string. So, `strncpy` would've been wrong.

src/scp.c
```diff
@@ -273,6 +273,7 @@ scp_recv(LIBSSH2_SESSION * session, const char *path, libssh2_struct_stat * sb)
     size_t cmd_len;
     int rc;
     int tmp_err_code;
+    size_t path_len;
```
vszakats, line 276: Can we move this to the `if()` block that uses the variable?

src/scp.c
```diff
@@ -838,6 +847,7 @@ scp_send(LIBSSH2_SESSION * session, const char *path, int mode,
...
```
vszakats, line 850: Also this.

src/scp.c
```diff
@@ -860,9 +870,18 @@ scp_send(LIBSSH2_SESSION * session, const char *path, int mode,
                  "scp -%st ", (mtime || atime)?"p":"");
 
         cmd_len = strlen((char *)session->scpSend_command);
-        cmd_len += shell_quotearg(path,
-                                  &session->scpSend_command[cmd_len],
-                                  session->scpSend_command_len - cmd_len);
+
+        if(!session->flag.quote_paths) {
+            path_len = strlen(path);
+            memcpy(&session->scpSend_command[cmd_len], path, path_len); /* no NUL-termination neeed, so memcpy will do */
```
vszakats, line 876: Comment needs to be moved to a separate line to pass `checksrc`.

src/scp.c
jorgensigvardsson: Done, pushed a second ago.

src/scp.c
jorgensigvardsson: Done, pushed a second ago.

src/scp.c
jorgensigvardsson: Done, pushed a second ago.

src/scp.c
```diff
@@ -299,9 +299,21 @@ scp_recv(LIBSSH2_SESSION * session, const char *path, libssh2_struct_stat * sb)
...
+            path_len = strlen(path);
+
+            /* no NUL-termination neeed, so memcpy will do */
+            memcpy(&session->scpRecv_command[cmd_len], path, path_len);
+            cmd_len += path_len;
+        }
+        else {
+            cmd_len += shell_quotearg(path,
+                                    &session->scpRecv_command[cmd_len],
+                                    session->scpRecv_command_len - cmd_len);
```
vszakats, line 315:
```suggestion
                                      &session->scpRecv_command[cmd_len],
                                      session->scpRecv_command_len - cmd_len);
```
Should still fit withing checksrc limits.

src/scp.c
```diff
@@ -860,9 +872,22 @@ scp_send(LIBSSH2_SESSION * session, const char *path, int mode,
...
+
+            /* no NUL-termination neeed, so memcpy will do */
+            memcpy(&session->scpSend_command[cmd_len], path, path_len);
+            cmd_len += path_len;
+
+        }
+        else {
+            cmd_len += shell_quotearg(path,
+                                    &session->scpSend_command[cmd_len],
+                                    session->scpSend_command_len - cmd_len);
```
vszakats, line 889:
```suggestion
                                      &session->scpSend_command[cmd_len],
                                      session->scpSend_command_len - cmd_len);
```

Issue #793, opened 2023-02-06
Hitting error -44 "Unable to ask for ssh-userauth service" with mbedtls


green-nsk:
**Describe the bug**

When downloading from SFTP via libcurl, I am hitting error code -44 with description "Unable to ask for ssh-userauth service". This happens for one particular SFTP server, and others work fine.

**To Reproduce**

I _think_ it has to do with server using aes256-cbc algorithm. As I have no control over the server, I'm trying to verify that assumption by changing advertised set of Ciphers on my side.

**Expected behavior**

I would like it to work :-D

**Version (please complete the following information):**
 - OS and version: Julia 1.8.5 running on Ubuntu Linux 22.04
 - libssh2 version: 1.10.0
 - crypto backend: MbedTLS

**Additional context**

It seems to fail to send encrypt the message before sending it. Relevant code [here](https://github.com/libssh2/libssh2/blob/821d50dad313b53fb2782f26aec1f52f1be34fc0/src/session.c#L775) and [here](https://github.com/libssh2/libssh2/blob/1e1a74eb7e97b004ccad86a32c1874eb0d986d6f/src/transport.c#L896).

I have found a similar issue [here](https://github.com/libssh2/libssh2/issues/463) which was resolved by fixing how openssl functions called [here](https://github.com/libssh2/libssh2/pull/556). I assume this might be a similar issue with mbedtls, but I'd be hard pressed to find out exactly what's wrong.

dfandrich:
I'm adding some end-to-end data transfer tests in PR#807 and the CI tests are showing all the following crypt methods failing with this error for MbedTLS:

```
The following tests FAILED:
	 13 - test_read_3des-cbc (Failed)
	 14 - test_read_aes128-cbc (Failed)
	 16 - test_read_aes192-cbc (Failed)
	 18 - test_read_aes256-cbc (Failed)
	 20 - test_read_rijndael-cbc (Failed)
```

vszakats:
@green-nsk Is this happening with mbedTLS 2.x or 3.x?

green-nsk:
@vszakats  it's 2.28.0 (built-in Julia one)

Markus-Schmidt:
I believe I have a similar (related) issue, also with a server over which I have no control.
This is a windows build with openssl, version libssh2 1.10 which works fine in any other regard.

When the client (libssh2) requests the userauth service the reply packet is apparently mangled.

In session.c at libssh2_NB_state_sent4 it calls 
[...]

PR #807, opened 2023-03-01
Add tests to read lots of data over a channel

dfandrich:
Additionally, add tests of individual crypt and hmac methods.


dfandrich:
This test rediscovered issue #793 and would have discovered some problems with PR#797 earlier had it been available then. I think this will be valuable.

dfandrich:
If someone wants to check the cmake parts of this it's a good idea. I have a feeling I'm doing some things in a noncanonical way.

PR #733, opened 2022-08-08
Fix macro redefinition warnings on MSVC

ien646:
Fixes the following type of warnings on windows:

```
[build] C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\shared\ntstatus.h(66,1): warning C4005: 'STATUS_WAIT_0': macro redefinition [***\build\libssh2\src\libssh2.vcxproj]
[build] C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\um\winnt.h(2510): message : see previous definition of 'STATUS_WAIT_0' [***\build\libssh2\src\libssh2.vcxproj]
[build] C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\shared\ntstatus.h(222,1): warning C4005: 'STATUS_ABANDONED_WAIT_0': macro redefinition [***\build\libssh2\src\libssh2.vcxproj]
[build] C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\um\winnt.h(2511): message : see previous definition of 'STATUS_ABANDONED_WAIT_0' [***\build\libssh2\src\libssh2.vcxproj]
[build] C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\shared\ntstatus.h(245,1): warning C4005: 'STATUS_USER_APC': macro redefinition [***\libssh2\src\libssh2.vcxproj]
[build] C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\um\winnt.h(2512): message : see previous definition of 'STATUS_USER_APC' 
...
```

Issue #804, opened 2023-02-25
WinCNG test failures


dfandrich:
Most runs, the WinCNG openssh test runs on Appveyor fail at least one test. The actual test changes (e.g. test_password_auth_fails_with_wrong_username, test_keyboard_interactive_auth_fails_with_wrong_response, test_hostkey) and sometimes the WinCNG runs complete without failures, but it's pretty common that at least one fails. It's often on the 
VS2015, WinCNG, Shared, x86_64 build (but maybe because that's one of the first WinCNG builds and the tests stop after the first error), but it happens on other WinCNG builds, too, including VS2013, Static and x86 builds, but always with WinCNG.

The tests fail with the error `9: libssh2_session_handshake failed (-8): Unable to exchange encryption keys`. Running the tests with `FIXTURE_TRACE_ALL=1` showed this log at the time of failure:

```
[00:05:13] 5: [libssh2] 0.951417 Transport: Packet type 33 received, length=1078
[00:05:13] 5: [libssh2] 0.951417 Transport: Looking for packet of type: 33
[00:05:13] 5: [libssh2] 0.951417 Key Ex: Server's MD5 Fingerprint: 0c:0e:d1:a5:bb:10:27:5f:76:92:4c:e1:87:ce:5c:5e
[00:05:13] 5: [libssh2] 0.951417 Key Ex: Server's SHA1 Fingerprint: f3:cd:59:e2:91:3f:44:22:b8:0f:7b:0a:82:b2:b8:9e:ae:44:93:87
[00:05:13] 5: [libssh2] 0.951417 Key Ex: Server's SHA256 Fingerprint: kuPaSd88f5moKPUF7YI5OXpdH2KRRFl2D4ePdRD1Y6M=
[00:05:13] 5: [libssh2] 1.045165 Failure Event: -11 - Unable to verify hostkey signature
[00:05:13] 5: [libssh2] 1.045165 Failure Event: -8 - Unrecoverable error exchanging keys
[00:05:13] 5: [libssh2] 1.045165 Failure Event: -8 - Unable to exchange encryption keys
[00:05:13] 5: libssh2_session_handshake failed (-8): Unable to exchange encryption keys
[00:05:13] 5: [libssh2] 1.045165 Transport: Disconnecting: reason=11, desc=test ended, lang=
[00:05:13] 5: => libssh2_transport_write plain (23 bytes)
[...]

PR #808, opened 2023-03-01
build fixes and improvements (mostly for Windows)

vszakats:
- in `hostkey.c` check the result of `libssh2_sha256_init()` and
  `libssh2_sha512_init()` calls. This avoid the warning that we're
  ignoring the return values.

- fix code using `int` (or `SOCKET`) for sockets. Use libssh2's
  dedicated `libssh2_socket_t` and `LIBSSH2_INVALID_SOCKET` instead.

- fix compiler warnings due to `STATUS_*` macro redefinitions between
  `ntstatus.h` / `winnt.h`. Solve it by manually defining the single
  `STATUS` value we need from `ntstatus.h` and stop including the whole
  header.
  Fixes #733

- improve Windows UWP/WinRT builds by detecting it with code copied
  from the curl project. Then excluding problematic libssh2 parts
  according to PR by Dmitry Kostjučenko.
  Fixes #734

- always use `SecureZeroMemory()` on Windows.

  We can tweak this if not found or not inlined by a C compiler which
  we otherwise support. Same if it causes issues with UWP apps.

  Ref: https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa366877(v=vs.85)
  Ref: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlsecurezeromemory

- always enable `LIBSSH2_CLEAR_MEMORY` on Windows. CMake and
  curl-for-win builds already did that. Delete `SecureZeroMemory()`
  detection from autotools' WinCNG backend logic, that this
  setting used to depend on.

  TODO: Enable it for all platforms in a separate PR.
  TODO: For clearing buffers in WinCNG, call `_libssh2_explicit_zero()`,
        insead of a local function or explicit `SecureZeroMemory()`.

- Makefile.inc: move `os400qc3.h` to `HEADERS`. This fixes
  compilation on non-unixy platforms. Recent regression.

- `libssh2.rc`: replace copyright with plain ASCII, as in curl.

  Ref: curl/curl@1ca62bb
  Ref: curl/curl#7765
  Ref: curl/curl#7776

- CMake fixes and improvements:

  - enable warnings with llvm/clang.
  - enable more comprehensive warnings with gcc and llvm/clang.
    Logic copied from curl:
    https://github.com/curl/curl/blob/233810bb5f6c5e7bedfc10bdd36607b958c0cfe4/CMakeLists.txt#L131-L148
  - fix `Policy CMP0080` CMake warning by deleting that reference.
  - add `ENABLE_WERROR` (default: `OFF`) option. Ported from curl.
  - add `PICKY_COMPILER` (default: `ON`) option, as known from curl.

    It controls both the newly added picky warnings for llvm/clang and
    gcc, and also the pre-existing ones for MSVC.

- `win32/GNUmakefile` fixes and improvements:

  - delete `_AMD64_` and add missing `-m64` for x64 builds under test.
  - add support for `ARCH=custom`.
    It disables hardcoded Intel 64-bit and Intel 32-bit options,
    allowing ARM64 builds.
  - add support for `LIBSSH2_RCFLAG_EXTRAS`.
    To pass custom options to windres, e.g. in ARM64 builds.
  - add support for `LIBSSH2_RC`. To override `windres`.
  - delete support for Metrowerks C. Last released in 2004.

- `win32/libssh2_config.h`: delete unnecessary socket #includes

  `src/libssh2_priv.h` includes `winsock2.h` and `ws2tcpip.h` further
   down the line, triggered by `HAVE_WINSOCK2_H`.

  `mswsock.h` does not seem to be necessary anymore.

  Double-including these (before `windows.h`) caused compiler failures
  when building against BoringSSL and warnings with LibreSSL. We could
  work this around by passing `-DNOCRYPT`. Deleting the duplicates
  fixes these issues.

  Timeline:
  2013: c910cd382dfa07fed2adaabf688af9e4a084fa1d deleted `mswsock.h` from `src/libssh2_priv.h`
  2008: 8c43bc52b1e3de2c8fc7899a80aec0e98de4e2d8 added `winsock2.h` and `ws2tcpip.h` to `src/libssh2_priv.h`
  2005: dc4bb1af967d2c53e90349f2f37324c622e714f5 added the now deleted #includes

- delete or replace `LIBSSH2_WIN32` with `WIN32`.

- replace hand-rolled `HAVE_WINDOWS_H` macro with `WIN32`. Also delete
  its detections/definitions.

- delete unused `LIBSSH2_DARWIN` macro.

- delete unused `writev()` Windows implementation

  There is no reference to `writev()` since 2007-02-02, commit
  9d55db6501aa4e21f0858cf36cdc2ddc11b96e83.

- fix a bunch of MSVC / llvm/clang / gcc compiler warnings:

  - `warning C4100: '...': unreferenced formal parameter`
  - using value of undefined PP macro `LIBSSH2DEBUG`
  - missing void from function definition
  - `if()` block missing in non-debug builds
  - unreferenced variable in non-debug builds
  - `warning: must specify at least one argument for '...' parameter of variadic macro [-Wgnu-zero-variadic-macro-arguments]`
    in `_libssh2_debug()`
  - `warning C4295: 'ciphertext' : array is too small to include a terminating null character`
  - `warning C4706: assignment within conditional expression`
  - `warning C4996: 'inet_addr': Use inet_pton() or InetPton() instead or
      define _WINSOCK_DEPRECATED_NO_WARNINGS to disable deprecated API warnings`
    By suppressning it. Would be best to use inet_pton() as suggested.
    On Windows this needs Vista though.
  - `warning C4152: nonstandard extension, function/data pointer conversion in expression`
    (silenced locally)
  - `warning C4068: unknown pragma`

  Ref: https://ci.appveyor.com/project/libssh2org/libssh2/builds/46354480/job/j7d0m34qgq8rag5w


willco007:
This all looks good so far, thanks for all this work!

vszakats:
Thank you @willco007!

vszakats:
What still needs to be addressed, but in a future PR:
- Bumping up warnings in autotools builds. CI uses these, so it'd be useful.
- Fixing the many `-Wconversion` warnings, some cast-induced alignment warnings, and printf mask issues.
- Enabling `LIBSSH2_CLEAR_MEMORY` for non-Windows systems. I have a patch for that.
- Some Windows tests on AppVeyor seem flaky. They also exit after the first flakyness.

dfandrich:
The flaky tests are issue #804.

PR #814, opened 2023-03-04
make new tests more friendly to cross-builds

vszakats:
Try to improve tests added in #807 by triggering `test_read` commands directly. This makes external batch/shell files unnecessary, and is friendlier with cross-builds, or builds run from non-default shells, like MSYS2.


tests/CMakeLists.txt
```diff
@@ -174,40 +174,36 @@ foreach(test ${TESTS})
...
+  hmac-sha1
+  hmac-sha1-96
+  hmac-sha2-256
+  hmac-sha2-512
   )
   add_test(
-      NAME test_${test} COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/test_${test}${test_ext}")
-  set_tests_properties(test_${test} PROPERTIES ENVIRONMENT "builddir=$<TARGET_FILE_DIR:test_read>")
+      NAME test_${test} COMMAND "${TARGET_FILE_DIR}/test_read")
+  set_tests_properties(test_${test} PROPERTIES ENVIRONMENT "FIXTURE_TEST_CRYPT=${test}")
```
dfandrich, line 190: The MAC tests above need to set FIXTURE_TEST_MAC, not FIXTURE_TEST_CRYPT.

dfandrich (dismissed):
I'm glad someone who knows cmake is looking at this!

tests/CMakeLists.txt
vszakats: Oh, many thanks Dan! :)

dfandrich:
I should mention that the other reason I created a file per test is that on autoconf builds, there's no other way to easily run the crypt & hash tests since they're not integrated into automake yet. I'm not sure if automake allows the kind of environment variable override that this PR is doing in the cmake builds. If not, then we should keep the test script stubs so that we can run the tests in automake as well.

vszakats:
For autotools, could a single script work instead of one per test? Since autotools needs a unixy shell anyway, do we also need the batch files?

dfandrich:
We don't need batch files, so if cmake is the only way for Windows people
without a Bourne shell to run tests, those can go. But, I'd rather see each
crypt and hash method test show up individually in automake's check output
rather than lumping them all into a single test.

vszakats:
Autotools never executed these new tests [^1], so this feels out of scope for this PR.

[^1]: https://github.com/libssh2/libssh2/actions/runs/4307684403/jobs/7513053415

For the heck of it, I've made a single script (easier to read, extend, and reuse with any build tool). Appended it to autotools' `TESTS` to get it run. Bear in mind this is the first time I'm dealing with `make check`.

UPDATE 1: It run successfully on Linux [^2], failed on MSYS2 [^3]. autotools stdout helpfully says: "See tests/test-suite.log". The actual test output is in that file. Some research turned up a [thread](https://lists.gnu.org/archive/html/automake/2018-09/msg00015.html) with a solution: `make check VERBOSE=1`.

UPDATE 2: Solution works [^4], with the output visible for each individual test. MSYS2 fails because "Failed to build docker image", so we now have another problem.

[^2]: https://github.com/libssh2/libssh2/actions/runs/4338492263/jobs/7575227478
[^3]: https://github.com/libssh2/libssh2/actions/runs/4338492263/jobs/7575226145
[^4]: https://github.com/libssh2/libssh2/actions/runs/4338656472/jobs/7575506729

dfandrich:
My near-term goal is to get them running on autotools as well. The only thing
that has stopped me from doing it yet is that the way the test suite gets the
IP address from Docker isn't working on my machine (this is a problem mentioned
by someone else, too) and I didn't want to submit anything without testing it
locally first.

If it working right, it should say TOTAL: 35 (or whatever) and the end of the
test run, not TOTAL: 3. The latter is confusing to the user, even if all the
tests were actually run.

vszakats:
Shell scripts restored.

PR #846, opened 2023-03-14
src: address compiler warnings

vszakats:
Bring down C compiler warnings (as seen in CI and local builds) from around 400 to zero.

To reach this: add casts, adjust (mostly extending) types, add missing declarations, stop reusing variables, add a few size checks, add FIXMEs where type extension to a public API would be useful, adjusted platform guards and printfs masks, simplify `_libssh2_ntohu64()`, rework WinCNG Win32 interfaces with alignment warnings.

Also:
- set `ENABLE_WERROR` in CMake CI builds.

Pending:
- [x] Close this without merging once #876, #877, #878, #879, #880 were merged.
- [x] Delete public header changes in favour of #862
- [x] Delete example/tests changes in favour of #861
- [x] WinCNG alignment [warnings](https://github.com/libssh2/libssh2/pull/846#issuecomment-1474805261).
- [x] Rebase on: `_libssh2_dh_secret()` memory leak fix → #858
- [x] Rebase on: #857
  - `ENABLE_WERROR=ON` seems to enable this for feature tests, causing `Performing Test HAVE_O_NONBLOCK - Failed` on Linux.
  - `ENABLE_WERROR=ON` breaks detection of `HAVE_IOCTLSOCKET` on Windows.
  - Rebase on: `ENABLE_WERROR=ON` breaks detection of `strtoll`, `_strtoi64`, `snprintf`, `_stricmp` on Windows with MSVC.
  - Rebase on: Use `cmake_push_check_state()` in `src/CMakeLists.txt` in place of `SAVE_*` methods (2x).
  - Rebase on: example: Replace `strcasecmp()` with `strcmp()` or other method, and delete detection logic.
  - Rebase on: example: Replace a single `__func__` with the function name, and delete detection logic.

PR #849, opened 2023-03-16
appveyor.yml: reorder tests to return relevant feedback earlier

vszakats:
- build x64 first

  x64 is the more interesting target. Most type conversion issues are revealed here. Also more commonly used by now.

- test VS 2013 earlier

- test WinCNG earlier

- delete reference to no longer used VS 2008

After this patch we end up starting with all Shared builds (2015, 2013, OpenSSL, WinCNG), then continue with Static ones. Shared/Static makes a minor if any difference in builds/tests compared to different VS versions of TLS backends.

--

CI run times:

Preparation + build takes:
8 x VS2015 4.5 mins -> total: 36
8 x VS2013 2   mins -> total: 16
Total: 52 mins

with our 30 tests, it increases to:
8 x VS2015 8-10 mins -> total: 72
8 x VS2013 6- 9 mins -> total: 60
Total: 132 mins

Without tests: https://ci.appveyor.com/project/libssh2org/libssh2/builds/46475315
With    tests: https://ci.appveyor.com/project/libssh2org/libssh2/builds/46480549

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,19 +33,9 @@ environment:
       BUILD_SHARED_LIBS: ON
       CRYPTO_BACKEND: "OpenSSL"
 
-    - job_name: "VS2015, OpenSSL, Static"
-      GENERATOR: "Visual Studio 14 2015"
-      BUILD_SHARED_LIBS: OFF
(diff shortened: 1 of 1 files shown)


vszakats:
Tested as part of #846.

PR #851, opened 2023-03-16
ci: add timeouts to ctest and GitHub CI jobs

vszakats:
- `ctest` shows a the default timeout '10000000', which may cause
  infinite waits e.g. in case the necessary server worker did not
  launch.

  CMake CI tests currently take approx:
  - GitHub   / Linux  : 125 seconds
  - AppVeyor / Windows: 300 seconds

  New defaults are: 450 and 900 seconds respectively.

- set timeouts for style-check, fuzz, Linux and Windows GitHub CI
  jobs to avoid hanging forever.

Also:

- move `choco install` to before_test to make builds start faster
  in `appveyor.yml`.

- fix some yamllint `ON`/`OFF`-confusion issue by quoting these
  values in `appveyor.yml`.

- fix indentation in `appveyor.yml`.

- convert to GitHub workflows to LF line-ending.

Closes #851


vszakats:
This is probably wrong because GitHub CI tests appear to rely on the GitHub Action being launched by AppVeyor CI. So the earlier it's started, the better.

I'll convert this change to a comment stating the above. Perhaps also adding `--timeout` to `ctest` to override the default `10000000` (seconds? milliseconds?).

Issue #790, opened 2023-01-17
When  next release is planned?


liranHaim:
Hello
I have an issue connecting to  OpenSSH_8.7p1 and OpenSSL 3.0.1 servers with the  last version 1.10
I  know that it was fixed in the main branch 
So I wait for the new release with the fix.
Thanks in advance

xeno-john:
I was going to ask the same thing. On the official site it states "The latest release: [libssh2 1.10.0](https://www.libssh2.org/download/libssh2-1.10.0.tar.gz) from August 29th 2021"

That version is kind of old. When can we expect a new release?

Zenju:
For what it's worth, my app is currenly using the master version from 2022-11-11, with no reported issues (*)

(*) except for the pull requests opened by me.

sbz:
Hi,

I would be interested also to know when the next release will be. It's been few years since last [release 1.10](https://github.com/libssh2/libssh2/releases/tag/libssh2-1.10.0) and https://github.com/libssh2/libssh2/issues/572

@dfandrich @vszakats @willco007  @bagder could you help here with a new release?

adamyg:
Interested in formal release for mbedTLS 3.x.x integration.

fxcoudert:
As a maintainer for Homebrew, we would like to ship a more recent release without carrying patches. I know other projects also do, like Julia, which ships patches for https://github.com/libssh2/libssh2/pull/711 and https://github.com/libssh2/libssh2/pull/650

dfandrich:
Under the mantra of "release early, release often" I think a release now makes sense. Yes there are a lot of PRs and issues outstanding, but there are also enough improvements already submitted to warrant a release now. We can always do another release in a few months once some of some of the remaining know issues have been dealt with.

@willco007 is there anything that *has* to go in now? Current git seems pretty stable. @bagder has volunteered to handle the mechanics of a release once you give the all clear.

willco007:
It's in good shape, nothing else has to go in to this release. Nice to haves may be this outstanding PR https://github.com/libssh2/libssh2/pull/712 and a null check to resolve this: https://github.com/libssh2/libssh2/issues/802 .

dfandrich:
Is there someone around who can approve PR#712? My cmake-fu is pretty slim.

jakob:
@willco007 Since PR #721 was closed and reopened as PR #816, which was merged, it should be good to go now?

I'd love to go back to using official releases, I've been using patched versions for too long.

[...]

PR #860, opened 2023-03-19
openssl: require `EVP_aes_128_ctr()` support

### LINKED ISSUE #739: [libssh 1.10.1_dev] Error "Invalid MAC received" on Windows when cloning git+ssh

vszakats:
libssh2 built with OpenSSL and without its `EVP_aes_128_ctr()`, aka
`HAVE_EVP_AES_128_CTR`, option are working incorrectly. This option
wasn't always auto-detected by autotools up until recently (#811).
Non-cmake, non-autotools build methods never enabled it automatically.

OpenSSL supports this options since at least v1.0.2, which is already
EOLed and considered obsolete. OpenSSL forks (LibreSSL, BoringSSL)
supported it all along.

In this patch we enable this option unconditionally, now requiring
OpenSSL supporting this function, or one of its forks.

Also modernize OpenSSL lib references to what 1.0.2 and newer versions
have been using.

Fixes #739


jeroen:
Thanks so much for addressing this. Would it be possible to tag a release, so that distros can pick up on this fix?

vszakats:
You're welcome! We track the question of when to make a new release here: https://github.com/libssh2/libssh2/issues/790

PR #866, opened 2023-03-22
ci: add MSVS 2022 jobs to appveyor

vszakats:


vszakats:
MSVS 2022 builds crash on each test, ARM support missing. Dropping this effort.

PR #892, opened 2023-03-28
Configurable session read timeout

danielsilverstone-ct:
This set of changes provides a mechanism to runtime-configure the previously `#define`'d timeout for reading packets from a session.  The intention here is to also extend `libcurl` to be able to use this interface so that when fetching from sftp servers which are very slow to return directory listings, connections do not time-out so much.

The curl patches are staged here: https://github.com/danielsilverstone-ct/curl/tree/configurable-sftp-timeout but I need to adjust them for a prediction of the libssh2 version which might contain this change if accepted.

### DIFF
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -916,6 +916,10 @@ LIBSSH2_API void libssh2_session_set_timeout(LIBSSH2_SESSION* session,
                                              long timeout);
 LIBSSH2_API long libssh2_session_get_timeout(LIBSSH2_SESSION* session);
 
+LIBSSH2_API void libssh2_session_set_read_timeout(LIBSSH2_SESSION* session,
+                                                  long timeout);
+LIBSSH2_API long libssh2_session_get_read_timeout(LIBSSH2_SESSION* session);
+
 /* libssh2_channel_handle_extended_data is DEPRECATED, do not use! */
 LIBSSH2_API void libssh2_channel_handle_extended_data(LIBSSH2_CHANNEL *channel,
                                                       int ignore_mode);
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -1092,8 +1092,8 @@ ssize_t _libssh2_recv(libssh2_socket_t socket, void *buffer,
 ssize_t _libssh2_send(libssh2_socket_t socket, const void *buffer,
                       size_t length, int flags, void **abstract);
 
-#define LIBSSH2_READ_TIMEOUT 60 /* generic timeout in seconds used when
-                                   waiting for more data to arrive */
+#define LIBSSH2_DEFAULT_READ_TIMEOUT 60 /* generic timeout in seconds used when
+                                           waiting for more data to arrive */
 
 
 int _libssh2_kex_exchange(LIBSSH2_SESSION * session, int reexchange,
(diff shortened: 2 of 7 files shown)


include/libssh2.h
```diff
@@ -926,6 +926,10 @@ LIBSSH2_API void libssh2_session_set_timeout(LIBSSH2_SESSION* session,
                                              long timeout);
 LIBSSH2_API long libssh2_session_get_timeout(LIBSSH2_SESSION* session);
 
+LIBSSH2_API void libssh2_session_set_read_timeout(LIBSSH2_SESSION* session,
+                                             long timeout);
```
vszakats, line 930: It'd be nice to indent the continued line.

src/libssh2_priv.h
```diff
@@ -1094,7 +1097,7 @@ ssize_t _libssh2_recv(libssh2_socket_t socket, void *buffer,
 ssize_t _libssh2_send(libssh2_socket_t socket, const void *buffer,
                       size_t length, int flags, void **abstract);
 
-#define LIBSSH2_READ_TIMEOUT 60 /* generic timeout in seconds used when
+#define LIBSSH2_DEFAULT_READ_TIMEOUT 60 /* generic timeout in seconds used when
                                    waiting for more data to arrive */
```
vszakats, line 1101: Here also.

src/packet.c
```diff
@@ -1267,7 +1267,7 @@ _libssh2_packet_require(LIBSSH2_SESSION * session, unsigned char packet_type,
         }
         else if(ret == 0) {
             /* nothing available, wait until data arrives or we time out */
-            long left = LIBSSH2_READ_TIMEOUT - (long)(time(NULL) -
+            long left = session->packet_read_timeout - (long)(time(NULL) -
                                                       state->start);
```
vszakats, line 1271: and here too.

include/libssh2.h
danielsilverstone-ct: Oops, thank you for spotting that.

src/libssh2_priv.h
danielsilverstone-ct: And again :D

src/packet.c
danielsilverstone-ct: We must have been asleep at the indentation wheel, fixed this one too, thank you.

vszakats (approved):

src/session.c
```diff
@@ -1469,6 +1470,30 @@ libssh2_session_get_timeout(LIBSSH2_SESSION * session)
...
 
+/* libssh2_session_set_read_timeout
+ *
+ * Set a session's timeout (in sec) when reading packets,
+ * or 0 to use default of 60 seconds.
+ */
+LIBSSH2_API void
+libssh2_session_set_read_timeout(LIBSSH2_SESSION * session, long timeout)
+{
+    if(timeout == 0) {
```
vszakats, line 1481: Would it be better to check for `timeout <= 0` here?

danielsilverstone-ct: Yes, that's probably a good plan, I'll do that and push an update.  Do you think that's the last bit for this?

vszakats: Thanks! Yes, with that I'll merge after CI is done.

vszakats:
Thank you @danielsilverstone-ct!

vszakats:
@danielsilverstone-ct: No worries and good luck with your libcurl patch.

[ 1.11.0 would be my guess for the next version, given the amount of small updates and the changes around the build systems and source tree. ]

PR #947, opened 2023-04-10
sftp: add open functions with custom attribute support

vszakats:
Before this patch, libssh2 sent hardcoded `LIBSSH2_SFTP_ATTRIBUTES`
struct on handle open. This can be problematic on some special OS,
where the file size should be known on new file creation. I added
two new functions to resolve this issue.

Patch-by: @vajdaakos on github via #506

Changes compared to #506:
- drop attr size fixup in favour of #946.
- move `memcpy()` under the state where we need it.
- bump filename length type to `size_t`.
- fix filenames in documentation and other nits.

Closes #506
Closes #947

---

Prerequisite:
- [x] #946

### DIFF
--- a/docs/Makefile.am
+++ b/docs/Makefile.am
@@ -138,6 +138,8 @@ dist_man_MANS = \
   libssh2_sftp_mkdir_ex.3 \
   libssh2_sftp_open.3 \
   libssh2_sftp_open_ex.3 \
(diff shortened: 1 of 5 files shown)


vszakats:
Are we happy with the new function names?: `libssh2_sftp_open_ex_r()`, `libssh2_sftp_open_r()`

PR #953, opened 2023-04-11
Make OS/400 implementation work again

monnerat:
After more than 6 years without an IBM i, I have been able to do some work on libssh2 for it thanks to pub400.com.
During this time, some "injuries" were done to this implementation. The commits in this PR contribute all together to fix them:

- Min target version is now V7R3M0. The build script now features ASCII C runtime development files in a non-standard path.
- A code style cleanup introduced a bug in big numbers: fixed.
- A pointer target type function parameter was incompatible: fixed.
- Block length in os400qc3 cipher definitions were wrong: fixed.
- Increase of Diffie-Hellman requested modulus size caused an error (QC3 is limited to 2048 bits). Fixed by reducing these for os400qc3.
- Failing Diffie-Hellman key pair generation now effectively returns an error.
- Implement an ASCII-compatible assert.h header file.
- Implement RSA2 in os400qc3 backend.

Thanks for considering this.

### DIFF
--- a/docs/HACKING-CRYPTO
+++ b/docs/HACKING-CRYPTO
@@ -406,6 +406,21 @@ TripleDES-CBC algorithm identifier initializer.
 
 5) Diffie-Hellman support.
 
+LIBSSH2_DH_GEX_MINGROUP
+The minimum Diffie-Hellman group length in bits supported by the backend.
+Usually defined as 2048.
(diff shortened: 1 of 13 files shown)


monnerat:
Thanks to you for merging, @vszakats :-)

PR #964, opened 2023-04-14
checksrc: fix `EQUALSNULL` warnings

vszakats:
`s/([a-z0-9._>*-]+) == NULL/!\1/g`

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -83,7 +83,6 @@ checksrc:
 	  -ASNPRINTF \
 	  -ACOPYRIGHT \
 	  -AFOPENMODE \
-	  -AEQUALSNULL \
 	  -ATYPEDEFSTRUCT \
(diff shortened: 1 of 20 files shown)


vszakats:
src\session.c(914): error C2220: the following warning is treated as an error
src\session.c(914): warning C4706: assignment within conditional expression
src\session.c(925): warning C4706: assignment within conditional expression
src\session.c(1111): warning C4706: assignment within conditional expression
src\pem.c(494): error C2220: the following warning is treated as an error
src\pem.c(494): warning C4706: assignment within conditional expression
src\pem.c(145): error C2220: the following warning is treated as an error
src\pem.c(145): warning C4706: assignment within conditional expression
src\kex.c(3330): error C2220: the following warning is treated as an error
src\kex.c(3330): warning C4706: assignment within conditional expression

PR #971, opened 2023-04-15
tests: switch Dockerfile to debian:testing-slim

vszakats:
From debian:bullseye

- doesn't need manual bumps.
- is ahead of stable and should be stable enough for our purpose.
- slim is saving resources.

Closes #971


vszakats:
It seems this cannot be CI tested as the tests assume the update already being in master.

PR #975, opened 2023-04-16
build: add `-Wbad-function-cast` picky warning

vszakats:
Also adjust minimum gcc versions in comment.

Closes #975


vszakats:
Missed extra info from the commit message: This option was already used by gcc with autotools. This patch adds it to clang with autotools and both clang and gcc with CMake.

PR #976, opened 2023-04-16
try debian:stable

vszakats:


vszakats:
Couldn't figure out how to make latest Debian image work, closing.

vszakats:
Switched to `bullseye-slim` as a result. Though `stable-slim` or `testing-slim` would be ideal.

PR #979, opened 2023-04-17
WinCNG test failures: do more detailed logs

vszakats:


mback2k:
`hasAlgDHwithKDF` is actually the more interesting variable to log, because the raw KDF support is the important part here.

vszakats:
In this test (no. 24): https://ci.appveyor.com/project/libssh2org/libssh2/builds/46833132/job/8d8jm0893ntwnvk8

called from `kex.c` / `diffie_hellman_sha_algo()` in this order:
_libssh2_dh_key_pair() → _libssh2_wincng.hasAlgDHwithKDF=0
_libssh2_dh_secret() → status=0, _libssh2_wincng.hasAlgDHwithKDF=1

vszakats:
@mback2k: According to the logs, `hasAlgDHwithKDF` is 0 while `_libssh2_dh_key_pair()` is running, even on Server 2019. This value is set in a subsequent call in `libssh2_dh_secret()`, to 1, which is the correct value for this OS, AFAIU. But, by then it's too late to help generating the key pair.

I force bumped this value to 1 in `_libssh2_dh_key_pair()`, and it seems to make Server 2019 tests go green. And even Server 2016.

Do you think `hasAlgDHwithKDF=0` is correct in `_libssh2_dh_key_pair()`?

mback2k:
`_libssh2_dh_key_pair()` and `libssh2_dh_secret()` will try to work with the new API as long as the value is not -1, see: https://github.com/libssh2/libssh2/blob/fba0b52b6a8f719b5260530c3f220a653a19f33f/src/wincng.c#L2407
and:
https://github.com/libssh2/libssh2/blob/fba0b52b6a8f719b5260530c3f220a653a19f33f/src/wincng.c#L2592

So, the problematic mod_exp function will only be called at the end of both functions if the KDF is really not supported and the value is set to -1 here:
https://github.com/libssh2/libssh2/blob/fba0b52b6a8f719b5260530c3f220a653a19f33f/src/wincng.c#L2704-L2707

The only thing that will be different between the first and any subsequent calls of `_libssh2_dh_key_pair()`  depending on the determination state of KDF support, is the extraction of just the public or also the private key from the bignum handle:
https://github.com/libssh2/libssh2/blob/fba0b52b6a8f719b5260530c3f220a653a19f33f/src/wincng.c#L2473-L2484

vszakats:
Thanks Marc. This means those green runs were just mere luck.

What else may be causing these errors, even on Windows 10?

mback2k:
Not sure. If the error also appears on Windows 8 or newer I guess my assumption about the potential root cause was wrong.

One more thing for pre Windows 8: we could make it always generate the DH key using the new API if available even if the KDF is not supported by never setting the value to -1 and just keeping it on 0. Then the bignum_rand function would never be called.

vszakats:
Let's try that in 72cbf455908bc5ed6cdcf6f3ec2a0e1937b6844a.

vszakats:
2019/2016 passed (of one try each), 2012 R2 did not: https://ci.appveyor.com/project/libssh2org/libssh2/builds/46855259

mback2k:
But the change shouldn't influence the behavior of 2016 or 2019, right?

I guess I am out of ideas now and my root cause assumption seems to have been wrong all along, right?

vszakats:
It shouldn't if the code detects the feature correctly before it affects logic. Your theory sounds plausible, it's also possible my tests are wrong. Maybe we should bisect it, I'm not sure.

vszakats:
It shouldn't if the code detects the feature correctly before it affects logic. Your theory sounds plausible, it's also possible my tests are wrong. Maybe we should bisect it, I'm not sure.

vszakats:
Putting this aside. Might merge the kex error message patch in a separate commit.

PR #989, opened 2023-04-20
ci: test batch -> shell mod

vszakats:


mback2k:
There was a reason I stayed with `.bat` and `.ps1` for the respective steps ;)

vszakats:
I had to give it a try! I think it's possible to make it work, but run out of patience today.

PR #992, opened 2023-04-20
tests: add AES-GCM read tests

dfandrich:


tests/CMakeLists.txt
```diff
@@ -127,14 +127,17 @@ foreach(test
...
   aes192-cbc
   aes192-ctr
   aes256-cbc
   aes256-ctr
+  aes256-gcm@openssh.com
   rijndael-cbc@lysator.liu.se
 )
-  add_test(NAME test_${test} COMMAND "$<TARGET_FILE:test_read>")
-  set_tests_properties(test_${test} PROPERTIES ENVIRONMENT "FIXTURE_TEST_CRYPT=${test}")
+  string(REGEX REPLACE "@.*" "" testfn "${test}")
```
vszakats, line 138: '@' should be fine in filenames.

dfandrich: Yes, but it's ugly. I also wanted to keep the CMake test name and the automake test name the same.

vszakats: The internal CMake has nothing to do with the filename or automake. It's not using any external files.

vszakats:
What is the point in syncing internal CMake identifiers with files that we're not using for anything?

tests/CMakeLists.txt
dfandrich: The internal CMake name *does* have something to do with the filename *and* automake. I want the test names to be the same no matter how they're built—cmake or automake. This makes them the same.

dfandrich: I want somebody to be able to say "Test test_aes256-gcm is failing" and have that refer to a single test, no matter how it was built. These are not internal identifiers—these are the test names that CMake shows to users.

vszakats: The native, as-is, mac/crypt names are the best for this IMO.

PR #995, opened 2023-04-21
ci: reduce algo test runtime on AppVeyor

vszakats:
Make the block count customizable in `test_read` via environment
`FIXTURE_XFER_COUNT`.

Set the custom count lower than the default when running on AppVeyor.

The goal is to reduce CI roundtrip times.

Closes #995

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,7 @@
 
 environment:
   CONFIGURATION: "Release"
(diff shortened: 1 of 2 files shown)


dfandrich:
Maybe make the variables unsigned since negative numbers make no sense there.

vszakats:
I went with those first, then re-aligned with existing signed variables. I'll do that though, it'll be indeed nicer.

PR #198, opened 2017-07-11
A collection of small fixes

tiennou:
This is a handful of small test-harness tweaks I collected while trying to have Xcode run the test cases (for the record, you need to set your PATH env-var correctly so Xcode finds `docker`, for each separate test case you need to debug).

The second commit moves our headers first in the list, so Xcode wouldn't try to use the ones from "system" locations (ie. Homebrew's `/usr/local`).

The last one hides OpenSSL's AES-CTR thingy inside its own initialization so it doesn't force other backends to `NOP` it.

PR #1003, opened 2023-04-25
tidy-up: drop no-op use of `getcwd` and `sys/param.h`

vszakats:
If not `FIXTURE_WORKDIR` envvar of compile-time macro is present, just don't do anything. Before this patch, we retrieved the current directory, then changed current directory to it, which is pointless.

This also makes `sys/param.h` unnecessary to detect and use, and we can drop an MSVC-specific hack.

Delete an extra idle use of the above header.

Still remains the two methods of setting `FIXTURE_WORKDIR`, of which only the compile-time method is actually used and only with CMake, for an Xcode workaround.

Ref: https://github.com/libssh2/libssh2/pull/198/commits/10a5cbf945abcc60153ee3d59284d09fc64ea152
Ref: 9f8e69cca0b94266c174ae13a860a02c762a3f4a

Closes #1003


vszakats:
This one is a deeper rabbit-hole. Throughout the times, up to recently there were various patches fixing the current working directory:

4c9ed51f962f542b98789b15bedaaa427f4029a2 2023 autotools + out-of-tree build tests
10a5cbf945abcc60153ee3d59284d09fc64ea152 2018 CMake + Xcode
13f8addd1bc17e6c55d52491cc6304319ac38c6d 2015 autotools + distcheck

vszakats:
Closing in favour of #1009.

PR #752, opened 2022-10-04
Agent forwarding implementation

MichaelBuckley:
This PR contains a series of patches that date back many years and I believe were discussed on the mailing list, but never merged. We have been using these in our local copy of libssh2 without issue since 2015, if not earlier. I believe this is the full set of changes, as we tried to use comments to mark where our copy of libssh2 differs from the canonical version.

This also contains changes I made earlier this year, but which were not discussed on the mailing list, to support certificates and FIDO2 keys with agent forwarding.

Note that this is not a complete implementation of agent forwarding, as that is outside the scope of libssh2. Clients still need to provide their own implementation that parses ssh-agent methods after calling libssh2_channel_read() and calls the appropriate callback messages in libssh2. See the man page changes in this PR for more details.

PR #1004, opened 2023-04-25
Couple of small fixes

migueldeicaza:
The first merely makes sure that we include the proper header file, otherwise this fails to compile with recent versions of Xcode.

The second patch fixes an issue that would corrupt the data stream when attempting to initialize the agent in non-blocking mode, as it is necessary to propagate the EAGAIN signal upstream when the transport returns EAGAIN.

### DIFF
--- a/src/channel.c
+++ b/src/channel.c
@@ -1165,7 +1165,7 @@ static int channel_request_auth_agent(LIBSSH2_CHANNEL *channel,
         if(rc == LIBSSH2_ERROR_EAGAIN) {
             _libssh2_error(session, rc,
                            "Would block sending auth-agent request");
-	    return rc;
+            return rc;
         }
         else if(rc) {
             channel->req_auth_agent_state = libssh2_NB_state_idle;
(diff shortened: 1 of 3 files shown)


src/channel.c
```diff
@@ -1165,6 +1165,7 @@ static int channel_request_auth_agent(LIBSSH2_CHANNEL *channel,
         if(rc == LIBSSH2_ERROR_EAGAIN) {
             _libssh2_error(session, rc,
                            "Would block sending auth-agent request");
+	    return rc;
```
vszakats, line 1168: This needs spaces for indent.

vszakats:
Including `openssl/err.h` makes sense, but do you have an idea what does it have to do with Xcode? What is the error you're seeing?

src/channel.c
migueldeicaza: Good call, I think my editor was going full tabs.

vszakats:
`packet.c` patch: regression from #752.

migueldeicaza:
Without it, a file mater on references an ERR_ symbol which has not been declared, and it becomes an error to call due to a warning-as-error for using an undeclared function.

vszakats:
Got it, so that is a regression from #789. It seems that OpenSSL builds with `OPENSSL_NO_ENGINE` could have fallen into this after that patch.

PR #1015, opened 2023-04-28
ci: disable WinCNG tests due to flakiness

vszakats:


mback2k:
Not a fan as I would rather like to find the root cause, I am out of ideas at this point.

vszakats:
Not a good solution for sure, but having almost everything red is also confusing. For every run one needs to inspect what happened, instead of trusting the red/green mark.

Maybe automatically re-running failed WinCNG tests could work also? But that would make AppVeyor CI runs slower.
Another workaround is to let the tests run, but ignore the result.

vszakats:
Testing the retry approach here: #1017

vszakats:
Closing this in favour of #1017.

PR #1019, opened 2023-05-01
sftp: parse attribute extensions, if present, to avoid stream parsing errors

nicryptolaus:
This change prevents directory listing errors when attribute extensions are present by advancing stream parsing past extensions.

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -725,6 +725,27 @@ sftp_bin2attr(LIBSSH2_SFTP_ATTRIBUTES *attrs, const unsigned char *p,
         attrs->mtime = mtime;
     }
 
+    /* Parse extended data, if present, to avoid stream parsing errors */
+    if(attrs->flags & LIBSSH2_SFTP_ATTR_EXTENDED) {
+        uint32_t extended_count;
+        uint32_t i;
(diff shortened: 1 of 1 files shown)


src/sftp.c
```diff
@@ -725,6 +725,27 @@ sftp_bin2attr(LIBSSH2_SFTP_ATTRIBUTES *attrs, const unsigned char *p,
...
+        unsigned char *edata;
+
+        if(_libssh2_get_u32(&buf, &extended_count) != 0) {
+            return LIBSSH2_ERROR_BUFFER_TOO_SMALL;
+        }
+
+        for(i = 0; i < extended_count; ++i) {
+            if(_libssh2_get_string(&buf, &etype, &etype_len) != 0 ||
+               _libssh2_get_string(&buf, &edata, &edata_len) != 0) {
+                    return LIBSSH2_ERROR_BUFFER_TOO_SMALL;
```
vszakats, line 744: `return` has one level too many indention.

Also, I think we can drop ` != 0`s from the `if()`s.

Otherwise LGTM.

vszakats:
Merged, thank you @nicryptolaus!

PR #1026, opened 2023-05-04
cmake: assume `bcrypt.h` with WinCNG

vszakats:
autotools already didn't check for `bcrypt.h`, and such check is only
required for old/legacy mingw without obsolete/incomplete Windows
headers.

curl deprecated old-mingw support just recently and will delete support
in September 2023.

This patch saves some complexity and detection time by dropping this
check for CMake. Meaning that mingw-w64 is now required to compile
libssh2 when using the WinCNG backend for 32-bit builds. Other backends
and CPU platforms are not affected.

Ref: https://github.com/curl/curl/commit/e4d5685cb5d6eb07e1b43156fd7e3ba3563afba5

Closes #1026

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -363,21 +363,12 @@ endif()
 # Detect platform-specific crypto-backends last:
 
 if(CRYPTO_BACKEND STREQUAL "WinCNG" OR NOT CRYPTO_BACKEND)
+  set(CRYPTO_BACKEND "WinCNG")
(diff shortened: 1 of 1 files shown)


vszakats:
Making CMake detection on Windows a bit more faster. Ref: https://github.com/libssh2/libssh2/pull/732

neheb (approved):

neheb:
Makes sense. In the meson port this part made no sense to me.

PR #1027, opened 2023-05-04
cmake: do not check for `poll()` on Windows

vszakats:
While it seems to exist on mingw in theory, it's not detected as of this
writing. It also has issues, and not ready for production use:
https://stackoverflow.com/questions/1671827/poll-c-function-on-windows

On MSVC it's even less supported.

Skip checking this to save CMake detection time.

Closes #1027


vszakats:
Making CMake detection on Windows a bit more faster 2. Ref: https://github.com/libssh2/libssh2/pull/732

PR #1030, opened 2023-05-04
ci: add mingw-w64 cmake jobs

vszakats:
Closing #1030


vszakats:
Merged via individual commits.

PR #1036, opened 2023-05-07
cmake: try shared/static lib updates

vszakats:
- cmake: enable shared libssh2 library by default.

- cmake: add `IMPORT_LIB_SUFFIX` (like `STATIC_LIB_SUFFIX`).

- cmake: tweak static/import lib name collision avoidance logic.

- ci: drop default cmake shared/static configuration options.

- ci: do not disable shared lib with msys2/autotools in GHA.

- Makefile.mk: fix `DYN=1 test` by skipping tests needing static lib.


vszakats:
Merged via individual commits.

PR #1038, opened 2023-05-09
examples to upload/download directory

lampmanyao:


vszakats:
To avoid portability issues with `mkdir`/`chdir` I'd suggest not to use them, and download to the current directory instead.

This also needs to be fixed:
C:\projects\libssh2\example\sftpdir_download.c(118,28): warning C4244: 'function': conversion from 'ssize_t' to 'unsigned int', possible loss of data

`write()` is non-portable. This existing hack can be reused here:
https://github.com/libssh2/libssh2/blob/c7074ccc9d47d4dcc07ada9c68e56342af833318/example/scp.c#L8-L10

Bigger issue is that `dirent.h`/`opendir`/`readdir` are also non-portable. To avoid them, maybe this should pick filenames directly from the command-line and upload those in batch.

vszakats:
Without addressing the portability issues, we cannot merge this. Possibly a solution integrated into existing upload/download examples to support multiple files while avoiding directory operation would be simpler to implement while also providing a useful libssh2 API example.

Issue #1020, opened 2023-05-02
wolfSSL: libssh2_session_handshake failed: -12/-44 DECRYPT/ENCRYPT


vszakats:
**Describe the bug**
When running `tests/test_ssh2`, either of these errors happen:

- libssh2_session_handshake failed (-12) `LIBSSH2_ERROR_DECRYPT`: Failed to get response to ssh-userauth request
  https://github.com/libssh2/libssh2/actions/runs/4851558817/jobs/8645524071#step:7:1087

- libssh2_session_handshake failed (-44) `LIBSSH2_ERROR_ENCRYPT`: Unable to ask for ssh-userauth service
  https://github.com/libssh2/libssh2/actions/runs/4847366472/jobs/8637553577#step:7:1052
  https://github.com/libssh2/libssh2/actions/runs/4846389013/jobs/8635903998#step:7:96

This is the first packet with encryption.

**To Reproduce**
Run `tests/test_ssh2` against an sshd server.

**Expected behavior**
Same result as with OpenSSL/LibreSSL backends:
https://github.com/libssh2/libssh2/actions/runs/4847366472/jobs/8637553403#step:7:72
https://github.com/libssh2/libssh2/actions/runs/4847366472/jobs/8637553480#step:7:72
https://github.com/libssh2/libssh2/actions/runs/4851558817/jobs/8645523936#step:7:71

**Version:**
 - OS and version: macOS (possibly others)
 - libssh2 version: 1.10.1-DEV
 - crypto backend: wolfSSL v5.6.0

vszakats:
Reverting the AES-GCM patch #797 fixes it: 3c953c05d67eb1ebcfd3316f279f12c4b1d600b4

The issue might be inside `_libssh2_cipher_crypt()`. Possibly with the interpretation of the return value of `EVP_Cipher()`, or `EVP_CIPHER_CTX_ctrl()`.

@dfandrich Can you take a look?

The issue happens with both AES-GCM disabled (as in master) or enabled for wolfSSL via this patch:
```diff
--- a/src/openssl.h
+++ b/src/openssl.h
@@ -160,7 +160,8 @@
 # define LIBSSH2_AES_CBC 0
 #endif
 
-#if (OPENSSL_VERSION_NUMBER >= 0x01010100fL && !defined(OPENSSL_NO_AES))
+#if (OPENSSL_VERSION_NUMBER >= 0x01010100fL && !defined(OPENSSL_NO_AES)) || \
[...]

PR #1045, opened 2023-05-25
wolfssl: fix detection of AES-GCM feature

vszakats:
Follow-up to df513c0128e1a811ad863d153892618e728845f0

Ref: https://github.com/libssh2/libssh2/issues/1020#issuecomment-1562069241

Closes #1045

### DIFF
--- a/src/openssl.h
+++ b/src/openssl.h
@@ -161,7 +161,8 @@
 #endif
 
 #if (OPENSSL_VERSION_NUMBER >= 0x01010100fL && !defined(OPENSSL_NO_AES)) || \
-    (defined(LIBSSH2_WOLFSSL) && defined(WOLFSSL_AESGCM_STREAM))
+    (defined(LIBSSH2_WOLFSSL) && \
+    defined(HAVE_AESGCM) && defined(WOLFSSL_AESGCM_STREAM))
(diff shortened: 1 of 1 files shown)


src/openssl.h
```diff
@@ -161,7 +161,7 @@
 #endif
 
 #if (OPENSSL_VERSION_NUMBER >= 0x01010100fL && !defined(OPENSSL_NO_AES)) || \
-    (defined(LIBSSH2_WOLFSSL) && defined(HAVE_AESGCM))
+    (defined(LIBSSH2_WOLFSSL) && defined(WOLFSSL_AESGCM_STREAM))
```
kareem-wolfssl, line 164: While our AES-GCM stream support should depend on having AES-GCM in general enabled, I'd recommend checking for both here to be safe, ie:
`(defined(HAVE_AESGCM) && defined(WOLFSSL_AESGCM_STREAM))`

kareem-wolfssl (changes requested):

src/openssl.h
vszakats: Restored the previous check and now checking for both.

kareem-wolfssl (approved):

PR #1058, opened 2023-05-30
Handle EINTR from send/recv/poll/select to try again as the error is not fatal

### LINKED ISSUE #955: System calls interrupted by profilers (such as poll) cause EINTR errors that break libssh2
**Describe the bug**

Hey 👋! Thanks for libssh2 :tada: 

I work on [Datadog's Ruby profiler](https://github.com/datadog/dd-trace-rb/) and I arrived here while investigating a customer issue where they were using the [rugged library](https://github.com/libgit2/rugged) which uses [libgit2](https://github.com/libgit2/libgit2) which uses libssh2, and it broke [in combination with our profiler](https://github.com/DataDog/dd-trace-rb/issues/2721).

Specifically, what happened is that our profiler uses unix signals (specifically `SIGPROF`) to interrupt the application and then gather information on what it was doing.

This has a downside: some system calls get interrupted by the signal handler ([as documented on the man page](https://man7.org/linux/man-pages/man7/signal.7.html#:~:text=Interruption%20of%20system%20calls%20and%20library%20functions%20by%20signal%20handlers)) and return `EINTR` to tell the caller this has happened and that it should try again.

Unfortunately, libssh2 does not specifically handle `EINTR` errors -- it treats them as a real error, and fails connections when it happens, returning an error to the caller.

Note that this approach of using signals is not unique to our profiler:
* https://github.com/libssh2/libssh2/issues/591 is an earlier report of this issue [which turns out to be caused by the built-in Dart profiler](https://github.com/flutter/flutter/issues/82488)
* The [stackprof](https://github.com/tmm1/stackprof) Ruby profiler uses signals as well
* The Java/JVM [async-profiler](https://github.com/async-profiler/async-profiler) also uses signals
* (...etc)

(Funnily enough, there's an old essay on unix issues written in 1991 that complains exactly about this -- [link](https://www.dreamsongs.com/RiseOfWorseIsBetter.html))

**To Reproduce**
I have modified `ssh2.c` to simulate having a profiler running to trigger the issue:

```diff
$ git diff
diff --git a/example/ssh2.c b/example/ssh2.c
index a762d70..a96af2e 100644
--- a/example/ssh2.c
+++ b/example/ssh2.c
@@ -35,6 +35,27 @@
 #include <stdlib.h>
 #include <ctype.h>
 
+#include <pthread.h>
+#include <signal.h>
+#include <errno.h>
+
+pthread_t main_thread;
+
+static void *background_thread(void *unused) {
+    fprintf(stderr, "Hello from background thread!\n");
+
+    while (1) {
+        pthread_kill(main_thread, SIGPROF);
+        usleep(1000);
+    }
+
+    return NULL;
+}
+
+static void signal_handler_function(int foo, siginfo_t *bar, void *baz) {
+    //fprintf(stderr, "Hello from signal handler!\n");
+}
+
 #if defined(_MSC_VER) && _MSC_VER < 1900
 #define snprintf _snprintf
 #endif
@@ -73,6 +94,22 @@ int main(int argc, char *argv[])
     char *userauthlist;
     LIBSSH2_SESSION *session;
     LIBSSH2_CHANNEL *channel;
+    int error_code;
+    pthread_t thread;
+
+    struct sigaction signal_handler_config = {
+      .sa_flags = SA_RESTART | SA_SIGINFO,
+      .sa_sigaction = signal_handler_function
+    };
+    sigemptyset(&signal_handler_config.sa_mask);
+
+    if (sigaction(SIGPROF, &signal_handler_config, NULL) != 0) {
+        fprintf(stderr, "Could not install signal handler\n");
+    }
+
+    main_thread = pthread_self();
+
+    pthread_create(&thread, NULL, background_thread, NULL);
 
 #ifdef WIN32
     WSADATA wsadata;
@@ -137,8 +174,8 @@ int main(int argc, char *argv[])
         LIBSSH2_TRACE_SOCKET
     );
 
-    if(libssh2_session_handshake(session, sock)) {
-        fprintf(stderr, "Failure establishing SSH session\n");
+    if(error_code = libssh2_session_handshake(session, sock)) {
+        fprintf(stderr, "Failure establishing SSH session (error code %d)\n", error_code);
         return -1;
     }
 
diff --git a/src/session.c b/src/session.c
index 5d2b023..8a8fc8d 100644
--- a/src/session.c
+++ b/src/session.c
@@ -710,6 +710,7 @@ int _libssh2_wait_socket(LIBSSH2_SESSION *session, time_t start_time)
                               "Timed out waiting on socket");
     }
     if(rc < 0) {
+        perror("Poll failed");
         return _libssh2_error(session, LIBSSH2_ERROR_TIMEOUT,
                               "Error waiting on socket");
     }
```

And here's what I see:

```
$ example/example-ssh2 192.168.0.3 user password sleep 10
Hello from background thread!
Connecting to 192.168.0.3:22 as user user
Poll failed: Interrupted system call
Failure establishing SSH session (error code -9)
```

and here's the relevant excerpt from running strace as well:

```
sendto(3, "SSH-2.0-libssh2_1.10.1_DEV\r\n", 28, MSG_NOSIGNAL, NULL, 0) = 28
recvfrom(3, 0x7ffe3f6ec627, 1, MSG_NOSIGNAL, NULL, NULL) = -1 EAGAIN (Resource temporarily unavailable)
poll([{fd=3, events=POLLIN}], 1, -1)    = ? ERESTART_RESTARTBLOCK (Interrupted by signal)
--- SIGPROF {si_signo=SIGPROF, si_code=SI_TKILL, si_pid=358496, si_uid=1000} ---
rt_sigreturn({mask=[]})                 = -1 EINTR (Interrupted system call)
write(2, "Poll failed: Interrupted system "..., 37Poll failed: Interrupted system call
) = 37
write(2, "Failure establishing SSH session"..., 49Failure establishing SSH session (error code -9)
) = 49
exit_group(-1)                          = ?
+++ exited with 255 +++
```

**Expected behavior**

When `poll` and similar system calls are interrupted with an `EINTR` error, libssh2 should retry the call, instead of considering it as a real error.

 - OS and version: Linux, Ubuntu 22.04.2 LTS
 - libssh2 version: master
 - crypto backend: openssl

**Additional context**

N/A

sodabrew:
Fixes #955 

Would it make sense to handle EINTR in the wrapper macros `BLOCK_ADJUST` and `BLOCK_ADJUST_ERRNO` instead?

**macOS**:
`man select`:
> select() returns with an error, including one due to an interrupted call, the descriptor sets will be unmodified and the global variable errno will be set to indicate the error.

`man poll`:
> If poll() returns with an error, including one due to an interrupted call, the fds array will be unmodified and the global variable errno will be set to indicate the error.

`man recv`:
> [EINTR]            The receive was interrupted by delivery of a signal before any data were available.

`man send`:
> [EINTR]            A signal interrupts the system call before any data is transmitted.

**Linux**:
`man select`:
> EINTR The function was interrupted before any of the selected events occurred and before the timeout interval expired.

`man poll`:
> EINTR A signal was caught during poll().
Not super helpful, but the text implies behavior as if no other actions were taken.

`man recv`:
> EINTR The receive was interrupted by delivery of a signal before any data were available
`man send`:
>EINTR A signal interrupted send() before any data was transmitted.

**FreeBSD**:
Does not mention EINTR at all in `send` but for `recv`, `poll`, `select` the text follow as macOS and Linux.
https://man.freebsd.org/cgi/man.cgi?sektion=2&query=send

**Windows**:
It looks like `WSAEINTR` should be handled identically as above, ~however that is NOT implemented in this PR (yet! feedback appreciated!)~ Update: added Windows handling, but did not adapt a test for it.
: https://learn.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -133,7 +133,7 @@ int _libssh2_error(LIBSSH2_SESSION* session, int errcode, const char *errmsg)
 }
 
 #ifdef _WIN32
-static int wsa2errno(void)
+int _libssh2_wsa2errno(void)
 {
     switch(WSAGetLastError()) {
     case WSAEWOULDBLOCK:
@@ -167,23 +167,29 @@ _libssh2_recv(libssh2_socket_t sock, void *buffer, size_t length,
     (void)abstract;
 
     rc = recv(sock, buffer, length, flags);
-#ifdef _WIN32
-    if(rc < 0)
-        return -wsa2errno();
-#else
     if(rc < 0) {
+        int err;
+#ifdef _WIN32
+        err = _libssh2_wsa2errno();
+#else
+        err = errno;
+#endif
+        /* Profiling tools that use SIGPROF can cause EINTR responses.
+           recv() does not modify its arguments when it returns EINTR,
+           but there may be data waiting, so the caller should try again */
+        if(err == EINTR)
+            return -EAGAIN;
         /* Sometimes the first recv() function call sets errno to ENOENT on
            Solaris and HP-UX */
-        if(errno == ENOENT)
+        if(err == ENOENT)
             return -EAGAIN;
 #ifdef EWOULDBLOCK /* For VMS and other special unixes */
-        else if(errno == EWOULDBLOCK)
-          return -EAGAIN;
+        else if(err == EWOULDBLOCK)
+            return -EAGAIN;
 #endif
         else
-            return -errno;
+            return -err;
     }
-#endif
     return rc;
 }
 
@@ -200,18 +206,24 @@ _libssh2_send(libssh2_socket_t sock, const void *buffer, size_t length,
     (void)abstract;
 
     rc = send(sock, buffer, length, flags);
-#ifdef _WIN32
-    if(rc < 0)
-        return -wsa2errno();
-#else
     if(rc < 0) {
+        int err;
+#ifdef _WIN32
+        err = _libssh2_wsa2errno();
+#else
+        err = errno;
+#endif
+        /* Profiling tools that use SIGPROF can cause EINTR responses.
+           send() is defined as not yet sending any data when it returns EINTR,
+           so the caller should try again */
+        if(err == EINTR)
+            return -EAGAIN;
 #ifdef EWOULDBLOCK /* For VMS and other special unixes */
-        if(errno == EWOULDBLOCK)
+        if(err == EWOULDBLOCK)
             return -EAGAIN;
 #endif
-        return -errno;
+        return -err;
     }
-#endif
     return rc;
 }
 
--- a/src/misc.h
+++ b/src/misc.h
@@ -79,7 +79,7 @@ int _libssh2_error_flags(LIBSSH2_SESSION* session, int errcode,
                          const char *errmsg, int errflags);
 int _libssh2_error(LIBSSH2_SESSION* session, int errcode, const char *errmsg);
 
-#ifdef WIN32
+#ifdef _WIN32
 /* Convert Win32 WSAGetLastError to errno equivalent */
 int _libssh2_wsa2errno(void);
 #endif
--- a/src/misc.h
+++ b/src/misc.h
@@ -79,6 +79,11 @@ int _libssh2_error_flags(LIBSSH2_SESSION* session, int errcode,
                          const char *errmsg, int errflags);
 int _libssh2_error(LIBSSH2_SESSION* session, int errcode, const char *errmsg);
 
+#ifdef WIN32
+/* Convert Win32 WSAGetLastError to errno equivalent */
+int _libssh2_wsa2errno(void);
+#endif
+
 void _libssh2_list_init(struct list_head *head);
 
 /* add a node last in the list */
--- a/src/session.c
+++ b/src/session.c
@@ -674,7 +674,7 @@ int _libssh2_wait_socket(LIBSSH2_SESSION *session, time_t start_time)
     }
     if(rc < 0) {
         int err;
-#ifdef WIN32
+#ifdef _WIN32
         err = _libssh2_wsa2errno();
 #else
         err = errno;
--- a/src/session.c
+++ b/src/session.c
@@ -673,6 +673,18 @@ int _libssh2_wait_socket(LIBSSH2_SESSION *session, time_t start_time)
                               "Timed out waiting on socket");
     }
     if(rc < 0) {
+        int err;
+#ifdef WIN32
+        err = _libssh2_wsa2errno();
+#else
+        err = errno;
+#endif
+        /* Profiling tools that use SIGPROF can cause EINTR responses.
+           poll() / select() do not set any descriptor states on EINTR,
+           but some fds may be ready, so the caller should try again */
+        if(err == EINTR)
+            return 0;
+
         return _libssh2_error(session, LIBSSH2_ERROR_TIMEOUT,
                               "Error waiting on socket");
     }
--- a/tests/test_ssh2.c
+++ b/tests/test_ssh2.c
@@ -24,6 +24,11 @@
 #include <stdio.h>
 #include <stdlib.h>  /* for getenv() */
 
+#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
+#include <pthread.h>
+#include <signal.h>
+#endif
+
 static const char *hostname = "127.0.0.1";
 static const int port_number = 4711;
 static const char *pubkey = "key_rsa.pub";
@@ -39,6 +44,30 @@ static void portable_sleep(unsigned int seconds)
     sleep(seconds);
 #endif
 }
+
+#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
+static pthread_t main_thread;
+
+static void *background_thread(void *unused)
+{
+    (void)unused;
+    fprintf(stderr, "Profiling thread sending SIGPROF every 1ms!\n");
+
+    do {
+        pthread_kill(main_thread, SIGPROF);
+        usleep(1000);
+    } while(1);
+
+    return NULL;
+}
+
+static void signal_handler_function(int a, siginfo_t *b, void *c)
+{
+    (void)a;
+    (void)b;
+    (void)c;
+}
+#endif
 
 int main(int argc, char *argv[])
 {
@@ -61,6 +90,24 @@ int main(int argc, char *argv[])
         fprintf(stderr, "WSAStartup failed with error: %d\n", rc);
         return 1;
     }
+#endif
+
+#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
+    pthread_t thread;
+
+    struct sigaction signal_handler_config = {
+        .sa_flags = SA_RESTART | SA_SIGINFO,
+        .sa_sigaction = signal_handler_function
+    };
+    sigemptyset(&signal_handler_config.sa_mask);
+
+    if(sigaction(SIGPROF, &signal_handler_config, NULL) != 0) {
+        fprintf(stderr, "Could not install signal handler\n");
+    }
+
+    main_thread = pthread_self();
+
+    pthread_create(&thread, NULL, background_thread, NULL);
 #endif
 
     (void)argc;
@@ -160,7 +207,7 @@ int main(int argc, char *argv[])
                rc != LIBSSH2_ERROR_KEY_EXCHANGE_FAILURE ||
 #endif
                ++retry > retries) {
-                break;
+                goto shutdown;
             }
             fprintf(stderr, "Retrying... %d / %d\n", retry, retries);
         } while(1);
--- a/tests/test_ssh2.c
+++ b/tests/test_ssh2.c
@@ -24,9 +24,11 @@
 #include <stdio.h>
 #include <stdlib.h>  /* for getenv() */
 
-#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
+#if defined(HAVE_PTHREAD_H)
 #include <pthread.h>
 #include <signal.h>
+
+static pthread_t main_thread;
 #endif
 
 static const char *hostname = "127.0.0.1";
@@ -45,13 +47,11 @@ static void portable_sleep(unsigned int seconds)
 #endif
 }
 
-#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
-static pthread_t main_thread;
-
+#if defined(HAVE_PTHREAD_H)
 static void *background_thread(void *unused)
 {
     (void)unused;
-    fprintf(stderr, "Profiling thread sending SIGPROF every 1ms!\n");
+    fprintf(stderr, "Profiling thread sending SIGPROF every 1ms.\n");
 
     do {
         pthread_kill(main_thread, SIGPROF);
@@ -92,22 +92,24 @@ int main(int argc, char *argv[])
     }
 #endif
 
-#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
-    pthread_t thread;
+#if defined(HAVE_PTHREAD_H)
+    {
+        pthread_t thread;
 
-    struct sigaction signal_handler_config = {
-        .sa_flags = SA_RESTART | SA_SIGINFO,
-        .sa_sigaction = signal_handler_function
-    };
-    sigemptyset(&signal_handler_config.sa_mask);
+        struct sigaction signal_handler_config = {
+            .sa_flags = SA_RESTART | SA_SIGINFO,
+            .sa_sigaction = signal_handler_function
+        };
+        sigemptyset(&signal_handler_config.sa_mask);
 
-    if(sigaction(SIGPROF, &signal_handler_config, NULL) != 0) {
-        fprintf(stderr, "Could not install signal handler\n");
+        if(sigaction(SIGPROF, &signal_handler_config, NULL) != 0) {
+            fprintf(stderr, "Could not install signal handler.\n");
+        }
+
+        main_thread = pthread_self();
+
+        pthread_create(&thread, NULL, background_thread, NULL);
     }
-
-    main_thread = pthread_self();
-
-    pthread_create(&thread, NULL, background_thread, NULL);
 #endif
 
     (void)argc;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,8 +212,7 @@ if(NOT WIN32)
   check_include_files("sys/un.h" HAVE_SYS_UN_H)
   check_include_files("arpa/inet.h" HAVE_ARPA_INET_H)  # example and tests
   check_include_files("netinet/in.h" HAVE_NETINET_IN_H)  # example and tests
-  check_include_files(pthread.h HAVE_PTHREAD_H)  # tests
-  check_include_files(signal.h HAVE_SIGNAL_H) # tests
+  check_include_files("pthread.h" HAVE_PTHREAD_H)  # tests
 endif()
 
 # CMake uses C syntax in check_symbol_exists() that generates a warning with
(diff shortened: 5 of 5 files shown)


vszakats:
checksrc needs fixing indentation and long comments (for all 3 hunks):
```
./src/misc.c:167:86: warning: Longer than 79 columns (LONGLINE)
         /* Profiling tools that use SIGPROF can cause EINTR responses. For recv, this
./src/misc.c:168:96: warning: Longer than 79 columns (LONGLINE)
          * is defined as not returning any data, but there may be data waiting so call again */
./src/misc.c:170:11: warning: not indented 4 steps (uses 2) (INDENTATION)
           return 0;
           ^
```

sodabrew:
I fixed the indentation and added support for Windows WSA errors in src/session.c, please let me know if this is too invasive a change.

Other direct uses of errno in src/misc.c are wrapped for win32 with wsa2errno, which is static to that file.

What's the difference in use cases between libssh2_poll and _libssh2_wait_socket? Asking because I wonder if _libssh2_wait_socket could move to misc.c where there are some other socket-related and session-struct-argument but non-public-api (start with underscore) functions, along with wsa2errno()

The two functions are very different in terms of how they're handling the FDs arrays for poll/select, so there wasn't any obvious way to make one a convenience wrapper around the other. Notably, libssh2_poll only uses the return values from poll/select if greater-than-zero, so error conditions seem to be ignored. (I suppose that's as good a reason as any to mark this deprecated? https://www.libssh2.org/libssh2_poll.html)

sodabrew:
Added tests based on the example from #955 to show `master` branch fails then this branch solves for EINTR.
```
On master:
      Start 25: test_sshd
25/43 Test #25: test_sshd .................................***Failed    0.33 sec
On branch:
      Start 25: test_sshd
25/43 Test #25: test_sshd .................................   Passed    0.41 sec
```

sodabrew:
Thanks for approving the test run, found I can run `./ci/checksrc.sh` locally pre-check my fixes :) Pushed again to pass the linter!

src/misc.c
```diff
@@ -159,23 +159,28 @@ _libssh2_recv(libssh2_socket_t sock, void *buffer, size_t length,
...
     rc = recv(sock, buffer, length, flags);
+    if(rc < 0) {
 #ifdef WIN32
-    if(rc < 0)
-        return -wsa2errno();
+        int err = _libssh2_wsa2errno();
 #else
-    if(rc < 0) {
+        int err = errno;
+#endif
```
vszakats, line 167: I suggest deduping the `int err` declration into the top of the block. Also in the other similar cased in the patch:
```
... {
  int err;
#ifdef WIN32
  err = ...
#else
  err = ...
#endif
```

src/misc.c
```diff
@@ -192,18 +197,23 @@ _libssh2_send(libssh2_socket_t sock, const void *buffer, size_t length,
...
 
     rc = send(sock, buffer, length, flags);
+    if(rc < 0) {
 #ifdef WIN32
-    if(rc < 0)
-        return -wsa2errno();
+        int err = _libssh2_wsa2errno();
 #else
-    if(rc < 0) {
+        int err = errno;
```
vszakats, line 204: Here also.

src/session.c
```diff
@@ -668,6 +668,17 @@ int _libssh2_wait_socket(LIBSSH2_SESSION *session, time_t start_time)
                               "Timed out waiting on socket");
     }
     if(rc < 0) {
+#ifdef WIN32
+        int err = _libssh2_wsa2errno();
+#else
+        int err = errno;
+#endif
```
vszakats, line 675: Here also.

tests/test_ssh2.c
```diff
@@ -48,6 +78,24 @@ int main(int argc, char *argv[])
     LIBSSH2_CHANNEL *channel;
     int counter;
 
+#ifndef WIN32
+    pthread_t thread;
+
+    struct sigaction signal_handler_config = {
+      .sa_flags = SA_RESTART | SA_SIGINFO,
+      .sa_sigaction = signal_handler_function
```
vszakats, line 86:
```suggestion
        .sa_flags = SA_RESTART | SA_SIGINFO,
        .sa_sigaction = signal_handler_function
```
```suggestion
      .sa_flags = SA_RESTART | SA_SIGINFO,
      .sa_sigaction = signal_handler_function
```

tests/test_ssh2.c
```diff
@@ -35,6 +40,31 @@ static void portable_sleep(unsigned int seconds)
...
+        pthread_kill(main_thread, SIGPROF);
+        usleep(1000);
+    } while(1);
+
+    return NULL;
+}
+
+static void signal_handler_function(int a, siginfo_t *b, void *c)
+{
+    /* Prevent unused variable warnings */
```
vszakats, line 61:
```suggestion
```

src/misc.c
sodabrew: Updated

tests/test_ssh2.c
```diff
@@ -35,6 +40,31 @@ static void portable_sleep(unsigned int seconds)
 #endif
 }
 
+#ifndef WIN32
+static pthread_t main_thread;
+
+static void *background_thread(void *unused)
+{
+    fprintf(stderr, "Profiling thread sending SIGPROF every 1ms!\n");
+    (void)unused;
```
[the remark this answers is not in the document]
sodabrew: done

src/misc.c
```diff
@@ -159,23 +159,29 @@ _libssh2_recv(libssh2_socket_t sock, void *buffer, size_t length,
...
+        err = _libssh2_wsa2errno();
 #else
-    if(rc < 0) {
+        err = errno;
+#endif
+        /* Profiling tools that use SIGPROF can cause EINTR responses.
+           recv() does not modify its arguments when it returns EINTR,
+           but there may be data waiting, so the caller should try again */
+        if(err == EINTR)
+            return 0;
```
sodabrew, line 173: I observed in tests that returning `-EAGAIN` here also passes. That might be more consistent with this block of code.

tests/test_ssh2.c
sodabrew: Pushed with four spaces

tests/test_ssh2.c
```diff
@@ -48,6 +78,24 @@ int main(int argc, char *argv[])
...
+    sigemptyset(&signal_handler_config.sa_mask);
+
+    if(sigaction(SIGPROF, &signal_handler_config, NULL) != 0) {
+        fprintf(stderr, "Could not install signal handler\n");
+    }
+
+    main_thread = pthread_self();
+
+    pthread_create(&thread, NULL, background_thread, NULL);
+#endif
```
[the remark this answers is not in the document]
sodabrew: Updated to test for pthread.h / signal.h

sodabrew:
👋 friendly nudge, please let me know if there's anything I should adjust or things to discuss for this PR? thank you!

vszakats:
Aside from the smaller issues. My main concern is adding pthreads as a dependency to libssh2. I'm personally not ready to take this burden.

Any input from fellow maintainers?

sodabrew:
That’s fair, though I’ll say that the pthread library is only used in the
tests, and it should fail gracefully if not present (not checking the eintr
functionality in that run)

I could redo it used fork() if that would resolve this concern?

src/session.c
```diff
@@ -670,6 +670,18 @@ int _libssh2_wait_socket(LIBSSH2_SESSION *session, time_t start_time)
                               "Timed out waiting on socket");
     }
     if(rc < 0) {
+        int err;
+#ifdef WIN32
```
vszakats, line 674:
```suggestion
#ifdef _WIN32
```

src/misc.h
```diff
@@ -79,6 +79,11 @@ int _libssh2_error_flags(LIBSSH2_SESSION* session, int errcode,
                          const char *errmsg, int errflags);
 int _libssh2_error(LIBSSH2_SESSION* session, int errcode, const char *errmsg);
 
+#ifdef WIN32
```
vszakats, line 82:
```suggestion
#ifdef _WIN32
```

tests/test_ssh2.c
```diff
@@ -40,6 +45,30 @@ static void portable_sleep(unsigned int seconds)
 #endif
 }
 
+#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
+static pthread_t main_thread;
+
+static void *background_thread(void *unused)
```
vszakats, line 51:
```suggestion
#if defined(HAVE_PTHREAD_H)
static void *background_thread(void *unused)
```

tests/test_ssh2.c
```diff
@@ -24,6 +24,11 @@
 #include <stdio.h>
 #include <stdlib.h>  /* for getenv() */
 
+#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
+#include <pthread.h>
+#include <signal.h>
+#endif
```
vszakats, line 30:
```suggestion
#if defined(HAVE_PTHREAD_H)
#include <pthread.h>
#include <signal.h>

static pthread_t main_thread;
#endif
```
`signal.h` is standard C89, let's drop the guard. Also from the rest of the code.

`pthread.h` needs detection in autotools too.

These said, I'm also fine if we don't add this sample/test code to the test.

vszakats (changes requested):

tests/test_ssh2.c
```diff
@@ -63,6 +92,24 @@ int main(int argc, char *argv[])
...
+    pthread_t thread;
+
+    struct sigaction signal_handler_config = {
+        .sa_flags = SA_RESTART | SA_SIGINFO,
+        .sa_sigaction = signal_handler_function
+    };
+    sigemptyset(&signal_handler_config.sa_mask);
+
+    if(sigaction(SIGPROF, &signal_handler_config, NULL) != 0) {
+        fprintf(stderr, "Could not install signal handler\n");
```
vszakats, line 105:
```suggestion
        fprintf(stderr, "Could not install signal handler.\n");
```

tests/test_ssh2.c
```diff
@@ -63,6 +92,24 @@ int main(int argc, char *argv[])
     }
 #endif
 
+#if defined(HAVE_PTHREAD_H) && defined(HAVE_SIGNAL_H)
```
vszakats, line 95:
```suggestion
#if defined(HAVE_PTHREAD_H)
```

tests/test_ssh2.c
```diff
@@ -63,6 +92,24 @@ int main(int argc, char *argv[])
...
 #endif
 
+#if defined(HAVE_PTHREAD_H)
+    pthread_t thread;
+
+    struct sigaction signal_handler_config = {
+        .sa_flags = SA_RESTART | SA_SIGINFO,
+        .sa_sigaction = signal_handler_function
+    };
+    sigemptyset(&signal_handler_config.sa_mask);
```
vszakats, line 102: `sigemptyset` is non-portable.

tests/test_ssh2.c
```diff
@@ -63,6 +92,24 @@ int main(int argc, char *argv[])
...
+
+    if(sigaction(SIGPROF, &signal_handler_config, NULL) != 0) {
```
vszakats, line 104: `sigaction` also non-portable.

sodabrew:
Thanks for digging into this PR @vszakats! Let me know if there's anything you'd like me to work on, or if you've got it from here?

tests/test_ssh2.c
sodabrew: SIGPROF is the whole reason for this PR. It is used in Linux to gather process status. Failing to handle SIGPROF results in SSH connection failure rather than allowing the application using the library to be profiled by an external measurement tool.

It is OK to be non-portable so long as it's properly wrapped to prevent breaking on systems that don't have SIGPROF.

Because all systems (that I am aware of) with `SIGPROF` also have `sigaction`, I would also see no reason not to use `sigaction` HTH!

tests/test_ssh2.c
sodabrew: Also OK to be non-portable for the same reason as below

vszakats:
Thanks @sodabrew. This is now blocked by a non-portable signal functions and constants in `test_ssh2.c`. And pthread dependence. Also waiting for tests to finish. I'm siding with merging this, but without the test bits where all these problems are.

tests/test_ssh2.c
vszakats: We support more than just Linux. Also as I understand this is a causal example code committed to `tests`, but the essence of the feature is in `src`. CI runs fine now, but `pthread.h` detection is missing from autotools, and broken in CMake, so none of the example code is actually built.

Fixing autotools to have pthread detection might be one step forward, but it is complex. Even CMake is non-trivial if I look at how curl does these. We should also limit these detections to builds with tests enabled, otherwise it makes all builds slower. Only after implemeting these detection, could we see how the code compiles and runs in CI.

vszakats:
Thank you! Merged without the example code from tests.

vszakats:
For the record, here's the skipped part of the diff:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9b0ecca21..67c9fa8d7 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,6 +212,7 @@ if(NOT WIN32)
   check_include_files("sys/un.h" HAVE_SYS_UN_H)
   check_include_files("arpa/inet.h" HAVE_ARPA_INET_H)  # example and tests
   check_include_files("netinet/in.h" HAVE_NETINET_IN_H)  # example and tests
+  check_include_files("pthread.h" HAVE_PTHREAD_H)  # tests
 endif()
 
 # CMake uses C syntax in check_symbol_exists() that generates a warning with
diff --git a/tests/test_ssh2.c b/tests/test_ssh2.c
index 2e6fa2382..a7f0f3743 100644
--- a/tests/test_ssh2.c
+++ b/tests/test_ssh2.c
@@ -24,6 +24,13 @@
 #include <stdio.h>
 #include <stdlib.h>  /* for getenv() */
 
+#if defined(HAVE_PTHREAD_H)
+#include <pthread.h>
+#include <signal.h>
+
+static pthread_t main_thread;
+#endif
+
 static const char *hostname = "127.0.0.1";
 static const int port_number = 4711;
 static const char *pubkey = "key_rsa.pub";
@@ -40,6 +47,28 @@ static void portable_sleep(unsigned int seconds)
 #endif
 }
 
+#if defined(HAVE_PTHREAD_H)
+static void *background_thread(void *unused)
+{
+    (void)unused;
+    fprintf(stderr, "Profiling thread sending SIGPROF every 1ms.\n");
+
+    do {
+        pthread_kill(main_thread, SIGPROF);
+        usleep(1000);
+    } while(1);
+
+    return NULL;
+}
+
+static void signal_handler_function(int a, siginfo_t *b, void *c)
+{
+    (void)a;
+    (void)b;
+    (void)c;
+}
+#endif
+
 int main(int argc, char *argv[])
 {
     uint32_t hostaddr;
@@ -63,6 +92,26 @@ int main(int argc, char *argv[])
     }
 #endif
 
+#if defined(HAVE_PTHREAD_H)
+    {
+        pthread_t thread;
+
+        struct sigaction signal_handler_config = {
+            .sa_flags = SA_RESTART | SA_SIGINFO,
+            .sa_sigaction = signal_handler_function
+        };
+        sigemptyset(&signal_handler_config.sa_mask);
+
+        if(sigaction(SIGPROF, &signal_handler_config, NULL) != 0) {
+            fprintf(stderr, "Could not install signal handler.\n");
+        }
+
+        main_thread = pthread_self();
+
+        pthread_create(&thread, NULL, background_thread, NULL);
+    }
+#endif
+
     (void)argc;
     (void)argv;
 
@@ -160,7 +209,7 @@ int main(int argc, char *argv[])
                rc != LIBSSH2_ERROR_KEY_EXCHANGE_FAILURE ||
 #endif
                ++retry > retries) {
-                break;
+                goto shutdown;
             }
             fprintf(stderr, "Retrying... %d / %d\n", retry, retries);
         } while(1);
```

PR #663, opened 2022-01-14
use modern API in userauth_keyboard_interactive()

xalopp:
This refactors SSH_MSG_USERAUTH_INFO_REQUEST processing in `userauth_keyboard_interactive()` in order to improve robustness, correctness and readability or the code.

In order to be able to do unit testing, I moved some portion of the code from `src/userauth.c` to  `src/userauth_kbd_packet.c`

PR #1072, opened 2023-05-31
autotools: skip tests requiring static lib if `--disable-static`

### LINKED ISSUE #1056: `make check` fails with `--disable-static`
**Describe the bug**

`make check` fails with `--disable-static`.

**To Reproduce**

```
./configure --prefix=/usr --disable-static && make && make check
```

It produces:

```
/usr/bin/ld: test_auth_keyboard_info_request.o: in function `main':
test_auth_keyboard_info_request.c:(.text.startup+0xec): undefined reference to `userauth_keyboard_interactive_decode_info_request'
/usr/bin/ld: test_auth_keyboard_info_request.c:(.text.startup+0x201): undefined reference to `userauth_keyboard_interactive_decode_info_request'
collect2: error: ld returned 1 exit status
```

**Expected behavior**

Test succeeds.

**Version:**
 - OS and version: Linux (Linux From Scratch)
 - libssh2 version: 1.11.0

**Additional context**

It looks like the reason is `userauth_keyboard_interactive_decode_info_request` not exported in the shared library because we are linking libssh2.so with `-export-symbols-regex '^libssh2_.*'`.

xry111:
Regression from 83853f8aea0e2f739cacd491632eb7fd3d03ad2d #663
Fixes #1056


vszakats:
Thank you! Merged now.

PR #1082, opened 2023-06-04
copyright: remove years from copyright headers

vszakats:
Also:
- uppercase `(C)`.
- add missing 'All rights reserved.' lines.
- drop duplicate 'Author' lines.
- add copyright headers where missing.
- enable copyright header check in checksrc.

Reasons for deleting years (copied as-is from curl):
- they are mostly pointless in all major jurisdictions
- many big corporations and projects already don't use them
- saves us from pointless churn
- git keeps history for us
- the year range is kept in COPYING

Closes #1082

### DIFF
--- a/.checksrc
+++ b/.checksrc
@@ -1,4 +1,3 @@
-disable COPYRIGHT
 disable FOPENMODE
 disable SNPRINTF
 disable TYPEDEFSTRUCT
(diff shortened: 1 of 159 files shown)


bagder (approved):
One of these days we could also add SPDX identifiera and make a REUSE CI job...

PR #1089, opened 2023-06-08
test_sshd.test: set a safe PID directory

dfandrich:
The compiled in default to sshd can be a non-writable location since it
expects to be run as root.


tests/test_sshd.test
```diff
@@ -76,6 +76,7 @@ chmod go-rwx \
   -h "${d}/openssh_server/ssh_host_rsa_key" \
   -h "${d}/openssh_server/ssh_host_ecdsa_key" \
   -h "${d}/openssh_server/ssh_host_ed25519_key" \
+  -o "PidFile sshd.pid" \
```
vszakats, line 79: Looks good to me!

Minor suggestion:
```suggestion
  -o 'PidFile sshd.pid' \
```

PR #1093, opened 2023-06-14
hostkey: do not advertise ssh-rsa when SHA1 is disabled

### LINKED ISSUE #1092: SHA1/2 issue
I believe there is still some issue with the latest release of libssh2 related to SHA1 vs SHA2.

If I build libssh2 like this:

-DLIBSSH2_NO_RSA_SHA1 -DLIBSSH2_NO_DSA -DLIBSSH2_NO_RC4 -DLIBSSH2_NO_CAST -DLIBSSH2_NO_3DES -DLIBSSH2_NO_MD5

curl cannot negotiate keys:

* Set "rsa-sha2-256,rsa-sha2-512,ssh-rsa" as SSH hostkey type
* Failure establishing ssh session: -8, Unable to exchange encryption keys

I then tried to add -DLIBSSH2_RSA_SHA2 but the result was the same.

It only works if I enable SHA1 with -DLIBSSH2_RSA_SHA1.

Even though the negotiation occurs with the available SHA2 key, SHA1 is still allowed by libssh2, which is a problem.

vszakats:
Before this patch OpenSSL and mbedTLS advertised both SHA2 and SHA1 host key algos, even when SHA1 was not supported by the crypto backend or forcefully disabled via `LIBSSH2_NO_RSA_SHA1`.

Reported-by: João M. S. Silva
Fixes #1092
Closes #1093

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -1453,7 +1453,11 @@ _libssh2_supported_key_sign_algorithms(LIBSSH2_SESSION *session,
 #if LIBSSH2_RSA_SHA2
     if(key_method_len == 7 &&
        memcmp(key_method, "ssh-rsa", key_method_len) == 0) {
-        return "rsa-sha2-512,rsa-sha2-256,ssh-rsa";
+        return "rsa-sha2-512,rsa-sha2-256"
+#if LIBSSH2_RSA_SHA1
+            ",ssh-rsa"
+#endif
+            ;
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


willco007:
wining and os400 backends would also need this change. Also, I don't think this is the actual issue in #1092. If sha2 was configured correctly on that server, `rsa-sha2-512` or `rsa-sha2-256` would match before `ssh-rsa` and it would carry on it's merry way and auth successfully.

vszakats:
@willco007 Yes, it's strange; it'd be useful to understand what is happening there exactly. Thanks for the heads-up about OS/400 and WinCNG, tackled those as well now with a force-push.

willco007:
This looks good regardless of if it fixes #1092 or not.

vszakats:
According to https://github.com/libssh2/libssh2/issues/1092#issuecomment-1591656938, this patch fixes #1092.

PR #1098, opened 2023-06-21
tests: trap signals in scripts

vszakats:


vszakats:
Unrelated REUSE-check failure:
https://github.com/libssh2/libssh2/actions/runs/5333572682/jobs/9664211341?pr=1098#step:4:8

The checker tool was upgraded an hour ago to 2.0.0, which is the likely reason:
https://github.com/fsfe/reuse-tool/releases/tag/v2.0.0

The newly found reuse issues seem legit.

PR #1110, opened 2023-07-01
Fix a NULL pointer reference, when LIBSSH2_ALLOC() failed.

rmsh1216:
We should check whether *key_method is a NULL pointer instead of key_method


vszakats:
Thank you!

PR #1111, opened 2023-07-01
misc: delete redundant NULL check and assignment

vszakats:
Follow-up to 724effcb47ebb713d3ef1776684b8f6407b4b6a5 #1109

Ref: https://github.com/libssh2/libssh2/pull/1109#discussion_r1246613274
Closes #1111

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -901,8 +901,6 @@ int _libssh2_copy_string(LIBSSH2_SESSION *session, struct string_buf *buf,
         }
     }
     else {
-        if(outlen)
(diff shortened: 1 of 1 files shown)


src/misc.c
```diff
@@ -901,8 +901,6 @@ int _libssh2_copy_string(LIBSSH2_SESSION *session, struct string_buf *buf,
         }
     }
     else {
-        if(outlen)
-            *outlen = 0;
```
vszakats, line 905: This is a test.

PR #1121, opened 2023-07-14
cmake: CMAKE_SOURCE_DIR -> PROJECT_SOURCE_DIR

Tindarid:
Fixes compiling as dependency with FetchContent

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -174,7 +174,7 @@ export(TARGETS ${libssh2_export} NAMESPACE libssh2:: FILE libssh2-targets.cmake)
 export(PACKAGE libssh2)  # register it
 
 # Generate libssh2-config.cmake into build tree and install it
-configure_file(${CMAKE_SOURCE_DIR}/cmake/libssh2-config.cmake.in libssh2-config.cmake @ONLY)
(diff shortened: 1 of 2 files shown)


vszakats:
Should we also fixup these two uses of `CMAKE_SOURCE_DIR`?:
tests/CMakeLists.txt:  COMMAND gcovr -r "${CMAKE_SOURCE_DIR}" --exclude tests/*
tests/CMakeLists.txt:  COMMAND gcovr -r "${CMAKE_SOURCE_DIR}" --exclude tests/* --html-details --output "${CMAKE_CURRENT_BINARY_DIR}/coverage/index.html")

vszakats (changes requested):
It seems gcov tests are also affected by this. These uses of `CMAKE_SOURCE_DIR` also need to be updated to fix this in a single patch:
```
tests/CMakeLists.txt:  COMMAND gcovr --root "${CMAKE_SOURCE_DIR}" --exclude tests/*
tests/CMakeLists.txt:  COMMAND gcovr --root "${CMAKE_SOURCE_DIR}" --exclude tests/* --html-details --output "${CMAKE_CURRENT_BINARY_DIR}/coverage/index.html")
```

vszakats (approved):

vszakats:
Thanks @Tindarid! Merged now, with the extra two replacement in `tests/CMakeLists.txt`.

PR #1133, opened 2023-07-25
Fix private keys encrypted with aes-gcm methods

jakob:
libssh2 1.11.0 fails to decrypt private keys encrypted with aes128-gcm@openssh.com and aes256-gcm@openssh.com ciphers.

To reproduce the issue, you can create a test key with a command like the following:

```bash
ssh-keygen -Z aes256-gcm@openssh.com -f id_aes256-gcm
```

If you attempt to use this key for authentication, libssh2 will return the not-so-helpful error message "Wrong passphrase or invalid/unrecognized private key file format".

The problem is that OpenSSH encrypts keys differently than packets. It does not include the length as AAD, and the 16 byte authentication tag is appended after the encrypted key. The length of the authentication tag is not included in the encrypted key length.

I have not found any documentation for this behaviour -- I discovered it by looking at the OpenSSH source. See the `private2_decrypt` function in [sshkey.c](https://github.com/openssh/openssh-portable/blob/master/sshkey.c).

This PR fixes the code for reading OpenSSH private keys encrypted with AES-GCM methods.

### DIFF
--- a/src/pem.c
+++ b/src/pem.c
@@ -624,6 +624,7 @@ _libssh2_openssh_pem_parse_data(LIBSSH2_SESSION * session,
 			if (!_libssh2_check_length(&decoded, 16)) {
 				ret = _libssh2_error(session, LIBSSH2_ERROR_PROTO,
 									 "GCM auth tag missing");
+				method->dtor(session, &abstract);
 				goto out;
 			}
 			if(method->crypt(session, decoded.dataptr, 16, &abstract, LAST_BLOCK)) {
--- a/src/pem.c
+++ b/src/pem.c
@@ -628,7 +628,7 @@ _libssh2_openssh_pem_parse_data(LIBSSH2_SESSION * session,
 			}
 			if(method->crypt(session, decoded.dataptr, 16, &abstract, LAST_BLOCK)) {
 				ret = _libssh2_error(session, LIBSSH2_ERROR_DECRYPT,
-									 "GCM auth tag invalid");;
+									 "GCM auth tag invalid");
 				method->dtor(session, &abstract);
 				goto out;
 			}
--- a/src/pem.c
+++ b/src/pem.c
@@ -623,19 +623,19 @@ _libssh2_openssh_pem_parse_data(LIBSSH2_SESSION * session,
 
         /* for the AES GCM methods, the 16 byte authentication tag is
          * appended to the encrypted key */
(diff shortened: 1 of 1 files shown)


src/pem.c
```diff
@@ -616,6 +621,27 @@ _libssh2_openssh_pem_parse_data(LIBSSH2_SESSION * session,
 
         /* No padding */
 
+        /* for the AES GCM methods, the 16 byte authentication tag is
+         * appended to the encrypted key */
+        if(strcmp(method->name, "aes256-gcm@openssh.com") == 0 ||
+        strcmp(method->name, "aes128-gcm@openssh.com") == 0) {
```
vszakats, line 627:
```suggestion
        if(strcmp(method->name, "aes256-gcm@openssh.com") == 0 ||
           strcmp(method->name, "aes128-gcm@openssh.com") == 0) {
```

src/pem.c
```diff
@@ -616,6 +621,27 @@ _libssh2_openssh_pem_parse_data(LIBSSH2_SESSION * session,
...
+            if(!_libssh2_check_length(&decoded, 16)) {
+                ret = _libssh2_error(session, LIBSSH2_ERROR_PROTO,
+                                     "GCM auth tag missing");
+                method->dtor(session, &abstract);
+                goto out;
+            }
+            if(method->crypt(session, decoded.dataptr, 16, &abstract,
+                             LAST_BLOCK
+                             )) {
```
vszakats, line 636:
```suggestion
                             LAST_BLOCK)) {
```

vszakats:
Thank you @jakob, looks good to me!

vszakats (approved):

src/pem.c
```diff
@@ -599,12 +599,17 @@ _libssh2_openssh_pem_parse_data(LIBSSH2_SESSION * session,
...
+             * loop.
+             */
             if(method->crypt(session, decrypted.data + len_decrypted,
                              blocksize,
                              &abstract,
-                             len_decrypted == 0 ? FIRST_BLOCK : (
-                         ((size_t)len_decrypted == decrypted.len - blocksize) ?
-                               LAST_BLOCK : MIDDLE_BLOCK)
+                             MIDDLE_BLOCK
                              )) {
```
vszakats, line 613:
```suggestion
                             MIDDLE_BLOCK)) {
```

PR #1134, opened 2023-07-25
NMakefile: delete

vszakats:
This make file was for long time unmaintained (last updated in 2014).
Despite best efforts to keep it working in the recent round of major
overhauls, it appears to be broken now. There is also no way to test it
without an actual MSVC env and it's also missing from our CI. Based on
our Issue tracker, it's also not widely used.

Since its addition in 2005, libssh2 got support for CMake in 2014.
CMake should be able to generate NMake makefiles with the option
`-G "NMake Makefiles"`. (I haven't tested this.)

Ref: #1129
Closes #1134


vszakats:
This brings down supported build systems to 5, of which 3 are generic, 2 are platform-specific: autotools, CMake, GNU Make + OS/400, VMS.

PR #1136, opened 2023-07-26
sftp: Set correct write offset in append mode

richardweinberger:
Currently the file offset starts with zero in all modes. In append mode, this causes SSH_FXP_WRITE commands with wrong offsets.

When appending N bytes to a X bytes long file on the remote side SSH_FXP_WRITE will have an offset parameter of N instead of N+X.

If the remote SFTP server is OpenSSH based, this is no problem since OpenSSH's SFTP server applies the offset only if SSH_FXF_APPEND was not present at open time.
It is also no problem if the remote operating system follows POSIX O_APPEND semantics (ignoring lseek()).

But there are others.
I've been facing problem with Core FTP Server on Microsoft Windows, this server seems to apply the offset blindly.
SFTP append using libssh worked, but not with libssh2.

So, let's stay on the safe side and do it like libssh, learn the remote file length at open time and apply the remote length always to the offset.

That way a SFTP implementation can correctly implement append mode by just looking at the offsets without relying on O_APPEND semantics.

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -1352,8 +1352,8 @@ sftp_open(LIBSSH2_SFTP *sftp, const char *filename,
         fp->u.file.offset_sent = 0;
 
         if(mode & LIBSSH2_FXF_APPEND) {
-            int ret = sftp_stat(fp->sftp, filename, (uint32_t)filename_len,
-                                    LIBSSH2_SFTP_STAT, &attrs);
+            int ret = sftp_stat(fp->sftp, filename, (unsigned int)filename_len,
+                                LIBSSH2_SFTP_STAT, &attrs);
 
             if(ret) {
                 LIBSSH2_FREE(session, fp);
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -101,6 +101,9 @@ static int sftp_packet_ask(LIBSSH2_SFTP *sftp, unsigned char packet_type,
                            uint32_t request_id, unsigned char **data,
                            size_t *data_len);
 static void sftp_packet_flush(LIBSSH2_SFTP *sftp);
+static int sftp_stat(LIBSSH2_SFTP *sftp, const char *path,
+                     unsigned int path_len, int stat_type,
+                     LIBSSH2_SFTP_ATTRIBUTES * attrs);
 
 /* _libssh2_store_u64
  */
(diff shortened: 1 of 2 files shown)


src/sftp.c
```diff
@@ -1348,6 +1351,22 @@ sftp_open(LIBSSH2_SFTP *sftp, const char *filename,
         fp->u.file.offset = 0;
         fp->u.file.offset_sent = 0;
 
+        if(mode & LIBSSH2_FXF_APPEND) {
+            int ret = sftp_stat(fp->sftp, filename, (uint32_t)filename_len,
+                                    LIBSSH2_SFTP_STAT, &attrs);
```
vszakats, line 1356:
```suggestion
            int ret = sftp_stat(fp->sftp, filename, (unsigned int)filename_len,
                                LIBSSH2_SFTP_STAT, &attrs);
```

dfandrich:
It sounds like this could result in a TOCTOU race. Is there any way to avoid that?

src/sftp.c
vszakats: No worries, indeed that's the type used in the called function.

vszakats:
If this introduces TOCTOU also for servers that work correctly without the patch, shouldn't we make this an option, or (auto?)enable it only for servers that do require it?

vszakats:
Another scenario if two uploaders append to the same file (if that's possible at all?). Not saying it's widespread or good practice, but behaviour will change for existing cases, for popular servers. I can't personally judge the impact.

willco007:
@vszakats I agree, I'd rather this behavior be opt-in. I've shipped an sftp client for 15+ years and have never run into a server that behaves this way.

jas4711:
What does the protocol spec say here?  What behavior is intended, and is this specified clearly enough?  If not, maybe this warrant for a protocol clarification.

vszakats:
@richardweinberger We might add a new local flag, e.g.:
```c
#define LIBSSH2_FXF_APPEND_FIXUP_OFFSET         0x10000001
```
and put this feature behind it.

vszakats:
I wasn't thinking of auto-detecting it, just offering it as an explicit option. Caller may provide a manual option for it, or may try to auto-detect it on its own.

vszakats:
For buggy servers, another alternative may be to manually do the open → stat → seek → write sequence. This doesn't require a libssh2 change. In effect it means to avoid the built-in append feature with servers that don't handle it correctly.

vszakats:
libssh2 is not my project by any definition. Feedback so far wasn't fully receptive and seems to prefer opt-in.

Maybe worth raising this issue with curl directly? The scope may be narrower there, with a wider space for workarounds. Also folks there may have more experience with this.

vszakats:
Thanks for your work on this PR, but, since we're not all aligned behind it, I'm closing this for now. We may revisit another take that makes this opt-in/optional or auto-detect.

Issue #773, opened 2022-11-16
libssh2_userauth_publickey_frommemory() may fail nondeterministically due to incorrect use of strlen()


mike-jumper:
The `libssh2_userauth_publickey_frommemory()` function accepts private key data in the form of an arbitrary buffer and its length. From [the documentation for `libssh2_userauth_publickey_frommemory()`](https://www.libssh2.org/libssh2_userauth_publickey_frommemory.html):

>
> `privatekeydata` - Buffer containing the contents of a private key file.
> `privatekeydata_len` - Length of private key data.
>

... but that buffer is passed down the chain via the `filename` member of a `struct privkey_file`:

https://github.com/libssh2/libssh2/blob/f6694beb21518491fe9d02c56ae14a0319d79a8d/src/userauth.c#L1815-L1820

When that buffer is ultimately processed by `sign_frommemory()`, the originally-provided buffer length has been effectively discarded, and its length is instead calculated as if it were a string using `strlen()`:

https://github.com/libssh2/libssh2/blob/f6694beb21518491fe9d02c56ae14a0319d79a8d/src/userauth.c#L830-L835

Depending on the size of the block and memory garbage, this will overread the buffer and give nondeterministic results.

willco007:
At first glance, it seems the easiest solution would be to add `filename_len` to the `privkey_file` struct. Having `filename` be overloaded with a bag of bytes or a file path could be improved. Seems two unique variables would help clarify what the intent is.

mike-jumper:
With `privkey_file` being an internal struct anyway, I'd suggest defining a struct specific to in-memory private keys such that its usage is clear. Continuing to repurpose an established structure with otherwise well-defined semantics is going to be asking for trouble.

vszakats:
@mike-jumper Could you prepare a patch for this?

PR #1143, opened 2023-08-01
Add a new structure to  separate memory read and file read.

rmsh1216:
We use different APIs  when we read one private key from memory, so it is improper to store the private key information in the structure that stores the private key file information.

Fixes https://github.com/libssh2/libssh2/issues/773
Reported-by: mike-jumper


vszakats (approved):

vszakats:
Thank you @rmsh1216!

PR #1147, opened 2023-08-08
ClearMemory: do not call SecureZeroMemory on UWP

JohannesKauffmann:
SecureZeroMemory is not supported on UWP, and there is no replacement. See:

https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa366877(v=vs.85)

### DIFF
--- a/src/misc.h
+++ b/src/misc.h
@@ -45,7 +45,8 @@
                                               (void)(size); \
                                           } while(0)
 #else
-#ifdef WIN32
+/* SecureZeroMemory is not supported on UWP */
+#ifdef WIN32 && !defined(LIBSSH2_WINDOWS_UWP)
 #define _libssh2_explicit_zero(buf, size) SecureZeroMemory(buf, size)
 #elif defined(HAVE_EXPLICIT_BZERO)
 #define _libssh2_explicit_zero(buf, size) explicit_bzero(buf, size)


vszakats:
Our UWP CI jobs build fine with  SecureZeroMemory enabled. It's entrinsic function, not a system DLL call. In what way does this break UWP apps?

JohannesKauffmann:
Hi, thanks for the response. It seems like the current version of MinGW does not have SecureZeroMemory for UWP, but the official Windows SDK does. So it is allowed after all, it's just that MinGW is broken. And I happened to try and build libssh2 with MinGW :)

See also:
- https://code.videolan.org/videolan/vlc/-/merge_requests/4065
- https://sourceforge.net/p/mingw-w64/mailman/message/37881675/

I'll close this.

vszakats:
Thanks for the info and links!

So what we are missing is a MinGW UWP job, plus a local workaround for the mingw-w64 header issue.

vszakats:
Fixing up the mingw-w64 header bug locally looks impractical. So, here is an untested patch that skips using `SecureZeroMemory()` with UWP and mingw-w64 older than 11.0.2 (assuming the fix is going to land in that future revision). In CMake builds this falls back to our local `_libssh2_memzero()` (subject to compiler optimizations, and may or may not actually zero memory.) autotools tries to detect unixy alternatives, but none of those is normally available on Windows.

I'd only commit this if we had a CI job needing it.

```diff
diff --git a/src/libssh2_priv.h b/src/libssh2_priv.h
index 9d6137e6..6e54da55 100644
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -84,7 +84,6 @@
 #include "libssh2.h"
 #include "libssh2_publickey.h"
 #include "libssh2_sftp.h"
-#include "misc.h"
 
 #ifdef WIN32
 /* Detect Windows App environment which has a restricted access
@@ -99,6 +98,8 @@
 # endif
 #endif
 
+#include "misc.h"
+
 #ifndef FALSE
 #define FALSE 0
 #endif
diff --git a/src/libssh2_setup.h b/src/libssh2_setup.h
index 8e56a060..e76edbf6 100644
--- a/src/libssh2_setup.h
+++ b/src/libssh2_setup.h
@@ -16,6 +16,14 @@
 /* Define mingw-w64 version macros, eg __MINGW{32,64}_{MINOR,MAJOR}_VERSION */
 #ifdef __MINGW32__
 #include <_mingw.h>
+#define MINGW64_VER(major, minor, bugfix) \
+    (__MINGW64_VERSION_MAJOR > major || \
+    (__MINGW64_VERSION_MAJOR == major && \
+    (__MINGW64_VERSION_MINOR > minor || \
+    (__MINGW64_VERSION_MINOR == minor && \
+    __MINGW64_VERSION_BUGFIX >= bugfix))))
+#else
+#define MINGW64_VER(major, minor, bugfix)  (0)
 #endif
 
 /* Configuration provided by build tools (autotools and CMake),
diff --git a/src/misc.h b/src/misc.h
index bf3e3c36..b9fe79e8 100644
--- a/src/misc.h
+++ b/src/misc.h
@@ -44,8 +44,10 @@
                                               (void)(buf); \
                                               (void)(size); \
                                           } while(0)
-#else
-#ifdef WIN32
+/* SecureZeroMemory() not defined in UWP mode with mingw-w64 < 11.0.2 */
+#elif defined(WIN32) && \
+    (!defined(LIBSSH2_WINDOWS_UWP) || \
+    !defined(__MINGW64_VERSION_MAJOR) || MINGW64_VER(11, 0, 2))
 #define _libssh2_explicit_zero(buf, size) SecureZeroMemory(buf, size)
 #elif defined(HAVE_EXPLICIT_BZERO)
 #define _libssh2_explicit_zero(buf, size) explicit_bzero(buf, size)
@@ -58,7 +60,6 @@
 void _libssh2_memzero(void *buf, size_t size);
 #define _libssh2_explicit_zero(buf, size) _libssh2_memzero(buf, size)
 #endif
-#endif
 
 struct list_head {
     struct list_node *last;
```

vszakats:
Added a mingw-w64 UWP build, but it doesn't reproduce this. Not sure why. #1155

PR #1155, opened 2023-08-10
ci: add mingw-w64 UWP build

vszakats:
Add a CI test for Windows UWP builds using mingw-w64. Before this patch
we had UWP builds tested with MSVC only.

Alike existing UWP jobs, it's not possible to run the binaries due to
the missing UWP runtime DLL:
https://github.com/libssh2/libssh2/actions/runs/5821297010/job/15783475118#step:11:42

We could install `winstorecompat-git` in the setup-msys2 step, but opted
to do it manually to avoid the overhead for every matrix job.

All this would work smoother with llvm-mingw, which features an UWP
toolchain prefix and provides all necessary implibs by default.

This also hit a CMake bug (with v3.26.4), where CMake gets confused and
sets up `windres.exe` to use the MSVC rc.exe-style command-line:
https://github.com/libssh2/libssh2/actions/runs/5819232677/job/15777236773#step:9:126

Notice that MS "sunset" UWP in 2021:
https://github.com/microsoft/WindowsAppSDK/discussions/1615

If this particular CI job turns out to be not worth the maintenance
burden or CPU time, or too much of a hack, feel free to delete it.

Ref: https://github.com/libssh2/libssh2/pull/1147#issuecomment-1670850890
Closes #1155

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,7 +154,7 @@ jobs:
           - { build: 'autotools', sys: clang64, env: clang-x86_64 }
           - { build: 'cmake'    , sys: ucrt64,  env: ucrt-x86_64 }
           - { build: 'cmake'    , sys: clang64, env: clang-x86_64 }
-          - { build: 'cmake'    , sys: mingw64, env: x86_64, mode: uwp }
+          - { build: 'cmake'    , sys: mingw64, env: x86_64, api: uwp }
           - { build: 'make'     , sys: mingw64, env: x86_64 }
       fail-fast: false
     steps:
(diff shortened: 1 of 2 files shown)


vszakats:
CMake gets confused and tries to use the MSVC command-line template when calling `windres`, then fails. Force-feeding it with the correct template (copied from CMake's own code) seems to fix it, though a terribly ugly hack.

vszakats:
This turned out to be super-ugly hack, as somewhat expected. While also bumping into a CMake bug. It is capable of detecting UWP issues though: https://github.com/libssh2/libssh2/actions/runs/5825785029

Suprisingly, `SecureZeroMemory()` worked without issues, so it wasn't effective in replicating the issue reported in #1147.

PR #1185, opened 2023-09-17
libssh2_session_get_blocking.3: Add description

concussious:


docs/libssh2_session_get_blocking.3
```diff
@@ -1,8 +1,8 @@
 .\" Copyright (C) The libssh2 project and its contributors.
 .\" SPDX-License-Identifier: BSD-3-Clause
-.TH libssh2_session_get_blocking 3 "1 Jun 2007" "libssh2 0.15" "libssh2"
+.TH libssh2_session_get_blocking 3 "17 Sep 2023" "libssh2" "libssh2"
```
vszakats, line 3: This entry is without an `AVAILABILITY` section, so the version number here seems useful. Why do we bump the date?

concussious: I'm happy to revert the changes to line 3.

The date is usually date of last change.
I saw version stripped from other pages in this directory. I suggested both of them because I wanted my patch for line 5 to be correct.

vszakats: Let's keep the `.TH` unchanged in this case (or move it to an `AVAILABILITY` section). Also AFAICS we set the date once when adding a new man page, this tells readers when we introduced that function. Let's keep that too.

vszakats (approved):

vszakats:
Thank you @concussious, merged.

PR #1187, opened 2023-09-21
autotools: fix selecting wincng in cross-builds

### LINKED ISSUE #1186: Configure fails with commit 88a960a codebase
**Describe the bug**
Configure fails with commit 88a960a codebase producing the error message,

```
configure: ERROR: No wincng crypto library found!

Specify --with-crypto=$backend and/or the necessary library search prefix.

Known crypto backends: auto, openssl, libgcrypt, mbedtls, wincng, wolfssl
configure: error: Required dependencies are missing.
```

**To Reproduce**
executing the following command from the added subdirectory build and with $PREFIX exported to a known directory,
`CC=x86_64-w64-mingw32-gcc .././configure --build=x86_64-linux-gnu --host=x86_64-w64-mingw32 --prefix=$PREFIX --disable-shared --enable-static --with-crypto=wincng --with-libz-prefix=$PREFIX LIBS=-lws2_32`

**Expected behavior**
The latest release version 1.11.0 of the library works fine with no configure errors. The cryto backend wincing is found and the library compiles as expected

 - OS `Arch Linux up-to-date rolling version`
 - libssh2 version `commit 88a960a codebase`
 - crypto backend and version: `wincng`

vszakats:
- Fix explicitly selecting WinCNG in autotools cross-builds by
  moving `windows.h` header check before the WinCNG availability
  check.
  Follow-up to d43b8d9b0b9cd62668459fe5d582ed83aabf77e7

  Reported-by: Jack L
  Fixes #1186

- Add Linux -> mingw-w64 cross-builds for autotools and CMake.
  This doesn't detect #1186, because that only happened when
  explicitly specifying WinCNG via `--with-crypto=wincng`, but
  not when falling back to WinCNG by default.

- autotools: fix to strip suffix from gcc version

  Before this patch we expected `n.n` `-dumpversion` output, but Ubuntu
  may return `n-win32` (also with `-dumpfullversion`). Causing these
  errors and failing to enable picky warnings:
  ```
  ../configure: line 23845: test: : integer expression expected
  ```
  Ref: https://github.com/libssh2/libssh2/actions/runs/6263453828/job/17007893718#step:5:143

  Fix that by stripping any dash-suffix.

  gcc version detection is still half broken because we translate '10' to
  '10.10' because `cut -d. -f2` returns the first word if the delimiter
  missing.

  More possible `-dumpversion` output:
  `10-posix`, `10-win32`, `9.3-posix`, `9.3-win32`, `6`, `9.3.0`, `11`, `11.2`, `11.2.0`
  Ref: https://github.com/mamedev/mame/pull/9767

Closes #1187

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,6 +165,43 @@ jobs:
           export OPENSSH_SERVER_IMAGE=ghcr.io/libssh2/ci_tests_openssh_server:$(git rev-parse --short=20 HEAD:tests/openssh_server)
           cd bld && ctest -VV --output-on-failure
 
+  build_linux_cross_mingw64:
+    name: 'linux -> mingw-w64'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        build: [autotools, cmake]
+    env:
+      TRIPLET: 'x86_64-w64-mingw32'
+    steps:
+      - uses: actions/checkout@v3
+      - name: 'install packages'
+        run: sudo apt-get --quiet 2 --option Dpkg::Use-Pty=0 install mingw-w64
+      - name: 'autotools autoreconf'
+        if: ${{ matrix.build == 'autotools' }}
+        run: autoreconf -fi
+      - name: 'autotools configure'
+        if: ${{ matrix.build == 'autotools' }}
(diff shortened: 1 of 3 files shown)


vszakats:
Windows MSVC OpenSSL job likely broken due to: https://github.com/actions/runner-images/issues/8344

PR #1200, opened 2023-09-30
appveyor: move to single quotes, rename a variable

vszakats:
Closes #xxxx


mback2k:
May I ask: why?

vszakats:
We've been using a mixture before and single quotes are easier to type (and read IMO) and won't try evaluating variables-etc inside, and also don't need as much char quoting.

PR #1207, opened 2023-10-30
Use non-deprecated OpenSSL APIs when linking against OpenSSL 3

MichaelBuckley:
Fixes #1206

PR #1214, opened 2023-11-08
openssl: use `OPENSSL_VERSION_NUMBER`

vszakats:
Uniformly use `OPENSSL_VERSION_NUMBER` to check for OpenSSL version.
Before this patch some places used `OPENSSL_VERSION_MAJOR`.

According to https://github.com/openssl/openssl/issues/17517, the macro
`OPENSSL_VERSION_NUMBER` is safe to use and not deprecated.

Closes #1214

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -565,8 +565,7 @@ _libssh2_cipher_crypt(_libssh2_cipher_ctx * ctx,
 #endif
     }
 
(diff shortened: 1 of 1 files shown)


vszakats:
Pending #1207, which addresses one of the fixes here.

vszakats:
Merging via #1207. Closing.

PR #1220, opened 2023-11-15
fix 32-bit timestamp issue in public API #735

### LINKED ISSUE #735: libssh2_sftp_stat_ex() uses "long" for times
**Describe the bug**

The `LIBSSH2_SFTP_ATTRIBUTES` struct has several fields using `unsigned long` to transfer timestamps. This makes them limited to 32 bits on Windows and 32 bit machines.

(I can't recall exactly how/if the SFTP protocol itself deals with the issue.)

doorsdown:
libssh2 issue #735 reported that timestamp fields in LIBSSH2_SFTP_ATTRIBUTES struct were limited to 32 bits on non 64 bit machines

Closes: #735

### DIFF
--- a/docs/libssh2_sftp_fstat_ex.3
+++ b/docs/libssh2_sftp_fstat_ex.3
@@ -52,7 +52,7 @@ struct _LIBSSH2_SFTP_ATTRIBUTES {
     unsigned long permissions;
 
     /* access time and modified time of file */
-    unsigned long atime, mtime;
+    time_t atime, mtime;
 };
 .fi
 
--- a/docs/libssh2_sftp_stat_ex.3
+++ b/docs/libssh2_sftp_stat_ex.3
@@ -56,8 +56,8 @@ struct LIBSSH2_SFTP_ATTRIBUTES {
     unsigned long uid;
     unsigned long gid;
     unsigned long permissions;
-    unsigned long atime;
-    unsigned long mtime;
+    time_t atime;
+    time_t mtime;
 };
 .fi
 .SH RETURN VALUE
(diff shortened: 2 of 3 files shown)


vszakats:
Are we comfortable with breaking the ABI for this?

If so, this will need an SO version bump. Then there is a bunch
of other things that might use a breaking upgrade if we're doing this.

willco007:
What reasonable options do we have that won't?

vszakats:
Introducing new API variants that work with the new types.

Perhaps _extending_ this struct with new 64-bit values (while
keeping the existing ones) might also work, I'm not sure.

willco007:
Adding new/duplicate API seems like not a great solution to a long term problem. Extending the struct would work, but again it's a bit of a fragile fix for a long term issue. I'd probably be OK breaking ABI but I could be convinced to keep it. :)

vszakats:
I'm not for or against an ABI break, but in case we do it, it's a bigger deal;
It needs coordination and completing the `At next SONAME bump` list in
`docs/TODO`.

vszakats:
I'm thinking putting this change behind an internal macro (e.g. `__LIBSSH2_ABI_BREAK_NEXT`) reserved for debug builds, alongside a CI test. This would allow to merge this early (now), and potentially allow to prepare new ABI breaking changes in the source, and release them when everything is ready.

Anyone opposing this?

vszakats:
To avoid the ABI break this would probably need a LIBSSH2_SFTP_ATTR64
struct and 64-bit versions of public functions accepting/returning this struct.
Then make the old APIs map to the new ones internally.

Seems doable, with a lot of legwork.

PR #1241, opened 2023-11-23
tests: fall back to `$LOGNAME` for username

### LINKED ISSUE #1240: Tests test_sshd.test 1 and 2 fail when environment variable $USER is not set
**Describe the bug**
If the tests are executed on a system where the environment variable `$USER` is not set, the tests `test_sshd.test` 1 and 2 fail.

I'm the maintainer for libssh2 in debian and I was informed in this bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1056348

The patch provided in the bug uses `$LOGNAME` if $USER isn't set.

**To Reproduce**
to reproduce the bug, unset the `$USER` variable, then run the tests:
`$ USER= make check`

**Expected behavior**
The test should pass but the tests `test_sshd.test` 1 and 2 fail when `$USER` isn't set

Linux Debian Bookworm with libssh2 1.11.0 and libssh2 github master branch

**Additional context**
The variable `$USER` isn't mandatory on POSIX systems, so some shell don't set it by default, the variable `$LOGNAME` is mandatory though, so it could be a replacement, or, as mentioned in the bugreport, calling `get(e)uid` and
`getpwent()` would be a better solution, but I'm relying on your wisdom here.

vszakats:
If the `$USER` variable is empty, fall back to using `$LOGNAME` to
retrieve the logged-in username.

In POSIX, `$LOGNAME` is a mandatory variable, while `$USER` isn't, and
on some systems it may not be set. Without this value, tests were unable
to provide the correct username when logging into the SSH server running
under the active user's session.

Reported-by: Nicolas Mora
Suggested-by: Nicolas Mora
Ref: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1056348
Fixes #1240
Closes #1241

### DIFF
--- a/tests/session_fixture.c
+++ b/tests/session_fixture.c
@@ -458,10 +458,13 @@ int test_auth_pubkey(LIBSSH2_SESSION *session, int flags,
     /* Ignore our hard-wired Dockerfile user when not running under Docker */
     if(!openssh_fixture_have_docker() && strcmp(username, "libssh2") == 0) {
         username = getenv("USER");
+        if(!username) {
 #ifdef _WIN32
-        if(!username)
             username = getenv("USERNAME");
+#else
+            username = getenv("LOGNAME");
 #endif
+        }
     }
 
     userauth_list = libssh2_userauth_list(session, username,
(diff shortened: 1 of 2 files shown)


steve-mcintyre:
Thanks for the quick turnaround!

Steve (initial bug reporter in Debian)

PR #1251, opened 2023-11-26
ci: new BSD actions flakey? try vmactions

vszakats:


vszakats:
They appear stable now.

PR #1252, opened 2023-11-26
ci: use `pkg_add` with NetBSD

vszakats:
"More native" and works with `vmactions/netbsd-vm` as well.

Ref: https://www.netbsd.org/docs/pkgsrc/using.html#installing-binary-packages

Follow-up to 65c7a7a55af037bcc9ee67c2f69ecdb9109ecbb1

Closes #1252


vszakats:
Nope, `pkg_add` missing here.

PR #1258, opened 2023-12-01
example: Replace remaining libssh2_scp_recv with libssh2_scp_recv2 in…

rmsh1216:
… a scp example

libssh2_scp_recv is deprecated and has been replaced by libssh2_scp_recv2 in prior commit.


vszakats:
Thank you!

Issue #1259, opened 2023-12-01
Use LIBSSH2_NO_DEPRECATED to mark DEPRECATED APIs. 


rmsh1216:
**Describe the bug**
I found out that some of the APIs in  libssh2 source code have been deprecated for over 10 years, which are too old. Use LIBSSH2_NO_DEPRECATED to mark DEPRECATED APIs.

PR #1260, opened 2023-12-01
Use LIBSSH2_NO_DEPRECATED to mark DEPRECATED APIs.

rmsh1216:
The following APIs have been deprecated for over 10 years, and it's time to delete them. 
libssh2_session_startup()
libssh2_banner_set()
libssh2_channel_receive_window_adjust()
libssh2_channel_handle_extended_data()
libssh2_scp_recv()

Fixes  https://github.com/libssh2/libssh2/issues/1259

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -242,6 +242,19 @@ AS_HELP_STRING([--disable-hidden-symbols],[Leave all symbols with default visibi
        AC_MSG_RESULT(no)
 )
 
+# Build without deprecated APIs?
+AC_ARG_ENABLE([deprecated-apis],
+  [AS_HELP_STRING([--disable-deprecated-apis], [Build without deprecated APIs @<:@default=no@:>@])],
+  [case "$enableval" in
+   *)
+     disable_deprecated_apis=yes
+     CFLAGS="$CFLAGS -DLIBSSH2_NO_DEPRECATED"
+     ;;
+esac],
+  [disable_deprecated_apis="no"])
+AC_MSG_RESULT($disable_deprecated_apis)
+AM_CONDITIONAL([NO_DEPRECATED_APIS], [test "x$disable_deprecated_apis" = xyes])
+
 # Build tests?
 AC_ARG_ENABLE([tests],
   [AS_HELP_STRING([--disable-tests], [Disable tests @<:@default=enabled@:>@])],
--- a/include/libssh2.h
+++ b/include/libssh2.h
@@ -617,10 +617,12 @@ LIBSSH2_API void *libssh2_session_callback_set(LIBSSH2_SESSION *session,
                                                int cbtype, void *callback);
 LIBSSH2_API int libssh2_session_banner_set(LIBSSH2_SESSION *session,
                                            const char *banner);
+#ifndef LIBSSH2_NO_DEPRECATED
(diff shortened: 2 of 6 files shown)


dfandrich:
This is going to require an SONAME bump, and I not convinced that this removal
is worthy enough to overcome the hardship that will cause users.

dfandrich:
Bumping the SONAME can be a big pain for people (not to mention removing APIs).
We have to balance that cost with the cost of keeping this old code. I haven't
heard how it's "necessary" to remove them.

willco007:
We've talked about an so name change in #1220, perhaps we could roll these to changes together.

rmsh1216:
Bumping the SONAME can be a big pain for people (not to mention removing APIs).
We have to balance that cost with the cost of keeping this old code. I haven't heard how it's "necessary" to remove them.
First of all, deleting APIs can be painful for users, but not unacceptable, especially since we provide better alternatives. Users can choose both old versions and new APIs, which are both not difficult for users.
Second, we don't delete old APIs immediately after adding new APIs, but after more than 10 years. libssh2 has been alerting users for more than 10 years. In the meantime, as many as 16 versions of libssh2 have been released. This period is long enough for the user to accept the impact of the change. It is certainly need to balance that cost with the cost of keeping the old code, and libssh2 has been doing that for over 10 years. However, libssh2 can't do this forever, because it's going to cost more and more to keep more and more old code. 
Also, libssh2 has historically removed some of the old APIs. Not only libssh2, but other packages also have APIs updates and remove old APIs.

vszakats:
Besides straight deletion, we can also:
1. mark functions deprecated so that they trigger compiler warnings when used.
2. guard them with `LIBSSH2_NO_DEPRECATED` (or similar) macro, to allow building libssh2 without.
3. do a combination of these.

@rmsh1216: An SO bump will certainly take some time, can you rework
this patch to not delete anything just yet, but mark these deprecated
functions with `#ifndef LIBSSH2_NO_DEPRECATED`?

This would allow us to merge now and finalize this decision as we
go with the rest of questions surrounding an SO bump. It allows testing
dependent code without these function right away. And also lets us
maintain these functions without introducing merge conflicts with this
PR in the meantime.

rmsh1216:
> Besides straight deletion, we can also:
> 
> 1. mark functions deprecated so that they trigger compiler warnings when used.
> 2. guard them with `LIBSSH2_NO_DEPRECATED` (or similar) macro, to allow building libssh2 without.
> 3. do a combination of these.
> 
> @rmsh1216: An SO bump will certainly take some time, can you rework this patch to not delete them just yet, but instead mark them deprecated with `#ifndef LIBSSH2_NO_DEPRECATED`?
> 
> This would allow us to merge now and finalize this decision as we go with the rest of questions surrounding an SO bump. It also allows testing dependent code without these function right away. And also lets us maintain these functions without introducing merge conflicts with this PR, in the meantime.

Okay, I'll redo the patch and use #ifndef LIBSSH2_NO_DEPRECATED to mark them as deprecated instead of removing them.
I still believe that libssh2 should have a plan to remove deprecated code. For example, inform users in RELEASE-NOTES of a future version that some deprecated codes will be deleted in next version, and record these incompatible changes in RELEASE-NOTES and Changelog when next version is released.

PR #1291, opened 2023-12-18
src: add 'strict KEX' to fix CVE-2023-48795 "Terrapin Attack"

### LINKED ISSUE #1290: Add support for "strict KEX" protocol extension
**Describe the bug**
libssh2 appears to be susceptible to "Terrapin attack" where MiTM attacker is able to tamper with the SSH early protocol exchange. https://terrapin-attack.com/

**Expected behavior**
libssh2 should preferably include support for "strict KEX" extension as defined in section 1.9 in https://github.com/openssh/openssh-portable/blob/8241b9c0529228b4b86d88b1a6076fb9f97e4a99/PROTOCOL#L140C1-L140C45

vszakats:
Refs:
https://terrapin-attack.com/
https://seclists.org/oss-sec/2023/q4/292
https://osv.dev/list?ecosystem=&q=CVE-2023-48795
https://github.com/advisories/GHSA-45x7-px36-x8w8
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-48795

Fixes #1290
Closes #1291

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -3032,6 +3032,13 @@ kex_method_extension_negotiation = {
     0,
 };
 
+static const LIBSSH2_KEX_METHOD
+kex_method_strict_client_extension = {
+    "kex-strict-c-v00@openssh.com",
+    NULL,
+    0,
+};
+
 static const LIBSSH2_KEX_METHOD *libssh2_kex_methods[] = {
 #if LIBSSH2_ED25519
     &kex_method_ssh_curve25519_sha256,
@@ -3050,6 +3057,7 @@ static const LIBSSH2_KEX_METHOD *libssh2_kex_methods[] = {
     &kex_method_diffie_helman_group1_sha1,
     &kex_method_diffie_helman_group_exchange_sha1,
     &kex_method_extension_negotiation,
+    &kex_method_strict_client_extension,
     NULL
 };
 
@@ -3302,13 +3310,13 @@ static int kexinit(LIBSSH2_SESSION * session)
     return 0;
 }
 
-/* kex_agree_instr
+/* _libssh2_kex_agree_instr
  * Kex specific variant of strstr()
  * Needle must be preceded by BOL or ',', and followed by ',' or EOL
  */
-static unsigned char *
-kex_agree_instr(unsigned char *haystack, size_t haystack_len,
-                const unsigned char *needle, size_t needle_len)
+unsigned char *
+_libssh2_kex_agree_instr(unsigned char *haystack, size_t haystack_len,
+                         const unsigned char *needle, size_t needle_len)
 {
     unsigned char *s;
     unsigned char *end_haystack;
@@ -3393,7 +3401,7 @@ static int kex_agree_hostkey(LIBSSH2_SESSION * session,
         while(s && *s) {
             unsigned char *p = (unsigned char *) strchr((char *) s, ',');
             size_t method_len = (p ? (size_t)(p - s) : strlen((char *) s));
-            if(kex_agree_instr(hostkey, hostkey_len, s, method_len)) {
+            if(_libssh2_kex_agree_instr(hostkey, hostkey_len, s, method_len)) {
                 const LIBSSH2_HOSTKEY_METHOD *method =
                     (const LIBSSH2_HOSTKEY_METHOD *)
                     kex_get_method_by_name((char *) s, method_len,
@@ -3427,9 +3435,9 @@ static int kex_agree_hostkey(LIBSSH2_SESSION * session,
     }
 
     while(hostkeyp && (*hostkeyp) && (*hostkeyp)->name) {
-        s = kex_agree_instr(hostkey, hostkey_len,
-                            (unsigned char *) (*hostkeyp)->name,
-                            strlen((*hostkeyp)->name));
+        s = _libssh2_kex_agree_instr(hostkey, hostkey_len,
+                                     (unsigned char *) (*hostkeyp)->name,
+                                     strlen((*hostkeyp)->name));
         if(s) {
             /* So far so good, but does it suit our purposes? (Encrypting vs
                Signing) */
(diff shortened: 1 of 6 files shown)


vszakats:
@MichaelBuckley The fuzzer is reporting a leak, would you mind taking a look?:
```
=================================================================
==32==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1 byte(s) in 1 object(s) allocated from:
    #0 0x5317c6 in malloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:69:3
    #1 0x570290 in libssh2_default_alloc /src/libssh2/src/session.c:77:12
    #2 0x57fa3d in _libssh2_transport_read /src/libssh2/src/transport.c:592:26
    #3 0x604eae in _libssh2_packet_require /src/libssh2/src/packet.c:1493:15
    #4 0x5bdea3 in _libssh2_kex_exchange /src/libssh2/src/kex.c:3901:17
    #5 0x5721b4 in session_startup /src/libssh2/src/session.c:772:14
    #6 0x5717bd in libssh2_session_handshake /src/libssh2/src/session.c:863:5
    #7 0x56edda in LLVMFuzzerTestOneInput /src/libssh2/tests/ossfuzz/ssh2_client_fuzzer.cc:67:6
    #8 0x440633 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #9 0x43fe1a in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #10 0x441c84 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:826:7
    #11 0x441eb9 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:857:3
    #12 0x43151f in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912:6
    #13 0x45ab72 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #14 0x7f756cf95082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: eebe5d5f4b608b8a53ec446b63981bba373ca0ca)

DEDUP_TOKEN: malloc--libssh2_default_alloc--_libssh2_transport_read
SUMMARY: AddressSanitizer: 1 byte(s) leaked in 1 allocation(s).
```
https://github.com/libssh2/libssh2/actions/runs/7254018903?pr=1291

MichaelBuckley:
> @MichaelBuckley The fuzzer is reporting a leak, would you mind taking a look?:

@vszakats My initial read is that the fuzzer is likely triggering the new behavior to disconnect after detecting the potential MITM attack, and the allocation in p->payload in _libssh2_transport_read isn't getting freed.

That would require the fuzzer to put libssh2 into strict key exchange mode though. I'm not familar with how the fuzzer works. Does it connect to a real OpenSSH server?

Otherwise, if we're not in strict KEX mode, I don't think anything has changed here that would affect this. p->payload is allocated in _libssh2_transport_read.

So assuming we're in strict KEX mode:

Looking at where p->packed is freed, it's freed once in decrypt(), twice in fullpacket(), and twice in _libssh2_transport_read(). fullpacket() is only called from _libssh2_transport_read(), and decrypt is only called from _libssh2_transport_read() and fullpacket(), so essentially, allocation and freeing is all done in _libssh2_transport_read().

In other words, when we return from _libssh2_transport_read(), p->payload should be freed, except when we're returning EAGAIN.

fullpacket() calls _libssh2_packet_add(), which, with this change, can now return LIBSSH2_ERROR_BUFFER_TOO_SMALL and LIBSSH2_ERROR_SOCKET_DISCONNECT in strict KEX mode. But when checking the return code of fullpacket(), _libssh2_transport_read() doesn't do any cleanup. This is because fullpacket() releases p->payload in certain cases, so we'd risk a double free by freeing it again here.

So I suspect we might need to check for p->payload in fullpacket() after calling _libssh2_packet_add, and free it for a non-EGAIN error.

Does that sound reasonable?

vszakats:
@MichaelBuckley It does to me, yes. Thanks for looking into it.

Can you prepare a patch, or push it into this branch?

[ Regarding the fuzzer, I haven't looked at it yet. At a glance it seems to feed test inputs via sockets, but I may be totally off. ]

MichaelBuckley:
It doesn't look like I have permission to push to this branch, but on closer reading, the change should just be one line. No need to check if p->packet exists. Here's a patch.

[packet_free.txt](https://github.com/libssh2/libssh2/files/13709360/packet_free.txt)

vszakats:
It is odd, `Allow edits and access to secrets by maintainers` is ticked – I wonder what else might be needed for this to work. I've pushed it now, no worries.

willco007:
@vszakats @MichaelBuckley isn't flagged as a maintainer. He probably could be. :)

MichaelBuckley:
Looks like I was too optimistic. Desite the comment at the top of _libssh2_packet_add(), it looks like we need to do the freeing in that function, not fullpacket(). I'll prep another change. Sorry about that.

vszakats:
Thanks @MichaelBuckley, looking good! May I merge?

MichaelBuckley:
Gave it one more read through, and I think it's good to merge, as long as no one else spotted anything.

jorgensigvardsson:
It looks like this change is "always on", and no special option is needed to be set to turn it on. Am I right?

Thanks for all the hard work on libssh2 by the way!

MichaelBuckley:
This change is always on, but dependent on both the client and the server
implementing the protocol extension. If libssh2 connects to a server that
does not implement the protocol extension, libssh2 will not use the
mitigation.

But in terms of code, there is nothing special you need to do to enable the
mitigation.

PR #1292, opened 2023-12-19
ci/gha: review/fixup auto-cancel settings

vszakats:
- use the group expression from `reuse.yml` (via curl).
- add auto-cancel for `ci` and `cifuzz`.
- add auto-cancel to `appveyor_docker`. I'm just guessing here.
  The hope is that it fixes AppVeyor CI runs when re-pushing a PR.
  This frequently caused the freshly pushed session to fail waiting for
  a connection.
- sync group expression in `appveyor_status` with `reuse`.

Closes #1292

### DIFF
--- a/.github/workflows/appveyor_docker.yml
+++ b/.github/workflows/appveyor_docker.yml
@@ -42,6 +42,10 @@ on:
       ssh_privkey:
         required: true
 
+concurrency:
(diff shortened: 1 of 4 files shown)


vszakats:
Well, that doesn't work as excpted. When pushing to master twice, the previous GHA session isn't cancelled.

PR #1297, opened 2023-12-22
mac: handle low-level errors

vszakats:
- update low-level hmac functions from macros to functions.
- libgcrypt: propagate low-level hmac errors.
- libgcrypt: add error checks for hmac calls.
- os400qc3: add error checks, propagate them.
  Assisted-by: Patrick Monnerat
- mbedtls: fix propagating low-level hmac errors.
- wincng: fix propagating low-level hmac errors.
- mac: verify success of low-level hmac functions.
- knownhost: verify success of low-level hmac functions.
- transport: verify success of MAC hash call.
- minor type cleanup in wincng.
- delete unused ripemd wrapper in wincng.
- delete unused SHA384 wrapper in mbedtls.

Reported-by: Paul Howarth
Closes #1297

### DIFF
--- a/src/knownhost.c
+++ b/src/knownhost.c
@@ -421,7 +421,8 @@ knownhost_check(LIBSSH2_KNOWNHOSTS *hosts,
                     */
                     unsigned char hash[SHA_DIGEST_LENGTH];
                     libssh2_hmac_ctx ctx;
-                    _libssh2_hmac_ctx_init(&ctx);
+                    if(!_libssh2_hmac_ctx_init(&ctx))
+                        break;
 
                     if(SHA_DIGEST_LENGTH != node->name_len) {
                         /* the name hash length must be the sha1 size or
--- a/src/knownhost.c
+++ b/src/knownhost.c
@@ -428,9 +428,14 @@ knownhost_check(LIBSSH2_KNOWNHOSTS *hosts,
                            we can't match it */
                         break;
                     }
-                    _libssh2_hmac_sha1_init(&ctx, node->salt, node->salt_len);
-                    _libssh2_hmac_update(&ctx, host, strlen(host));
-                    _libssh2_hmac_final(&ctx, hash);
+                    if(!_libssh2_hmac_sha1_init(&ctx,
+                                                node->salt, node->salt_len))
+                        break;
+                    if(!_libssh2_hmac_update(&ctx, host, strlen(host)) ||
+                       !_libssh2_hmac_final(&ctx, hash)) {
+                        _libssh2_hmac_cleanup(&ctx);
+                        break;
+                    }
                     _libssh2_hmac_cleanup(&ctx);
 
                     if(!memcmp(hash, node->name, SHA_DIGEST_LENGTH))
--- a/src/knownhost.c
+++ b/src/knownhost.c
@@ -430,10 +430,10 @@ knownhost_check(LIBSSH2_KNOWNHOSTS *hosts,
                     }
                     _libssh2_hmac_sha1_init(&ctx, (unsigned char *)node->salt,
                                             node->salt_len);
-                    _libssh2_hmac_update(ctx, (unsigned char *)host,
+                    _libssh2_hmac_update(&ctx, (unsigned char *)host,
                                          strlen(host));
-                    _libssh2_hmac_final(ctx, hash);
-                    _libssh2_hmac_cleanup(ctx);
+                    _libssh2_hmac_final(&ctx, hash);
+                    _libssh2_hmac_cleanup(&ctx);
 
                     if(!memcmp(hash, node->name, SHA_DIGEST_LENGTH))
                         /* this is a node we're interested in */
--- a/src/knownhost.c
+++ b/src/knownhost.c
@@ -421,7 +421,7 @@ knownhost_check(LIBSSH2_KNOWNHOSTS *hosts,
                     */
                     unsigned char hash[SHA_DIGEST_LENGTH];
                     libssh2_hmac_ctx ctx;
-                    libssh2_hmac_ctx_init(ctx);
+                    libssh2_hmac_ctx_init(&ctx);
 
                     if(SHA_DIGEST_LENGTH != node->name_len) {
                         /* the name hash length must be the sha1 size or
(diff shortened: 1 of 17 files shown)


vszakats:
@monnerat: I had to touch OS/400 code in this patch, and even went one notch further and made an attempt to add error checking for low-level MAC functions. But, this wasn't tested at all. Would you mind reviewing / testing it?

Also added a few FIXMEs for missing errors checks and a possible memory leak in the `pbkdf2()` function.

src/os400qc3.c
```diff
@@ -1421,6 +1491,7 @@ pbkdf2(LIBSSH2_SESSION *session, char **dk, const unsigned char *passphrase,
         pkcs5->hashlen;
     if(t > 0xFFFFFFFF)
         return -1;
+    /* FIXME: mac never freed? */
```
monnerat, line 1494: No, because `alloca()` allocates on stack an thus storage is automatically released at function return, like if it were an automatic local variable.

src/os400qc3.c
```diff
@@ -1003,51 +1003,121 @@ libssh2_os400qc3_hash(const unsigned char *message, unsigned long len,
...
+    return libssh2_os400qc3_hmac_init(ctx, Qc3_SHA512,                  \
+                                      SHA512_DIGEST_LENGTH,             \
+                                      key, keylen);
+}
+
+int _libssh2_hmac_update(libssh2_hmac_ctx *ctx,
+                         const void *data, size_t datalen)
+{
+    return libssh2_os400qc3_hmac_update(ctx,                            \
+                                        data, datalen);
```
monnerat, line 1108:
```
/home/MONNERAT/libssh2/src/os400qc3.c, 1108.41: CZM0280(30) Function argument assignment between types "unsigned char*"
    and "const void*" is not allowed.
```

monnerat:
Many thanks for your attempt on os400. I left 2 remarks in the code that, when fixed, make libssh2 pass the compilation.
However an execution test (example/sftp.c) stops with `Failure establishing SSH session: -52`: mac error. I don't know yet why and I have to investigate in the debugger.

I think the easiest way  of proceeding is you commit this when it's satisfactory for you and alert me on this. Then I'll complete/fix the os400 build and submit a PR for it.

src/os400qc3.c
vszakats: Thanks. Pushed a fix, though without testing it will be a hit or miss.

Also added error handling to `pbkdf2()`.

vszakats:
Thanks @monnerat! I committed the attempted fixes, add more error checks and optimized out two internal wrappers. Let me know if it made things better/worse. Also thanks for looking into the new test failure.

src/os400qc3.c
```diff
@@ -1104,14 +1074,28 @@ int _libssh2_hmac_sha512_init(libssh2_hmac_ctx *ctx,
 int _libssh2_hmac_update(libssh2_hmac_ctx *ctx,
                          const void *data, size_t datalen)
 {
-    return libssh2_os400qc3_hmac_update(ctx,                            \
-                                        data, datalen);
+    char dummy[64];
+    Qus_EC_t errcode;
```
monnerat, line 1078:
```
/home/MONNERAT/libssh2/src/os400qc3.c, 1082.38: CZM0045(30) Undeclared identifier len.
```
May be insert `int len = datalen;` here.

src/os400qc3.c
```diff
@@ -1003,51 +1003,105 @@ libssh2_os400qc3_hash(const unsigned char *message, unsigned long len,
...
         if(!lkey)
-            return;
+            return 0;
         memcpy(lkey, (char *) key, keylen);
         memset(lkey + keylen, 0, minkeylen - keylen);
         key = (void *) lkey;
         keylen = minkeylen;
     }
-    libssh2_os400qc3_hash_init(&ctx->hash, algo);
+    else
```
monnerat, line 1023: This else clause is wrong: the test above is for padding, not for length overflow. The case `keylen >= minkeylen` (32 < 32) then causes my current failure.

monnerat:
> I committed the attempted fixes, add more error checks and optimized out two internal wrappers. Let me know if it made things better/worse. Also thanks for looking into the new test failure.

After fixing locally the 2 last remarks, this compiles fine and example/sftp.c runs properly. Thanks a lot.

src/os400qc3.c
vszakats: Oops, pushed a fixed for this!

And for the datalen/len copy-paste mistake.

src/os400qc3.c
```diff
@@ -1079,7 +1077,7 @@ int _libssh2_hmac_update(libssh2_hmac_ctx *ctx,
 
     ctx->hash.Final_Op_Flag = Qc3_Continue;
     set_EC_length(errcode, sizeof(errcode));
-    Qc3CalculateHMAC((char *) data, &len, Qc3_Data, (char *) &ctx->hash,
+    Qc3CalculateHMAC((char *) data, &datalen, Qc3_Data, (char *) &ctx->hash,
```
monnerat, line 1080: No, I already tried it. It causes a type mismatch error .-(

src/os400qc3.c
vszakats: Synced the name in the function body. I hope it's OK.

src/os400qc3.c
vszakats: Indeed! It's also one of those places where we must truncate the size (from `size_t` to `int` here). Strictly speaking these should be checked and bailing out or making an internal loop I think. (Won't be tackling these in this commit though.)

monnerat:
> Indeed! It's also one of those places where we must truncate the size (from size_t to int here). Strictly speaking these should be checked and bailing out or making an internal loop I think. (Won't be tackling these in this commit though.)

For portability, I agree ! But I don't think this would be used on any other platform ever!
Os400 system header files for proprietary prototypes have always be very "minimalist", in the sense they do net emphasize signedness and const: this forces you to cast very often.
And the compiler is not very clever about `void *` being compatible with any non-void pointer !

src/knownhost.c
```diff
@@ -421,19 +421,22 @@ knownhost_check(LIBSSH2_KNOWNHOSTS *hosts,
                     */
                     unsigned char hash[SHA_DIGEST_LENGTH];
                     libssh2_hmac_ctx ctx;
-                    libssh2_hmac_ctx_init(ctx);
+                    _libssh2_hmac_ctx_init(&ctx);
```
MichaelBuckley, line 424: Given that this function has just changd to return int instead of void, should we check the return value here?

MichaelBuckley (commented):
Looks mostly good save for one comment. I don't have experience with some of these backends, but from reading their documentation, the code appears to be correct.

src/knownhost.c
vszakats: Thank you, good spot, we should indeed! I'v pushed an update.

MichaelBuckley (approved):

PR #1301, opened 2023-12-27
src: check hash init success

vszakats:
Before this patch, SHA2 and SHA1 init function results were cast to
`void`. This patch makes sure to verify these values.

Also:
- exclude an `assert(0)` from release builds in `_libssh2_sha_algo_ctx_init()`.
  (return error instead)
- fix indentation / whitespace

Closes #1301

### DIFF
--- a/src/bcrypt_pbkdf.c
+++ b/src/bcrypt_pbkdf.c
@@ -127,7 +127,10 @@ bcrypt_pbkdf(const char *pass, size_t passlen, const uint8_t *salt,
     memcpy(countsalt, salt, saltlen);
 
     /* collapse password */
-    (void)libssh2_sha512_init(&ctx);
+    if(!libssh2_sha512_init(&ctx)) {
+        free(countsalt);
+        return -1;
+    }
     libssh2_sha512_update(ctx, pass, passlen);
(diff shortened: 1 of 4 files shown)


src/bcrypt_pbkdf.c
```diff
@@ -127,7 +127,10 @@ bcrypt_pbkdf(const char *pass, size_t passlen, const uint8_t *salt,
     memcpy(countsalt, salt, saltlen);
 
     /* collapse password */
-    (void)libssh2_sha512_init(&ctx);
+    if(!libssh2_sha512_init(&ctx)) {
+        free(countsalt);
+        return -1;
+    }
     libssh2_sha512_update(ctx, pass, passlen);
     libssh2_sha512_final(ctx, sha2pass);
```
MichaelBuckley, line 135: Perhaps this is a separate task, but should we not also change libssh2_sha*_update and libssh2_sha*_final to return int and check their return values?

vszakats: That would be useful, I agree. I'd make this a separate PR when concluding the pending ones just to keep fewer balls in the air.

MichaelBuckley (approved):

PR #1304, opened 2024-01-11
openssl: fix cppcheck found NULL dereferences

doorsdown:
Toying with cppcheck and it found 4 cases of a null pointer being dereferenced in the code. There were 2 functions that called the respective sections of code that set NULL for method_len and pubkeydata_len causing the dereference when they get assigned a value without being checked.

---

https://github.com/libssh2/libssh2/pull/1304/files?w=1

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -1305,11 +1305,11 @@ gen_publickey_from_rsa_evp(LIBSSH2_SESSION *session,
     memcpy(method_buf, "ssh-rsa", 7);
     *method         = method_buf;
     if(method_len) {
-	*method_len = 7;
+        *method_len = 7;
     }
     *pubkeydata     = key;
     if(pubkeydata_len) {
-	*pubkeydata_len = key_len;
+        *pubkeydata_len = key_len;
     }
     return 0;
 
@@ -1748,11 +1748,11 @@ gen_publickey_from_dsa_evp(LIBSSH2_SESSION *session,
     memcpy(method_buf, "ssh-dss", 7);
     *method         = method_buf;
     if(method_len) {
-	*method_len = 7;
+        *method_len = 7;
     }
     *pubkeydata     = key;
     if(pubkeydata_len) {
-	*pubkeydata_len = key_len;
+        *pubkeydata_len = key_len;
     }
     return 0;
 
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -1303,11 +1303,11 @@ gen_publickey_from_rsa_evp(LIBSSH2_SESSION *session,
 #endif
 
     memcpy(method_buf, "ssh-rsa", 7);
-    *method         = method_buf;
+    *method = method_buf;
     if(method_len) {
         *method_len = 7;
     }
-    *pubkeydata     = key;
+    *pubkeydata = key;
     if(pubkeydata_len) {
         *pubkeydata_len = key_len;
     }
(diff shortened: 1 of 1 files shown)


vszakats:
We prefer indenting the `if`s. Can you push an update with this style?:
```c
if(method_len) {
  *method_len = 7;
}
```

vszakats:
Looking a level up, is there any valid reason to call these functions to return binary data _without their lengths_?
I wonder if it'd be a better fix to check for these conditions in the callers and return an error as early as possible.

willco007:
@vszakats `gen_publickey_from_rsa_openssh_priv_data()` can pass in NULL method with a 0 as method_len. Seems the simplest solution would just be to handle each null check individually.

doorsdown:
> We prefer indenting the `if`s. Can you push an update with this style?:
> 
> ```c
> if(method_len) {
>   *method_len = 7;
> }
> ```

Pushed to branch, checks are running again now

src/openssl.c
```diff
@@ -1304,9 +1304,13 @@ gen_publickey_from_rsa_evp(LIBSSH2_SESSION *session,
 
     memcpy(method_buf, "ssh-rsa", 7);
     *method         = method_buf;
```
vszakats, line 1306:
```suggestion
    *method = method_buf;
```

src/openssl.c
```diff
@@ -1304,9 +1304,13 @@ gen_publickey_from_rsa_evp(LIBSSH2_SESSION *session,
...
-    *method_len     = 7;
+    if(method_len) {
+        *method_len = 7;
+    }
     *pubkeydata     = key;
```
vszakats, line 1310:
```suggestion
    *pubkeydata = key;
```

src/openssl.c
```diff
@@ -1743,9 +1747,13 @@ gen_publickey_from_dsa_evp(LIBSSH2_SESSION *session,
 
     memcpy(method_buf, "ssh-dss", 7);
     *method         = method_buf;
```
vszakats, line 1749:
```suggestion
    *method = method_buf;
```

src/openssl.c
```diff
@@ -1743,9 +1747,13 @@ gen_publickey_from_dsa_evp(LIBSSH2_SESSION *session,
...
-    *method_len     = 7;
+    if(method_len) {
+        *method_len = 7;
+    }
     *pubkeydata     = key;
```
vszakats, line 1753:
```suggestion
    *pubkeydata = key;
```

doorsdown:
Suggested clean up has been completed

vszakats:
Do we have a reason to not add the same NULL checks for sister functions `en_publickey_from_ec_evp` and `en_publickey_from_ed_evp`? They are called the same (or at least very similar) ways, including NULLs for their length pointers.

vszakats:
@willco007 Both data & len ptr being NULL is a valid case indeed, but in this case it will be caught one level up. What's bothering me is that data-ptr and len-ptr are now checked at different levels in the code.

doorsdown:
> Do we have a reason to not add the same NULL checks for sister functions `en_publickey_from_ec_evp` and `en_publickey_from_ed_evp`? They are called the same (or at least very similar) ways, including NULLs for their length pointers.

Added the checks to these methods also

vszakats:
Thank you @doorsdown, merged now!

Issue #652, opened 2021-12-21
Confusing `libssh2_userauth_publickey_fromfile()` documentation


jnqnfe:
I'm finding the documentation (man page) for `libssh2_userauth_publickey_fromfile()` confusing.

Firstly, it has parameters for both a public key file and a private key file. Why? My experience with SSH involves the server having a copy of my public key, so why would it need to be provided here? I am also aware of the 'certificate' model in which my public key would be signed by a CA, in which case I guess that the server would not necessarily have a copy, I can imagine instead it being sent to the server and the server trusting this as authenticating who you are via trusting the CA signature. Is this the reason for the public key param, to pass to the server a local CA signed copy of your public key? In which case do you thus only need to use this param for the 'certificate' based model?

I did wonder at one point whether the public key file parameter might be the public key of the server you want to connect to, but I think not. Perhaps you could understand how a user of this API could wonder this though?

Secondly, the title of the manpage states `authenticate a session with a public key, read from a file`, while at the end of the description there's `Attempt public key authentication using a PEM encoded private key file stored on disk`. Perhaps the title should say something more like `authenticate a session with public-key cryptography, using a key or key-pair read from a file` to indicate that "public key" in that case refers to public-key cryptography to avoid confusion with the public key in a key pair and thus the public key file param? The second case should explain how the public key file comes into things.

Could you please explain, and clarify the documentation.

vszakats:
@jnqnfe We're happy to accept a documentation patch for this.

jnqnfe:
@vszakats Sorry but I reported this 16 months ago, I've long since completely forgotten what on earth I was even trying to use this for, and re-reading this it looks like I did not have anywhere near a good enough grasp of what is what to be in a suitable position to rewrite it, otherwise I expect I would have already contributed a patch. Someone who knows what they're talking about will have to improve it. :/

PR #1308, opened 2024-01-19
Update the man page for libssh2_userauth_publickey_fromfile_ex to eliminate confusion 

doorsdown:
Based on #652 I have clarified the man page for libssh2_userauth_publickey_fromfile_ex to show that it is not a public key but rather PKI and what is required to use it. Also added a missing possible error return value: LIBSSH2_ERROR_FILE

### DIFF
--- a/docs/libssh2_userauth_publickey_fromfile_ex.3
+++ b/docs/libssh2_userauth_publickey_fromfile_ex.3
@@ -2,7 +2,7 @@
 .\" SPDX-License-Identifier: BSD-3-Clause
 .TH libssh2_userauth_publickey_fromfile_ex 3 "1 Jun 2007" "libssh2 0.15" "libssh2"
 .SH NAME
-libssh2_userauth_publickey_fromfile_ex - authenticate a session based on local public key cryptography
+libssh2_userauth_publickey_fromfile_ex - reading a local file to authenticate a session with public key cryptography
 .SH SYNOPSIS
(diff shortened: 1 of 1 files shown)


docs/libssh2_userauth_publickey_fromfile_ex.3
```diff
@@ -2,7 +2,7 @@
 .\" SPDX-License-Identifier: BSD-3-Clause
 .TH libssh2_userauth_publickey_fromfile_ex 3 "1 Jun 2007" "libssh2 0.15" "libssh2"
 .SH NAME
-libssh2_userauth_publickey_fromfile_ex - authenticate a session with a public key, read from a file
+libssh2_userauth_publickey_fromfile_ex - authenticate a session based on local public key cryptography
```
vszakats, line 5: This is true, but isn't this tagline true for all `libssh2_userauth_publickey*` functions?
The 'read from a file' differenciated this from the rest of similar calls, thus having this 
it in the title seems useful. Also, to me, 'local public key cryptography' rings 
unfamiliar; the common wording being 'public key authentication'.

doorsdown: I reworded the line to be more clear about using a local file

vszakats:
Thank you, I merged this with a slightly different wording in #1347. Let us know if it needs more improvement.

PR #1309, opened 2024-01-19
os400: maintain up to date

monnerat:
- Handle MD5 conditionals in os400qc3.
- Check for errors in os400qc3 pbkdf1.
- Implement an optional build options override file.
- Sync ILE/RPG copy files with current C header files.
- Allow a null session within a string conversion cache.
- Add an ILE/RPG example.
- Adjust outdated copyrights in changed files.


vszakats:
Thank you!

PR #1310, opened 2024-01-19
wolfssl: enable debug logging in wolfSSL when compiled in

julek-wolfssl:


src/openssl.c
```diff
@@ -1101,6 +1101,9 @@ void _libssh2_openssl_crypto_init(void)
     ENGINE_register_all_complete();
 #endif
 #endif
+#ifdef DEBUG_WOLFSSL
```
vszakats, line 1104: Let's also guard it with `LIBSSH2_WOLFSSL`.

vszakats, line 1104:
```suggestion
#if defined(LIBSSH2_WOLFSSL) && defined(DEBUG_WOLFSSL)
```

julek-wolfssl: Added `LIBSSH2_WOLFSSL` to macro check.

vszakats: Thanks @julek-wolfssl! and merging.

PR #1314, opened 2024-01-24
Support RSA SHA2 cert-based authentication (rsa-sha2-512_cert and rsa-sha2-256_cert)

Tejaswikandula:
Replicating OpenSSH's behavior to handle RSA certificate authentication differently based on the remote server version. 
1. For OpenSSH versions >= 7.8, ascertain server's support for RSA Cert types by checking if the certificate's signature type is present in the 'server-sig-algs'.
2. For OpenSSH versions < 7.8, Set the "SSH_BUG_SIGTYPE" flag when the RSA key in question is a certificate to ignore "server-sig-algs" and only offer ssh-rsa signature algorithm for RSA certs.
This arises from the fact that OpenSSH versions up to 7.7 accept RSA-SHA2 keys but not RSA-SHA2 certificate types. Although OpenSSH <=7.7 includes RSA-SHA2 keys in the server-sig-algs, versions <=7.7 do not actually support RSA certs. Therefore, server sending RSA-SHA2 keys in server-sig-algs should not be interpreted as indicating support for RSA-SHA2 certs. So, server-sig-algs are ignored when the RSA key in question is a cert, and the remote server version is 7.7 or below.

Relevant sections of the OpenSSH source code:

[https://github.com/openssh/openssh-portable/blob/V_8_9_P1/sshconnect2.c#L1191-L1197
https://github.com/openssh/openssh-portable/blob/master/compat.c#L43
](url)

[sshconnect2.c](https://github.com/openssh/openssh-portable/blob/V_8_9_P1/sshconnect2.c#L1191-L1197)
[compat.c](https://github.com/openssh/openssh-portable/blob/master/compat.c#L43)


tests/gen_keys.sh
```diff
@@ -38,6 +38,7 @@ ssh-keygen -t rsa     -b 2048 -N ''                 -C 'key_rsa_openssh'
 ssh-keygen -t rsa     -b 2048 -N "${pw}"            -C 'key_rsa_aes256gcm'       -f 'key_rsa_aes256gcm'     -Z aes256-gcm@openssh.com
 ssh-keygen -t rsa     -b 4096 -N ''                 -C 'key_rsa_signed'          -f 'key_rsa_signed'
 ssh-keygen                    -I "${id}" -n "${pr}" -s 'openssh_server/ca_rsa'      'key_rsa_signed.pub'
+ssh-keygen -t rsa     -b 4096 -N ''                 -C 'key_rsa_sha2_256_signed' -f 'key_rsa_sha2_256_signed'
```
vszakats, line 41: Please also include the command that created `tests/key_rsa_sha2_256_signed-cert.pub`.

Tejaswikandula:
@willco007 Did changes to allow algo upgrades for SSH RSA certs if the server also enumerates the supported RSA cert based algs in server-sig-algs extension, fallback to default otherwise. Thoughts/Suggestions?

Tejaswikandula:
> @Tejaswikandula No, the only types it supports are `rsa-sha2-256` and `rsa-sha2-512`. At the time, these were the only upgrades OpenSSH supported, I'm not sure if OpenSSH now supports cert upgrades as well.

RFC 8332 mentions allowing for interoperable use of existing and new RSA keys with SHA-256 and SHA-512. Is this what you are referring to when you say "these were the only upgrades OpenSSH supported"?

While RFC 8332 doesn't explicitly discuss RSA certificates, RSA certs align with the hashing types used in RSA keys right? So, in case of certs, even though OpenSSH doesn't explicitly announce RSA certs in the server-sig-algs extension for flexible negotiation during key exchange process, can we infer that the support for RSA SHA-2 certificate-based authentications in newer OpenSSH versions indicates the potential for certificate upgrades?

willco007:
> RFC 8332 mentions allowing for interoperable use of existing and new RSA keys with SHA-256 and SHA-512. Is this what > you are referring to when you say "these were the only upgrades OpenSSH supported"?

Correct. 

> ...can we infer that the support for RSA SHA-2 certificate-based authentications in newer OpenSSH versions indicates the potential for certificate upgrades?

No, we can't infer this. When reviewing the OpenSSH server code while implementing this, it did not have RSA cert support.  New versions may have support, but that would need to be confirmed.

willco007:
This is what comes back from 9.4:

```Key Ex: Server to Client extension server-sig-algs: ssh-ed25519,sk-ssh-ed25519@openssh.com,ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,sk-ecdsa-sha2-nistp256@openssh.com,webauthn-sk-ecdsa-sha2-nistp256@openssh.com,ssh-dss,ssh-rsa,rsa-sha2-256,rsa-sha2-512```. I see no mention of cert support.

**Comment by Tejaswikandula** [comment]
> > RFC 8332 mentions allowing for interoperable use of existing and new RSA keys with SHA-256 and SHA-512. Is this what > you are referring to when you say "these were the only upgrades OpenSSH supported"?
> 
> Correct.
> 
> > ...can we infer that the support for RSA SHA-2 certificate-based authentications in newer OpenSSH versions indicates the potential for certificate upgrades?
> 
> No, we can't infer this. When reviewing the OpenSSH server code while implementing this, it did not have RSA cert support. New versions may have support, but that would need to be confirmed.

OpenSSH do support RSA certs from a long time and newer versions do also support RSA SHA2 certs. OpenSSH deprecated support for SHA1 certs along with the keys. Please refer to this PR - [https://github.com/openssh/openssh-portable/commit/2344750250247111a6c3c6a4fe84ed583a61cc11](url). I'm able to successfully test using certs against different OpenSSH versions

**Comment by willco007** [comment]
Yes, that is true, but I believe you're conflating `server-sig-algs` list and normal key auth. The `server-sig-algs` list is used to upgrade insecure keys while certs are used for normal key auth which is what you have implemented in this MR.

**Comment by Tejaswikandula** [comment]
@vszakats Requesting review again on this one. Also, the build is failing with docker error unrelated to my changes (probably flaky), can you help with restarting the build?

**Review by vszakats — APPROVED** [review]

**Comment by vszakats** [comment]
Thanks @Tejaswikandula, LGTM!

File: src/userauth.c [hunk 2]
```diff
@@ -1311,33 +1311,16 @@ size_t plain_method(char *method, size_t method_len)
...
-            return 0;
-        }
-
-        /* If the characters are not equal, compare them */
-        if(*version != *pattern) {
-            return *version < *pattern;
+    int major, minor;
+    if(sscanf(version, "OpenSSH_%d.%d", &major, &minor) == 2) {
+        if((major >= 1 && major <= 6)
+        || (major == 7 && minor >= 0 && minor <= 7)) {
```
vszakats, line 1319:
```suggestion
        if((major >= 1 && major <= 6) ||
           (major == 7 && minor >= 0 && minor <= 7)) {
```

src/openssl.c
```diff
@@ -5134,8 +5134,11 @@ _libssh2_supported_key_sign_algorithms(LIBSSH2_SESSION *session,
     (void)session;
 
 #if LIBSSH2_RSA_SHA2
-    if(key_method_len == 7 &&
-       memcmp(key_method, "ssh-rsa", key_method_len) == 0) {
+    if((key_method_len == 7 &&
+       memcmp(key_method, "ssh-rsa", key_method_len) == 0)
+       || (key_method_len == 28 &&
```
vszakats, line 5139:
```suggestion
       memcmp(key_method, "ssh-rsa", key_method_len) == 0) ||
       (key_method_len == 28 &&
```

src/userauth.c
```diff
@@ -1412,15 +1459,29 @@ _libssh2_key_sign_algorithm(LIBSSH2_SESSION *session,
...
     if(match) {
-        if(*key_method)
-            LIBSSH2_FREE(session, *key_method);
-
-        *key_method = LIBSSH2_ALLOC(session, match_len);
-        if(*key_method) {
-            memcpy(*key_method, match, match_len);
-            *key_method_len = match_len;
+        if(*key_method_len == 28 &&
+        memcmp(key_method, "ssh-rsa-cert-v01@openssh.com", *key_method_len)) {
```
vszakats, line 1463:
```suggestion
        if(*key_method_len == 28 &&
            memcmp(key_method, "ssh-rsa-cert-v01@openssh.com",
                   *key_method_len)) {
```

src/userauth.c
```diff
@@ -1345,6 +1375,23 @@ _libssh2_key_sign_algorithm(LIBSSH2_SESSION *session,
...
+       RSA certs */
+    remote_banner = libssh2_session_banner_get(session);
+    /* Extract version information from the banner */
+    remote_version_prefix = "OpenSSH_";
+    remote_version_start = strstr(remote_banner, remote_version_prefix);
+
+    if(remote_version_start) {
+        SSH_BUG_SIGTYPE = is_version_less_than_78(remote_version_start);
+        if(SSH_BUG_SIGTYPE && *key_method_len == 28 &&
+        memcmp(key_method, "ssh-rsa-cert-v01@openssh.com", *key_method_len)) {
```
vszakats, line 1390:
```suggestion
           memcmp(key_method, "ssh-rsa-cert-v01@openssh.com",
                  *key_method_len)) {
```

Tejaswikandula:
> Thanks @Tejaswikandula, LGTM!

@vszakats Thanks so much for the review. Need your help with restarting the build again, thanks!!

src/userauth.c
```diff
@@ -1299,6 +1308,21 @@ size_t plain_method(char *method, size_t method_len)
     return method_len;
 }
 
+/* Function to check if the given version is less than pattern (OpenSSH 7.8)
+ * Returns 1 if the version is less than OpenSSH_7.8, 0 otherwise
+ */
+static int is_version_less_than_78(const char *version)
+{
+    int major, minor;
+    if(sscanf(version, "OpenSSH_%d.%d", &major, &minor) == 2) {
```
vszakats, line 1317: `sccanf()` is considered slow and heavy, and it would be its only use in libssh2.
Would there be a way to replace it with e.g. `strtol()`? (we already use this one)

Tejaswikandula: @vszakats Removed sscanf. Made changes to just extract the required version char without using any library functions.

src/openssl.c
```diff
@@ -5134,8 +5134,11 @@ _libssh2_supported_key_sign_algorithms(LIBSSH2_SESSION *session,
...
+       memcmp(key_method, "ssh-rsa-cert-v01@openssh.com", key_method_len) == 0)
```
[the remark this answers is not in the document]
Tejaswikandula: Although there is no official documentation which directly mentions that RSA certs can be upgraded in the same way as regular RSA keys, OpenSSH itself has implemented this capability years ago to force use of RSA-SHA2 signatures with certificate keys as well. Please refer to the below PR for this.
[https://github.com/openssh/openssh-portable/commit/4ba0d54794814ec0de1ec87987d0c3b89379b436](url)

This is my understanding. Now, I know that server-sig-algs are different from normal key auth. All OpenSSH versions above 7.8 accept rsa-sha2-* certificates, even though they're not explicitly enumerated in the server-sig-algs list.For that matter, no RSA cert type keys are enumerated in the server-sig-algs list. Also, we are not just interpreting the existence of rsa-sha2 keys in the server-sig-algs configuration, as it also extends support to certificates. This is the rationale behind OpenSSH mandating RSA-SHA2 signatures only on versions >=7.8 and above knowing that the server supports RSA SHA2 certificates on these versions, and not earlier versions.

In essence, this approach is also driven by the server's capabilities knowing what the remote server supports, rather than solely relying on server-sig-algs.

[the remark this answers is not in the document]
willco007: Ok, that commit does show RSA1 certs being upgraded. Looks OK to me.

vszakats (approved):

vszakats:
Tests all green: #1350

vszakats:
Thank you @Tejaswikandula for this, also for your patience. To @willco007 for all the help.
Merged now!

src/userauth.c
```diff
@@ -1345,6 +1388,25 @@ _libssh2_key_sign_algorithm(LIBSSH2_SESSION *session,
         return rc;
     }
 
+    /* Set "SSH_BUG_SIGTYPE" flag when the remote server version is OpenSSH 7.7
+       or lower and when the RSA key in question is a certificate to ignore
+       "server-sig-algs" and only offer ssh-rsa signature algorithm for
+       RSA certs */
```
vszakats, line 1394: @Tejaswikandula Do you recall what this section of code is meant to do?

The comment (and linked openssh code) refers to `SSH_BUG_SIGTYPE` as a flag, but here it's a local variable that's set then checked right away, and never read again. Is this how it's intended?

Also, is the `memcmp()` call accurate? shouldn't it be compared to zero for a positive match? (there are two `memcmp()` calls like this)

src/userauth.c
```diff
@@ -1299,6 +1310,31 @@ size_t plain_method(char *method, size_t method_len)
...
+        return 0;
+
+    major = strtol(version, &endptr, 10);
+    if(!endptr || *endptr != '.')
+        return 0; /* Not a valid number */
+    minor = (endptr + 1)[0] - '0';
+    if((major >= 1 && major <= 6) ||
+       (major == 7 && minor >= 0 && minor <= 7)) {
+        return 1; /* Version is in the specified range */
+    }
```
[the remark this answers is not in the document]
vszakats: Fixed later I believe

src/userauth.c
```diff
@@ -1345,6 +1388,25 @@ _libssh2_key_sign_algorithm(LIBSSH2_SESSION *session,
...
+        if(remote_ver_start) {
+            remote_ver = remote_ver_start + strlen(remote_ver_pre);
+            SSH_BUG_SIGTYPE = is_version_less_than_78(remote_ver);
+            if(SSH_BUG_SIGTYPE && *key_method_len == 28 &&
+               memcmp(key_method, "ssh-rsa-cert-v01@openssh.com",
+                      *key_method_len)) {
+                return LIBSSH2_ERROR_NONE;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
vszakats: Subsequently fixed.

src/userauth.c
```diff
@@ -1412,15 +1474,27 @@ _libssh2_key_sign_algorithm(LIBSSH2_SESSION *session,
...
-
-        *key_method = LIBSSH2_ALLOC(session, match_len);
-        if(*key_method) {
-            memcpy(*key_method, match, match_len);
-            *key_method_len = match_len;
+        if(*key_method_len == 28 &&
+        memcmp(key_method, "ssh-rsa-cert-v01@openssh.com", *key_method_len)) {
+            if(*key_method)
+                LIBSSH2_FREE(session, *key_method);
+            *key_method = LIBSSH2_ALLOC(session, match_len + suffix_len);
```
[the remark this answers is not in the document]
vszakats: Subsequently fixed.

src/openssl.c
```diff
@@ -5134,8 +5134,12 @@ _libssh2_supported_key_sign_algorithms(LIBSSH2_SESSION *session,
...
 
 #if LIBSSH2_RSA_SHA2
-    if(key_method_len == 7 &&
-       memcmp(key_method, "ssh-rsa", key_method_len) == 0) {
+    if((key_method_len == 7 &&
+        memcmp(key_method, "ssh-rsa", key_method_len) == 0) ||
+       (key_method_len == 28 &&
+        memcmp(key_method, "ssh-rsa-cert-v01@openssh.com",
+               key_method_len) == 0)
+       ) {
```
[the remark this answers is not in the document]
vszakats: → PR #2252 (failed attempt)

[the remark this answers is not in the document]
vszakats: → PR #2483

PR #1316, opened 2024-02-08
docs: update INSTALL_AUTOTOOLS

andreiaugustin:
corrected --with-libmbedtls-prefix to current option --with-libmbedcrypto-prefix

Edit:

Updated to latest version at time of writing (1.11.0) and scripting up build process for libssh2 which will use mbedTLS as crypto backend, installed in a non-standard location. 

Docs mentioned `--with-libmbedtls-prefix` which resulted in:

```
configure: WARNING: unrecognized options: --with-libmbedtls-prefix
configure: error: Required dependencies are missing!
```

However, using `--with-libmbedcrypto-prefix` seems to work OK.

### DIFF
--- a/docs/INSTALL_AUTOTOOLS
+++ b/docs/INSTALL_AUTOTOOLS
@@ -286,7 +286,7 @@ Some ./configure options deserve additional comments:
 
  * --with-mbedtls
  * --without-mbedtls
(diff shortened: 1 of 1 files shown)


vszakats (approved):

vszakats:
Thank you!

PR #1327, opened 2024-03-12
kex: always add extension indicators to kex_algorithms

### LINKED ISSUE #1326: KEX extension indicators are lost in libssh2_session_method_pref()
**Describe the bug**
Default list of preferred KEX methods contains also extension indicators "ext-info-c" and "kex-strict-c-v00@openssh.com".
If new list is configured using `libssh2_session_method_pref()` function these indicators are not preserved unless the caller
explicitly includes them in input `*prefs` argument.

This can cause mismatched result of handshake - remote side disables the extensions because libssh2 does not announce them 
while libssh2 enables them according to peer's announcement.

**To Reproduce**
Set valid list of KEX methods before `libssh2_session_handshake()` that does not explicitly include "ext-info-c" and "kex-strict-c-v00@openssh.com" pseudo-methods.

Enable debug logging and connect to a server supporting these extensions.  Check if both sides are using them.

**Expected behavior**
libssh2 based client announce available KEX extensions and both sides have enabled them after handshake.

**Versions:**

SLES12-SP5
 - libssh2-1-1.11.0-29.9.1.x86_64 (or upstream master 4bba038e)
 - openssl-1_0_0-1.0.2p-3.87.1.x86_64

OpenSUSE LEAP 15-5
 - libssh2  master branch 4bba038e
 - openssl-1_1-1.1.1l-150500.17.22.1

**Additional context**
I found this issue when mismatching strict-kex setup after kex init caused failed connection with error:

```
  [libssh2] 0.524957 Failure Event: -43 - Failed to get response to ssh-userauth request
  Failure establishing SSH session: -43 
```

However that depends on negotiated ciphers - with different configuration was the connection established but with different strict-kex setup on both sides.

IMO this bug can disable recent fix for Terrapin Attack (CVE-2023-48795).

malyzelenyhnus:
KEX pseudo-methods "ext-info-c" and "kex-strict-c-v00@openssh.com" must be kept in list in libssh2_session_method_pref() to properly announce available extensions.

Fixes #1326

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -4027,13 +4027,25 @@ libssh2_session_method_pref(LIBSSH2_SESSION * session, int method_type,
                             const char *prefs)
 {
     char **prefvar, *s, *newprefs;
+    char *tmpprefs = NULL;
     size_t prefs_len = strlen(prefs);
     const LIBSSH2_COMMON_METHOD **mlist;
+    const char *kex_extensions = "ext-info-c,kex-strict-c-v00@openssh.com,";
+    size_t kex_extensions_len = strlen(kex_extensions);
 
     switch(method_type) {
     case LIBSSH2_METHOD_KEX:
         prefvar = &session->kex_prefs;
         mlist = (const LIBSSH2_COMMON_METHOD **)libssh2_kex_methods;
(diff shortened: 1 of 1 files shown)


vszakats (approved):

vszakats:
Thank you!

PR #1328, opened 2024-03-12
Need to also copy in the matching public key

bandi13:
Server complains when the private+public keys don't match. The modified Dockerfile will copy in the public key as well. Needed to generate new matching keys to copy in.


bandi13:
Errors related to: https://github.com/libssh2/libssh2/issues/1329

bandi13:
Good idea. I've made the changes you suggested.

vszakats:
I wonder if it'd be best to add the `.pub` files to the repo. In such way
it would also fix non-Docker tests. What do you think?

BTW, do you know at which sshd version did the public keys become required?
Or any pointer why this became an issue?

bandi13:
The public keys were always in the Docker container. When starting the container, you'd see an error from the openssh-server saying that the private keys don't match the public keys. I'm not sure the public keys really need to be in the repo. As you correctly pointed out, they can be generated from the private keys so no need to check it in. It would be just another thing to maintain in the repo.

In the grand scheme of things it doesn't hurt anything either way. The issue I was looking into was the problem with the `libssh2_session_handshake failed` and saw this fix along the way.

vszakats:
Can you point/link to the openssh-server log showing the 'don't match' errors, to understand better? (I looked at https://github.com/libssh2/libssh2/actions/runs/8469053725/job/23203473070, but couldn't spot it.)

Tried a new CI run, and what's failing now is two autotools OpenSSL jobs and one with libgcrypt in 32-bit. (The `distrcheck` job became flaky a good while ago, couldn't figure out why.)
https://github.com/libssh2/libssh2/actions/runs/8470987797/job/23210022514?pr=1337

Why these?
Screen Shot 2024-03-28 libssh2

bandi13:
```
andras:~/git/wolfssl-libssh2/tests/openssh_server (master)$ git checkout fixPubKeyError 
Switched to branch 'fixPubKeyError'
Your branch is up to date with 'origin/fixPubKeyError'.
andras:~/git/wolfssl-libssh2/tests/openssh_server (fixPubKeyError)$ docker build -t libssh2-fixpubkeyerror . 
...
andras:~/git/wolfssl-libssh2/tests/openssh_server (fixPubKeyError)$ docker run --rm -it --name libssh2 libssh2-fixpubkeyerror
Server listening on 0.0.0.0 port 22.
Server listening on :: port 22.
Received signal 15; terminating.
```
Note there are no warnings. Let's try on the current `master`:
```
andras:~/git/wolfssl-libssh2/tests/openssh_server (fixPubKeyError)$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
andras:~/git/wolfssl-libssh2/tests/openssh_server (master)$ docker build -t libssh2-master . 
...
andras:~/git/wolfssl-libssh2/tests/openssh_server (master)$ docker run --rm -it --name libssh2 libssh2-master
Public key for /etc/ssh/ssh_host_rsa_key does not match private key
Public key for /etc/ssh/ssh_host_ecdsa_key does not match private key
Public key for /etc/ssh/ssh_host_ed25519_key does not match private key
Server listening on 0.0.0.0 port 22.
Server listening on :: port 22.
Received signal 15; terminating.
```

vszakats:
Thanks! This is a benign warning; sshd extract the public key from the private one automatically after that. Then the server starts normally. Behaviour was already present at least in 8.4 (2020), ruling it out as a cause for the new fallouts.

The questions remain why these specific 3-4 jobs fail, and why now?

It's also strange that _all_ Dockerized tests failed with this PR applied.

bandi13:
It's nothing to do with my PR. Look at #1329 that I reference. You can open up an empty PR, and all libssh2 Docker tests will fail.

vszakats:
Yeah, it's not this PR. I've linked a brand new (unrelated) PR above, with the screenshot, and that one had just the 4 failures seen earlier.

It seems the Docker stuff has become completely flaky, in particular those 3 ones, preceded by `distcheck` that has been flaky for months now.

No clue what this is.

Maybe switching in `Dockerfile` to `debian:bookrowm-slim` (from `debian:testing-slim`) may help?

vszakats:
This could have been this backdoor in action, active on Linux x86_64 debian:testing, injecting itself into sshd:
https://www.openwall.com/lists/oss-security/2024/03/29/4

PR #1332, opened 2024-03-21
transport: check ETM on remote end when receiving

### LINKED ISSUE #1331: Encrypt-then-MAC feature should be tested in remote end's configuration when receiving data
**Describe the bug**
Both `_libssh2_transport_read() `and `_libssh2_transport_send()` are testing presence of encrypt-then-mac feature by the same code

  `etm = encrypted && session->local.mac ? session->local.mac->etm : 0;`

When reading `session->remote` should be used instead.

**To Reproduce**

Configure only one from LIBSSH2_METHOD_MAC_CS and LIBSSH2_METHOD_MAC_SC to use etm feature:

```
 libssh2_session_method_pref(session, LIBSSH2_METHOD_MAC_CS, "hmac-sha2-256-etm@openssh.com");
 libssh2_session_method_pref(session, LIBSSH2_METHOD_MAC_SC, "hmac-sha2-256");
```

Received data are then misinterpreted and usually it fails with error

>  [libssh2] 0.684036 Failure Event: -41 - Failed to get response to ssh-userauth request

because obtained payload length is random value > `LIBSSH2_PACKET_MAXPAYLOAD`.

 - OS and version SLES12-SP5
 - libssh2 version:  current git master (b2738391)
 - crypto backend and version: openssl-1_0_0-1.0.2p-3.90.1

malyzelenyhnus:
We should check if encrypt-then-MAC feature is enabled in remote end's configuration.

Closes #1331

### DIFF
--- a/src/transport.c
+++ b/src/transport.c
@@ -425,7 +425,7 @@ int _libssh2_transport_read(LIBSSH2_SESSION * session)
                                    make the checks below work fine still */
         }
 
-        etm = encrypted && session->local.mac ? session->local.mac->etm : 0;
+        etm = encrypted && session->remote.mac ? session->remote.mac->etm : 0;
 
         /* read/use a whole big chunk into a temporary area stored in
            the LIBSSH2_SESSION struct. We will decrypt data from that


palmin:
I don't have permission to merge PRs and my work with encode-then-mac lies so far back now that I don't feel confident reviewing the change even though it is small.

Sorry

willco007:
This fix is correct and should be merged.

vszakats (approved):

vszakats:
@palmin: No worries, thanks for your quick feedback!

@willco007: Thanks!

@malyzelenyhnus: Thanks for your patch, merged now.

PR #1347, opened 2024-04-03
docs: improve `libssh2_userauth_publickey_from*` manpages

### LINKED ISSUE #652: Confusing `libssh2_userauth_publickey_fromfile()` documentation

vszakats:
Reported-by: Lyndon Brown
Assisted-by: Ryan Kelley
Fixes #652
Closes #1308
Closes #1347

### DIFF
--- a/docs/libssh2_userauth_publickey_fromfile_ex.3
+++ b/docs/libssh2_userauth_publickey_fromfile_ex.3
@@ -40,7 +40,7 @@ Return 0 on success or negative on failure. It returns
 LIBSSH2_ERROR_EAGAIN when it would otherwise block. While
 LIBSSH2_ERROR_EAGAIN is a negative number, it is not really a failure per se.
 .SH ERRORS
-\fILIBSSH2_ERROR_FILE\fP - An issue opening, reading or parsing disk file(s).
+\fILIBSSH2_ERROR_FILE\fP - An issue opening, reading or parsing the disk file.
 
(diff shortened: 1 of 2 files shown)


vszakats:
Merged via 581b72aab750bd7dcb00acdcbb189ddfee8c93fc.

PR #1348, opened 2024-04-03
ci/appveyor: update OpenSSL path

vszakats:
Ref: https://github.com/appveyor/build-images/commit/702e8cdca01f28f6a40687783f493c786cebbe2c


vszakats:
Could not make CMake recognize the updated OpenSSL 3.1 / 3.2 installation.

With manual config, I managed to make CMake recognize OpenSSL. Then with another manual config, made `CMakeLists.txt` find the DLL, but even then it fails while linking libcrypto.

FWIW https://www.appveyor.com/docs/windows-images-software/ still lists the old OpenSSL 3 path, but that no longer exists on the disk.

vszakats:
Meanwhile mbedTLS 3.6.0 reached Homebrew, and it features breaking changes apparently:
```
../../src/mbedtls.c:1295:8: error: implicit declaration of function 'mbedtls_pk_load_file' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    if(mbedtls_pk_load_file(filename, &data, &data_len))
       ^
../../src/mbedtls.c:1295:8: note: did you mean 'mbedtls_mpi_read_file'?
```
Release Notes has no mention.

PR #1354, opened 2024-04-04
src: check the return value from `_libssh2_bn_*()` functions

MichaelBuckley:
Found by oss-fuzz. In `diffie_hellman_sha_algo()`, we were calling
`_libssh2_bn_from_bin()` with data recieved by the server without
checking whether that data was zero-length or ridiculously long.
In the OpenSSL backend, this would cause `_libssh2_bn_from_bin()`
to fail an allocation, which would eventually lead to a NULL
dereference when the bignum was used.

Add the same check for `_libssh2_bn_set_word()` and
`_libssh2_bn_to_bin()`.

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -522,9 +522,9 @@ static int diffie_hellman_sha_algo(LIBSSH2_SESSION *session,
             goto clean_exit;
         }
 
-        if (_libssh2_bn_from_bin(exchange_state->f,
-                                 exchange_state->f_value_len,
-                                 exchange_state->f_value)) {
+        if(_libssh2_bn_from_bin(exchange_state->f,
+                                exchange_state->f_value_len,
+                                exchange_state->f_value)) {
             ret = _libssh2_error(session, LIBSSH2_ERROR_HOSTKEY_INIT,
                                  "Invalid DH-SHA f value");
             goto clean_exit;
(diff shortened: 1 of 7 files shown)


src/wincng.c
```diff
@@ -2410,8 +2410,11 @@ _libssh2_wincng_bignum_from_bin(_libssh2_bn *bn, ULONG len,
         if(bignum) {
             bn->bignum = bignum;
             bn->length = length;
+            return 0;
         }
     }
+
+    return -1;
```
vszakats, line 2417:
```suggestion
        }
    }

    return 0;
```
This fixes our CI:
https://ci.appveyor.com/project/libssh2org/libssh2/builds/49553846

src/openssl.h
```diff
@@ -402,7 +402,7 @@ libssh2_curve_type;
 #define _libssh2_bn_set_word(bn, val) !BN_set_word(bn, val)
 extern int _libssh2_bn_from_bin(_libssh2_bn *bn, size_t len,
                                 const unsigned char *v);
-#define _libssh2_bn_to_bin(bn, val) BN_bn2bin(bn, val)
+#define _libssh2_bn_to_bin(bn, val) BN_bn2bin(bn, val) <= 0
```
vszakats, line 405:
```suggestion
#define _libssh2_bn_to_bin(bn, val) (BN_bn2bin(bn, val) <= 0)
```

src/kex.c
```diff
@@ -563,7 +563,12 @@ static int diffie_hellman_sha_algo(LIBSSH2_SESSION *session,
         _libssh2_htonu32(exchange_state->k_value,
                          (uint32_t)(exchange_state->k_value_len - 4));
         if(_libssh2_bn_bits(exchange_state->k) % 8) {
-            _libssh2_bn_to_bin(exchange_state->k, exchange_state->k_value + 4);
+            if (_libssh2_bn_to_bin(exchange_state->k,
+                                   exchange_state->k_value + 4)) {
```
vszakats:
```suggestion
            if(_libssh2_bn_to_bin(exchange_state->k,
                                  exchange_state->k_value + 4)) {
```
Minor style nit.

MichaelBuckley:
I have no further changes for this PR. Thank you so much, @vszakats for pointing out a few issues.

vszakats:
No worries, thank you for the PR!

PR #1361, opened 2024-04-09
negative repro test

vszakats:


vszakats:
Successful:
```
Binary files pass1/libssh2-99.98.97.zip and pass2/libssh2-99.98.97.zip differ
```
https://github.com/libssh2/libssh2/actions/runs/8613743002/job/23605701574?pr=1361#step:13:311

PR #1364, opened 2024-04-13
Some fixes for the os/400

monnerat:
Add two recent files to the distribution.

Build scripts must be executed by the os/400 shell (sh), not bash which is a PASE program: The `-ot` non-posix test extension works in os/400 as well.

This is a follow-up to 8457c37 and c662570.

Thanks for considering it.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -26,6 +26,7 @@ WIN32FILES = src/libssh2.rc
 
 OS400FILES = os400/README400 os400/initscript.sh os400/make.sh \
   os400/make-src.sh os400/make-rpg.sh os400/make-include.sh \
+  os400/config400.default \
   os400/os400sys.c os400/ccsid.c \
   os400/libssh2_config.h os400/macros.h os400/libssh2_ccsid.h \
   os400/include/alloca.h os400/include/sys/socket.h os400/include/stdio.h \
@@ -33,7 +34,8 @@ OS400FILES = os400/README400 os400/initscript.sh os400/make.sh \
   os400/libssh2rpg/libssh2.rpgle.in \
   os400/libssh2rpg/libssh2_ccsid.rpgle.in \
   os400/libssh2rpg/libssh2_publickey.rpgle \
-  os400/libssh2rpg/libssh2_sftp.rpgle
+  os400/libssh2rpg/libssh2_sftp.rpgle \
+  os400/rpg-examples/SFTPXMPLE
 
 EXTRA_DIST = $(WIN32FILES) get_ver.awk \
   maketgz RELEASE-NOTES libssh2.pc.in $(VMSFILES) config.rpath \
--- a/os400/config400.default
+++ b/os400/config400.default
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 # Copyright (C) The libssh2 project and its contributors.
 # SPDX-License-Identifier: BSD-3-Clause
 
--- a/os400/initscript.sh
+++ b/os400/initscript.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 # Copyright (C) The libssh2 project and its contributors.
 # SPDX-License-Identifier: BSD-3-Clause
 
--- a/os400/make-include.sh
+++ b/os400/make-include.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 # Copyright (C) The libssh2 project and its contributors.
 # SPDX-License-Identifier: BSD-3-Clause
 #
--- a/os400/make-rpg.sh
+++ b/os400/make-rpg.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 # Copyright (C) The libssh2 project and its contributors.
 # SPDX-License-Identifier: BSD-3-Clause
 #
--- a/os400/make-src.sh
+++ b/os400/make-src.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 # Copyright (C) The libssh2 project and its contributors.
 # SPDX-License-Identifier: BSD-3-Clause
 #
--- a/os400/make.sh
+++ b/os400/make.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 # Copyright (C) The libssh2 project and its contributors.
 # SPDX-License-Identifier: BSD-3-Clause
 #


vszakats:
Thanks for jumping in @monnerat.

The new shebang excludes these files from shellcheck completely, which isn't what we'd want. The warning would re-appear if we make the test recognize this shebang variation.

To avoid these I suggest restoring the original shebang `#!/bin/sh` and silence the `-ot` warning with a `# shellcheck disable=SC3013` comment right above the use.

What do you think?

vszakats:
@monnerat: Is there a particular reason you choose `#!/usr/bin/env sh`, over the previous `#!/bin/sh`?

monnerat:
> Is there a particular reason you choose #!/usr/bin/env sh, over the previous #!/bin/sh?

No. In fact, I don't like /usr/bin/env at all and just wanted to keep your changes as much as possible.

> To avoid these I suggest restoring the original shebang #!/bin/sh and silence the -ot warning with a # shellcheck disable=SC3013 comment right above the use.

`#!/bin/sh` is pretty OK for me. But I think shebangs should be the first line so this is incompatible with preceding it with a shellcheck comment.

IMO those kind of checker programs like shellcheck, lint, etc. should be a source of hints rather than a master commander. The developer finally decides what is good or not, not the checker.

BTW: you completely wrecked the `get_make_vars()` function in the curl project. As it's really picky about escaping, I'll have hours to spend on rewriting it :-(
As they were perfectly working, I really would have preferred os400 files be treated as a _dialect_ and left out from checkers.

vszakats:
> No. In fact, I don't like /usr/bin/env at all and just wanted to keep your changes as much as possible.
>
> > To avoid these I suggest restoring the original shebang #!/bin/sh and silence the -ot warning with a # shellcheck disable=SC3013 comment right above the use.
> 
> `#!/bin/sh` is pretty OK for me. But I think shebangs should be the first line so this is incompatible with preceding it with a shellcheck comment.

Got it, thanks!

I meant right before the line causing the warning (not before the shebang).

> IMO those kind of checker programs like shellcheck, lint, etc. should be a source of hints rather than a master commander. The developer finally decides what is good or not, not the checker.

shellcheck is usually quite sane, and false positive is easy to silence with a comment.

> BTW: you completely wrecked the `get_make_vars()` function in the curl project. As it's really picky about escaping, I'll have hours to spend on rewriting it :-( As they were perfectly working, I really would have preferred os400 files be treated as a _dialect_ and left out from checkers.

I'm sorry about `get_make_vars()`. It was one of the cases that
I wasn't sure of.

The upside of handling them all equally, is to improve to chance
of modifying them without breaking them by accident. That's the
intent anyway.

There wasn't many changes necessary to make the OS/400 scripts
pass shellcheck.

vszakats:
Attempt to fixup `get_make_vars()`: https://github.com/curl/curl/pull/13362/commits/c53443209b22c914eb3289871f4f60b0a4516aab via https://github.com/curl/curl/pull/13362

monnerat:
> Attempt to fixup get_make_vars(): https://github.com/curl/curl/commit/c53443209b22c914eb3289871f4f60b0a4516aab via https://github.com/curl/curl/pull/13362

Thanks but please don't: I'm currently working on it and I'll submit at PR when it will be ready.

> shellcheck is usually quite sane, and false positive is easy to silence with a comment.

I'm quite sure of that. What disturbs me is another checker may, for example, prefer backticks over `$()` (which did not exist in early shell(s) BTW) and suggest the opposite change. Both are right, as it's a matter of convention, like a portughese text reviewer and a brazilian one will not tag language constructs the same way!

in other words, if no new feature and no bug fix, why changing what is working ?

vszakats:
shellcheck steers to objectively better shell syntax. It's a good tool. And not a style checker. Speaking of backticks, those are a relic, and as the above regression shows, and the struggle with double-triple escape-hell that it requires, I think we are better off to not use it. It's compatibility advantage also seems merely theoretical at this point.

I'm sorry to have broken that function, and I have made my best to fix it. It works here. There's not much more than I can do here.

monnerat:
> shellcheck steers to objectively better shell syntax. It's a good tool. And not a style checker.

From their point of you (and obvioulsy yours): yes. But as I wrote before, there may be other opinions. Personally, I don't really have one: the best tool is the one you're used to. It's like the war between emacs and vi users: completely sterile!

And backticks were just an example. I used them because I pratice shell since 1981 and I don't even know the os/400 shell supported `$()`. But really, I don't care. It must work, that's all.

> I'm sorry to have broken that function, and I have made my best to fix it.

Never mind, that's all right. But it's probably better I fix it myself and you check it afterwards with shellcheck because I can first test the behavior on pub400.com: I'll move towards the `$()` direction and you'll comment for shellcheck pragma addition on the PR, I think this is the best way to handle it.

For the moment, I work on curl, not libssh2 because I can't have both at the same time on the limited pub400.com storage assigned to me :-(

monnerat:
You closed this current PR but forgot to include the Makefile.am changes in your commit. Would you like another PR just for them ?

vszakats:
> > shellcheck steers to objectively better shell syntax. It's a good tool. And not a style checker.
> 
> From their point of you (and obvioulsy yours): yes. But as I wrote before, there may be other opinions. Personally, I don't really have one: the best tool is the one you're used to. It's like the war between emacs and vi users: completely sterile!

It's a standard compliance and syntax checker. These are not a matter of opinion,
like vim vs emacs or tabs vs spaces. I'm happy to learn about specific alternative 
tools, but reading this space for the last 5-10 years, I haven't ssen one that 
compares. But hey, if someone does, it'll surely be an improvement to replace 
this one with it.

Speaking of OS/400 scripts, they were all pretty clean, except some minor 
stuff, mostly the backticks. I've made the detailed commit message to make my
changes easier to review, not because the issues were grave.

> Never mind, that's all right. But it's probably better I fix it myself and you check it afterwards with shellcheck because I can first test the behavior on pub400.com: I'll move towards the `$()` direction and you'll comment for shellcheck pragma addition on the PR, I think this is the best way to handle it.

Sure, I'm good with that!

> For the moment, I work on curl, not libssh2 because I can't have both at the same time on the limited pub400.com storage assigned to me :-(

Oh okay, no worries. I have committed the fix for this particular PR,
which was basically a revert of a sub-commit from my initial PR, and
the FIXME comment deleted. I also reverted the two line-splittings that
made my editor's syntax highlighter happy. This was not of public interest.

In curl I have prepared these too, plus my `eval sed` fix attempt. I'm
happy to drop it or update in favour of your patch. Whichever works 
best for you.

Thanks for looking into these!

vszakats:
I apologize, totally forgot about `Makefile.am`. Reopened and merged that update now: e4c65e5b1b96ced372aeb8244ad97ee7aece7d47

monnerat:
> Reopened and merged that update now

Many thanks. Will retest as soon as I finish with curl :-)
Good night.

monnerat:
I just retested it on OS/400 and everything seems OK now. Thanks again.

PR #1377, opened 2024-04-25
ci/GHA: fix gcrypt with autotools/macOS/Homebrew/ARM64

vszakats:
mbedtls configure fails to detect anything due to this:
```
configure:23101: gcc -o conftest -g -O2 -I/opt/homebrew/include  conftest.c  -lmbedcrypto -lz >&5
ld: library 'mbedcrypto' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -542,7 +542,7 @@ jobs:
             cmake: -DCRYPTO_BACKEND=wolfSSL
           - name: 'libgcrypt'
             install: libgcrypt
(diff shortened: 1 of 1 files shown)


vszakats:
It seems autotools fails to detect libgcrypt and mbedtls on Homebrew macOS/ARM64. May have to do with these libs being not at `/usr/local` but at `/opt/homebrew`. No libdir is passed with mbedtls.

PR #1380, opened 2024-05-03
OpenSSL 3: Fix calculating DSA public key

MichaelBuckley:
Caught this while merging into Panic's libssh2 repo, but I probably missed it in code review earlier. My goal is to upstream all the changes from our repo, so more pull requests coming soon.

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -1671,8 +1671,8 @@ gen_publickey_from_dsa(LIBSSH2_SESSION* session, libssh2_dsa_ctx *dsa,
     BIGNUM * pub_key = NULL;
 
     EVP_PKEY_get_bn_param(dsa, OSSL_PKEY_PARAM_FFC_P, &p_bn);
-    EVP_PKEY_get_bn_param(dsa, OSSL_PKEY_PARAM_FFC_G, &q);
-    EVP_PKEY_get_bn_param(dsa, OSSL_PKEY_PARAM_FFC_Q, &g);
+    EVP_PKEY_get_bn_param(dsa, OSSL_PKEY_PARAM_FFC_Q, &q);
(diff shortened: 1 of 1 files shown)


MichaelBuckley:
There's one failing CI check, but it was because the CI couldn't find zlib, which has nothing to do with my change.

vszakats:
> There's one failing CI check, but it was because the CI couldn't find zlib, which has nothing to do with my change.

I can confirm. The mbedTLS detection (or something around it) no longer works since the runners got bumped to ARM.

If no one plans to work on this, we should probably disable this job. I didn't do it right to away to keep the issue visible (but an actual open Issue might serve this better).

Ref: ae2770de25949bc7c74e60b4cc6a011bbe1d3d7c #1377

vszakats (approved):

MichaelBuckley:
I'll see what I can do for this runner. It looks like it's just unable to find zlib.

MichaelBuckley:
Fixed it, broke three others, but I can break them into separate rules to make it work.

The core problem is that the configure script can't seem to handle it if we specify both --with-libmbedcrypto-prefix and --with-libz-prefix, so I had to set the LDFLAGS enviornment variable.

I'll make the fix in a new branch with a new pull request, once that passes, merge it, and then rebase this branch on top of that.

MichaelBuckley:
I'm just going to merge this one. It looks like the AppVeyor runners are tired from how many builds we've put them through today, but we know they passed on an earlier commit.

vszakats:
No worries, this already finished green earlier.

AppVeyor is prone to fail when GHA jobs are queued up (or something). One GHA workflow is providing the server for the AppVeyor tests, and in such cases the server either doesn't come up online or sometimes dies while tests are still executing. It would be nice to stabilize this.

PR #1381, opened 2024-05-08
ci: fix mbedtls runners on macOS

MichaelBuckley:


vszakats:
Isn't this ultmately a bug in our configure scripts?

MichaelBuckley:
It might be, but I also don't think this would affect normal macOS installations. I could be wrong here, but the issue appears to be that libz is only installed on /opt/homebrew on the runners, while most macOS installations also have it in /usr.

But I definitely could be mistaken.

vszakats:
Ah okay, if this is specific to the GHA runners, and not to regular macOS ARM + Homebrew installs, I agree we shouldn't chase it further. (Aside from possibly reporting it to GitHub.)

MichaelBuckley:
I can try it on a clean macOS ARM install this evening and see if it works.

PR #1384, opened 2024-05-08
autotools: fix to update `LDFLAGS` for each detected dependency

vszakats:
autotools lib detection routine failed to extend LDFLAGS for each
detection. This could cause successful detection of a dependency, but
later failing to use it. This did not cause an issue as long as all
dependencies lived under the same prefix, but started breaking on macOS
ARM + Homebrew where this was no longer true for mbedTLS and zlib in
particular.

Follow-up to 844115393bffb4e92c6569204cbe4cd8e553480d #1381
Follow-up to ae2770de25949bc7c74e60b4cc6a011bbe1d3d7c #1377
Closes #1384

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -558,8 +558,6 @@ jobs:
       - name: 'autotools configure'
         if: ${{ matrix.build == 'autotools' }}
         run: |
-          # Workaround for GHA ARM runner issue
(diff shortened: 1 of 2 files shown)


MichaelBuckley:
That works, and is much cleaner. Nice.

MichaelBuckley (approved):

PR #1385, opened 2024-05-09
Add flags parameter to LIBSSH2_AUTHAGENT_SIGN_FUNC

MichaelBuckley:
Another change from the Panic repo. Allows agent signing using rsa-sha2-256 and rsa-sha-512 when setting a callback with `LIBSSH2_CALLBACK_AUTHAGENT_SIGN`.

### DIFF
--- a/docs/libssh2_session_callback_set2.3
+++ b/docs/libssh2_session_callback_set2.3
@@ -121,6 +121,7 @@ The prototype of the callback:
 
 .nf
 void sign(LIBSSH2_SESSION* session,
+          unsigned int flags,
           unsigned char *blob, unsigned int blen,
           const unsigned char *data, unsigned int dlen,
           unsigned char **sig, unsigned int *sig_len,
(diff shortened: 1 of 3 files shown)


vszakats:
Thanks!

Wouldn't this be an ABI breaking change? In such case perhaps we might do a `libssh2_session_callback_set3`, while also fixing up other stuff like going with `size_t`, if it makes sense here?

MichaelBuckley:
It's definitely an ABI breaking change. In this case, I just created a pull request from a commit in our repo that someone else at Panic wrote, without considering if it broke ABI.

If we wanted to preserve ABI, rather than making a `libssh2_session_callback_set3`, I think we'd have to create a  `LIBSSH2_CALLBACK_AUTHAGENT_SIGN_2` and a `LIBSSH2_AUTHAGENT_SIGN_2`. The callback registration wouldn't need to change, since how users interact with this function is by calling `LIBSSH2_AUTHAGENT_SIGN`.

vszakats:
> If we wanted to preserve ABI, rather than making a `libssh2_session_callback_set3`, I think we'd have to create a `LIBSSH2_CALLBACK_AUTHAGENT_SIGN_2` and a `LIBSSH2_AUTHAGENT_SIGN_2`. The callback registration wouldn't need to change, since how users interact with this function is by calling `LIBSSH2_AUTHAGENT_SIGN`.

You're right, that's a much better way to go.

vszakats:
@MichaelBuckley Are you working on a `LIBSSH2_CALLBACK_AUTHAGENT_SIGN_2` / `LIBSSH2_AUTHAGENT_SIGN_2` solution? Should we keep this PR open?

PR #1387, opened 2024-05-10
lib: detect and use `bzero()` if available

vszakats:
Ref: https://github.com/curl/curl/pull/13589
Closes #1387


vszakats:
According to https://www.man7.org/linux/man-pages/man3/bzero.3.html, `bzero()` doesn't give guarantees, making it like `memset()`.

willco007:
This commit checks for bzero but then uses explicit_bzero(), was that intentional?

vszakats:
This was a mistake, `bzero()` is not a secure way to zero memory. Closed without merging when I realized.

willco007:
Oh sure, I missed you closed without merging. 👍

PR #1389, opened 2024-05-12
ci/GHA: cygwin experiments

vszakats:


vszakats:
No improvements. Cygwin job doesn't seem to have `set -e` and setting it makes most commands (e.g. `autoreconf`) stop the build. Closing without merged.

PR #1399, opened 2024-05-30
ci/GHA: try ubuntu 24.04 runners

vszakats:
Use this workaround to fix an image issue:
https://github.com/actions/runner-images/pull/9956

Closes #1399


vszakats:
Blocked by: https://github.com/actions/runner-images/pull/9956

vszakats:
Letting GitHub fixing up the image to not perform and auto-shutdown after `apt-get install`, then rolling with this when it becomes the default.

PR #1406, opened 2024-06-11
src: fix type warning in `libssh2_sftp_unlink` macro

tarinyoom:
The `libssh2_sftp_unlink` macro was implicitly casting the `size_t` returned by `strlen` to the `unsigned int` type expected by `libssh2_sftp_unlink_ex`. This fix adds an explicit cast to match similar macro definitions in the same file (e.g. `libssh2_sftp_rename`, `libssh2_sftp_mkdir`).


vszakats:
Nice catch, thanks!

PR #1394, opened 2024-05-24
Call EVP_CipherFinal when using WolfSSL.

MichaelBuckley:
This should fix #1020 and #1299.

PR #797, opened 2023-02-15
Add support for AES-GCM crypto protocols

dfandrich:
Add support for aes256-gcm@openssh.com and aes128-gcm@openssh.com
ciphers, which are the OpenSSH implementations of AES-GCM cryptography.
It is similar to RFC5647 but has changes to the MAC protocol
negotiation.  These are implemented for recent versions of OpenSSL only.

The ciphers work differently than most previous ones in two big areas:
the cipher includes its own integrated MAC, and the packet length field
in the SSH frame is left unencrypted.  The code changes necessary are
gated by flags in the LIBSSH2_CRYPT_METHOD configuration structure.

These differences mean that both the first and last parts of a block
require special handling during encryption. The first part is where the
packet length field is, which must be kept out of the encryption path
but in the authenticated part (as AAD).  The last part is where the
Authentication Tag is found, which is calculated and appended during
encryption or removed and validated on decryption. As encryption/
decryption is performed on each packet in a loop, one block at a time,
flags indicating when the first and last blocks are being processed are
passed down to the encryption layers.

The strict block-by-block encryption that occurs with other protocols is
inappropriate for AES-GCM, since the packet length shifts the first
encrypted byte 4 bytes into the block. Additionally, the final part of
the block must contain the AES-GCM's Authentication Tag, so it must be
presented to the lower encryption layer whole. These requirements mean
added code to consolidate blocks as they are passed down.

When AES-GCM is negotiated as the cipher, its built-in MAC is
automatically used as the SSH MAC so further MAC negotiation is not
necessary.  The SSH negotiation is skipped when _libssh2_mac_override()
indicates that such a cipher is in use.  The virtual MAC configuration
block mac_method_hmac_aesgcm is then used as the MAC placeholder.

This work was sponsored by Anders Borum.

PR #1407, opened 2024-06-21
wolfssl: fix `EVP_Cipher()` use with v5.6.0 and older

### LINKED ISSUE #1020: wolfSSL: libssh2_session_handshake failed: -12/-44 DECRYPT/ENCRYPT

### LINKED ISSUE #1299: Encryption failure during handshake when using libssh2 and wolfssl
I'm getting a -44 error code and message of "Unable to ask for ssh-userauth service" when using wolfssl as the encryption engine during ssh handshake.  I've tracked the error to this code on line 584 of openssl.c:

```
if(IS_LAST(firstlast)) {
  /* This is the last block.
     encrypt: compute tag, if applicable
     decrypt: verify tag, if applicable
     in!=NULL is equivalent to EVP_CipherUpdate
     in==NULL is equivalent to EVP_CipherFinal */
#ifdef HAVE_OPAQUE_STRUCTS
  ret = EVP_Cipher(*ctx, NULL, NULL, 0); /* final */
#else
  ret = EVP_Cipher(ctx, NULL, NULL, 0); /* final */
#endif
```

The comment indicates the expectation that EVP_CipherFinal should be equivalent to EVP_Cipher when finalizing the encryption.  The EVP_Cipher method resolves to wolfSSL_EVP_Cipher on line 7080 of the wolfssl/wolfcrypt/src/evp.c file.  There a check of NULL on the incoming buffers returns a WOLFSSL_FATAL_ERROR.  After modifying the code to follow the comment, EVP_CipherFinal is used instead of EVP_Cipher, then the encryption seems to work as expected:

```
int final_length = 0;
ret = EVP_CipherFinal(*ctx, buf + aadlen, &final_length); /* final */
```

I do not know enough about these internals to know if this is the correct solution or not and would appreciate any assistance in getting this fixed.  The comment suggests a break in the interface expectations with wolfssl.

**To Reproduce**
Extract libssh2 1.11.0, wolfssl 5.6.4.  Build wolfssl.  Build libssh2 to use wolfssl.  Attempt to connect to an ssh server.  In my case OpenSSH 8.9.

**Expected behavior**
For the encryption to work and the ssh handshake to complete.

**Version**
 - OS and version: Linux Ubuntu 22.04.3 LTS
 - libssh2 version: 1.11.0
 - crypto backend and version: wolfSSL 5.6.4

**Additional context**

vszakats:
Add workaround for the wolfSSL `EVP_Cipher(*p, NULL, NULL, 0)` bug to
make libssh2 work with wolfSSL v5.6.0 and older.

wolfSSL fixed this issue in v5.7.0:
https://github.com/wolfSSL/wolfssl/pull/7143
https://github.com/wolfSSL/wolfssl/commit/b0de0a1c95119786cf5651dd76dd7d7bdfac5a04

Without our local workaround:

- v5.3.0 and older fail most tests:
  Ref: https://github.com/libssh2/libssh2/actions/runs/9646827522/job/26604211476#step:17:1263

- v5.4.0, v5.5.x, v5.6.0 fail these:
  ```
  29 - test_read-aes128-cbc (Failed)
  30 - test_read-aes128-ctr (Failed)
  32 - test_read-aes192-cbc (Failed)
  33 - test_read-aes192-ctr (Failed)
  34 - test_read-aes256-cbc (Failed)
  35 - test_read-aes256-ctr (Failed)
  ```
  Ref: https://github.com/libssh2/libssh2/actions/runs/9646827522/job/26604233819#step:17:978

Oddly enough the workaround breaks OpenSSL tests, so only enable it for
the affected wolfSSL versions.

Also add new build-from-source wolfSSL CI job to test the new codepath.

wolfSSL has a build bug where `wolfssl/options.h` and
`wolfssl/version.h` are not copied to the `install` destination with
autotools. With CMake it has a different bug where `wolfcrypt/sp_int.h`
is not copied (with v5.4.0). And another with CMake where `FIPS_mode()`
remains missing (with v5.6.0 and earlier.)

Therefore use CMake with v5.5.4 and a workaround for `FIPS_mode()`.
Another option is autotools with v5.4.0 and a workaround for `install`,
but CMake builds quicker.

Regression-from 3c953c05d67eb1ebcfd3316f279f12c4b1d600b4 #797
Fixes #1020
Fixes #1299
Assisted-by: Michael Buckley via #1394
Closes #1394 (another attempt to fix the mentioned wolfSSL bug)
Closes #1407

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,11 @@ jobs:
             crypto: wolfSSL-from-source
             build: cmake
             zlib: 'ON'
+          - compiler: clang
+            arch: amd64
+            crypto: wolfSSL-from-source-prev
+            build: cmake
+            zlib: 'ON'
           - compiler: gcc
             arch: amd64
             crypto: OpenSSL-3-no-deprecated
@@ -162,18 +167,25 @@ jobs:
           echo "TOOLCHAIN_OPTION=$TOOLCHAIN_OPTION -DCMAKE_PREFIX_PATH=$PWD/usr" >> $GITHUB_ENV
 
       - name: 'install wolfSSL from source'
-        if: ${{ matrix.crypto == 'wolfSSL-from-source' }}
+        if: ${{ startsWith(matrix.crypto, 'wolfSSL-from-source') }}
         run: |
-          WOLFSSLVER=5.7.0
+          if [ '${{ matrix.crypto }}' = 'wolfSSL-from-source' ]; then
+            WOLFSSLVER=5.7.0
+          else
+            WOLFSSLVER=5.5.4
+            # Required to include `FIPS_mode()` API:
+            options='-DWOLFSSL_OPENSSLEXTRA=ON'
+            cppflags='-DOPENSSL_ALL'
+          fi
           curl -fsS -L https://github.com/wolfSSL/wolfssl/archive/refs/tags/v$WOLFSSLVER-stable.tar.gz | tar -xzf -
           cd wolfssl-$WOLFSSLVER-stable
-          cmake -B bld \
+          cmake -B bld ${options} \
             -DWOLFSSL_LIBSSH2=ON \
             -DBUILD_SELFTEST=OFF \
             -DWOLFSSL_OPENSSLALL=ON \
             -DWOLFSSL_EXAMPLES=OFF \
             -DWOLFSSL_CRYPT_TESTS=OFF \
-            '-DCMAKE_C_FLAGS=-fPIC -DWOLFSSL_AESGCM_STREAM' \
+            "-DCMAKE_C_FLAGS=-fPIC -DWOLFSSL_AESGCM_STREAM ${cppflags}" \
             "-DCMAKE_INSTALL_PREFIX=$PWD/../usr"
           make -j5 -C bld install
           cd ..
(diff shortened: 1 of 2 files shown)


vszakats:
What remains a puzzle for another time is why `EVP_CipherFinal()`
doesn't work with OpenSSL (and other forks). OpenSSL actually
recommends it (and `EVP_CipherUpdate()`) over `EVP_Cipher()`.

PR #1415, opened 2024-06-26
tests: avoid using `MAXPATHLEN`, for portability

### LINKED ISSUE #1414: MAXPATHLEN not defined on hurd systems
**Describe the bug**
When building libssh2 on hurd* systems, the tests fail to build because `MAXPATHLEN` is not defined on those systems:

```
session_fixture.c:231:36: error: ‘MAXPATHLEN’ undeclared (first use in
this function)
  231 |     static char filepath[NUMPATHS][MAXPATHLEN];
      |                                    ^~~~~~~~~~
```

See [Debian build logs on hurd-i386](https://buildd.debian.org/status/fetch.php?pkg=libssh2&arch=hurd-i386&ver=1.11.0-5&stamp=1716421017&raw=0)

 - OS and version: Debian hurd i386 and Debian hurd amd64
 - libssh2 version: 1.11.0
 - crypto backend and version: OpenSSL 3.2.2

**Additional context**
The build issue is in the test file [tests/session_fixture.c](https://github.com/libssh2/libssh2/blob/master/tests/session_fixture.c#L257)

The bug open in debian is [#1068076](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1068076)

babelouest:
`MAXPATHLEN` is not present in some systems, e.g. GNU Hurd.

Fixes #1414
Closes #1415

### DIFF
--- a/tests/session_fixture.c
+++ b/tests/session_fixture.c
@@ -250,7 +250,7 @@ void stop_session_fixture(void)
 #define NUMPATHS 32
 char *srcdir_path(const char *file)
 {
-    static char* filepath[NUMPATHS];
+    static char *filepath[NUMPATHS];
     static int curpath;
     char *p = getenv("srcdir");
     if(curpath >= NUMPATHS) {
@@ -497,7 +497,8 @@ int test_auth_pubkey(LIBSSH2_SESSION *session, int flags,
         free(buffer);
     }
     else {
-        char *path_fn_pub = srcdir_path(fn_pub), *path_fn_priv = srcdir_path(fn_priv);
+        char *path_fn_pub = srcdir_path(fn_pub),
+             *path_fn_priv = srcdir_path(fn_priv);
         rc = libssh2_userauth_publickey_fromfile_ex(session, username,
                                                 (unsigned int)strlen(username),
                                                     path_fn_pub,
--- a/tests/session_fixture.c
+++ b/tests/session_fixture.c
@@ -48,7 +48,7 @@
 #endif
 
 #include <stdio.h>
-#include <stdlib.h>
+#include <stdlib.h>  /* for getenv() */
 #include <assert.h>
 
 static LIBSSH2_SESSION *connected_session = NULL;
--- a/tests/session_fixture.c
+++ b/tests/session_fixture.c
@@ -255,7 +255,7 @@ char *srcdir_path(const char *file)
     static char *filepath[NUMPATHS];
     static int curpath;
     char *p = getenv("srcdir");
-    if(!file) {
+    if(file) {
         if(curpath >= NUMPATHS) {
             fprintf(stderr, "srcdir_path ran out of filepath slots.\n");
         }
--- a/tests/session_fixture.c
+++ b/tests/session_fixture.c
@@ -254,26 +254,36 @@ char *srcdir_path(const char *file)
     static int curpath;
     char *p = getenv("srcdir");
     if(file) {
+        int len;
         if(curpath >= NUMPATHS) {
             fprintf(stderr, "srcdir_path ran out of filepath slots.\n");
         }
         assert(curpath < NUMPATHS);
         if(p) {
-            int len = snprintf(NULL, 0, "%s/%s", p, file);
-            filepath[curpath] = malloc(len + 1);
-            snprintf(filepath[curpath], len + 1, "%s/%s", p, file);
+            len = snprintf(NULL, 0, "%s/%s", p, file);
+            if(len >= 0) {
+                filepath[curpath] = calloc(1, (size_t)len + 1);
+                snprintf(filepath[curpath], len + 1, "%s/%s", p, file);
+            }
+            else {
+               return NULL;
+            }
         }
         else {
-            int len = snprintf(NULL, 0, "%s", file);
-            filepath[curpath] = malloc(len + 1);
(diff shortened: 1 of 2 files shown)


vszakats:
valgrind tests would be a nice addition, but I personally have not
much experience with it. (it never got supported in my dev env, macOS)

As for freeing pointers:
I think the idea was to keep a fixed list of pointers so callers
don't have to deal with freeing after use.

What do you think of an approach of freeing `filepath` array
on deinit in `stop_session_fixture()`?

If we move to freeing right after use, there are two more cases
to update in `test_agent_forward_ok.c` and `test_read.c`. And
in this case the static `filepath` list along with `curpath` counter 
becomes unnecessary.

babelouest:
Thanks for your feedback!

I've updated the PR to cleanup `filepath` during `stop_session_fixture()` instead.

Concerning valgrind, I use it by executing the test program via valgrind and checking the result, example:

```
$ valgrind --tool=memcheck --leak-check=full --show-leak-kinds=all ./test_auth_password_ok 2>valgrind-test_auth_password_ok.log
```

I don't know exactly where the test programs are executed in libssh2, I guess [tests/test_sshd.test](https://github.com/libssh2/libssh2/blob/master/tests/test_sshd.test) is one of the starting point.

Although, the docker tests should be updated too, but I don't know how yet.

vszakats:
Thanks for you updates!

checksrc has these complaints:
```
./tests/session_fixture.c:258:8: warning: we prefer !variable instead of "== NULL" comparisons (EQUALSNULL)
     if(NULL != file) {
        ^
./tests/session_fixture.c:259:7: warning: not indented 4 steps (uses 2) (INDENTATION)
       if(curpath >= NUMPATHS) {
       ^
./tests/session_fixture.c:280:9: warning: not indented 4 steps (uses 2) (INDENTATION)
         free(filepath[curpath]);
         ^
```

Tests are launched via the `TESTS` list in `tests/Makefile.am` for autotools and for CMake via `tests/CMakeLists.txt` /  `add_test()`. Some launched via `.test` scripts, others directly. Plugging valgrind into `.test` scripts could be a sensible starting point. Probably with a low `FIXTURE_XFER_COUNT` to maintain CI turnaround times.

vszakats:
Looking good.

Based on https://learn.microsoft.com/en-us/cpp/c-runtime-library/path-field-limits https://www.man7.org/linux/man-pages/man3/realpath.3.html and https://github.com/libssh2/libssh2/pull/198/commits/10a5cbf945abcc60153ee3d59284d09fc64ea152 we might be able to drop `<stdlib.h>` and `<sys/param.h>`. It seems these were included for `MAXPATHLEN`.

If CI remains green, we could drop `sys/param.h` detection:
```
CMakeLists.txt:    set(HAVE_SYS_PARAM_H 1)
CMakeLists.txt:  check_include_files("sys/param.h" HAVE_SYS_PARAM_H)  # tests
configure.ac:AC_CHECK_HEADERS([errno.h fcntl.h stdio.h unistd.h sys/param.h sys/uio.h])
src/libssh2_config_cmake.h.in:#cmakedefine HAVE_SYS_PARAM_H
src/libssh2_setup.h:# define HAVE_SYS_PARAM_H
tests/session_fixture.c:#ifdef HAVE_SYS_PARAM_H
tests/session_fixture.c:#include <sys/param.h>
```

Would you like to do these? Or let me know if I should do it.

babelouest:
Removing `<stdlib.h>` and `<sys/param.h>` doesn't seem to work. The errors seem to be about missing declarations from `stdlib`, I'll give another try without `<sys/param.h>` only.

If successful, I'd like you to remove `sys/param.h` detection from the build scripts, because I'm not familiar with those and I would be afraid to silently break something else, if you don't mind.

vszakats:
Agreed! `stdlib.h` still needed for `getenv`. I'll add such comment later.

babelouest:
It seems that removing `<sys/param.h>` does break `AppVeyor` CI tests as well, do you know if this could be fixed without re-including `<sys/param.h>`?

vszakats:
I think the AppVeyor job failed for unrelated reasons. It sometimes does that while trying to interact with a GHA server process and failing ("Waiting for SSH connection from GitHub Actions..[...] failed.", then timing out.

If you push (or force-push) the branch it will make it re-run.

babelouest:
> If you push (or force-push) the branch it will make it re-run.

After a small [change/push](https://github.com/libssh2/libssh2/pull/1415/commits/6cc363c4347c97c94f71ace3305ec5672a842dd0), still no luck with AppVeyor tests.

tests/session_fixture.c
```diff
@@ -46,9 +46,6 @@
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
 #endif
-#ifdef HAVE_SYS_PARAM_H
-#include <sys/param.h>
-#endif
 
 #include <stdio.h>
 #include <stdlib.h>
```
vszakats, line 51:
```suggestion
#include <stdlib.h>  /* for getenv() */
```

vszakats:
It looks like the server GHA workflow no longer works: https://github.com/libssh2/libssh2/actions/workflows/appveyor_docker.yml
https://github.com/libssh2/libssh2/actions/runs/9700643556/job/26772513786

vszakats:
Thank you Nicolas, merged!

babelouest:
Thanks for your help!

PR #1420, opened 2024-07-09
cmake: enable `pkg-config` to search for Libgcrypt, mbedTLS, wolfSSL

vszakats:
Also:
- set `MBEDTLS_INCLUDE_DIRS` for mbedTLS.
- stop marking `MBEDTLS_LIBRARIES` as advanced.
- sync up the content of the three custom `Find*` detection modules.
- formatting in `Findwolfssl.cmake`.

Closes #1420


vszakats:
Not a clear win. `pkg-config` finds stuff that's not actually useful:

- in a cross-build it finds the native package, then fails:
  https://github.com/libssh2/libssh2/actions/runs/9864949419/job/27240949458?pr=1420#step:8:14

- on macOS, it detects the package on its own prefix, which is good, but then it misses the headers of the package's dependency:
  https://github.com/libssh2/libssh2/actions/runs/9864949419/job/27240955176?pr=1420#step:9:17

vszakats:
It's complicated:

The common CMake pattern is to use `pkg-config` input as a path hint for CMake's native dependency lookup functions. This works for simple cases, but breaks when the `.pc` file contains extra/irregular stuff necessary for the dependency. Such case is Libgcrypt which has public headers depending on a header (`gpg-error.h`) from another project.

To better use `pkg-config` info, we'd need to use its output directly, instead of just using it as a path hint.

To do this we must decide if we want to check `pkg-config` first, or use it as a fallback.

Using it first most likely breaks manual configuration via `-DMBEDTLS_INCLUDE_DIR=` / `-DMBEDCRYPTO_LIBRARY=` (and similar) settings, because the detection will pick up any `pkg-config` hits first and never use the manual settings. (Unless force-disabling `pkg-config` at the same time)

Using it second bumps into a bunch of other issues that I haven't figured out yet.

All these methods steer the logic away from the "common" way, which might potentially cause more harm than good. Unless understanding all the possible angles in advance.

An extra complication is `pkg-config`s simplistic behaviour in cross-builds. Should we detect the problematic cases and skip using it automatically? Should we not? If we should, what would be the logic to detect the problematic (cross-build) cases?

PR #1421, opened 2024-07-11
expose mbedtls_pk_load_file for our use

orangepizza:
while it's moved to pk_internal, it won't removed in mbedtls 3.6 LTS
so it's safe to redeclare it on our side to find it

this is implementing emergency fix suggested from 
https://github.com/libssh2/libssh2/commit/2e4c5ec4627b3ecf4b6da16f365c011dec9a31b4#commitcomment-141379351


orangepizza:
I limited this patch only enabled in mbedtls 3.6.x , so any version newer than that 4.x or 3.7 etc, someone need to test/look at upstream code if patch is still appropriate

vszakats:
My concern is that the lazy fix just postpones this tech debt invisibly into the future.

Is there a way to reimplement this call using supported APIs?

orangepizza:
> My concern is that the lazy fix just postpones this tech debt invisibly into the future.
> 
> Is there a way to reimplement this call using supported APIs?

to do that it looks like you need to manually set ecdsa curve parameter from ecp.h (don't think it'd safe) or grab or write raw byte fileread function form somewhere else and tose it into _libssh2_mbedtls_ecdsa_new_private_frommemory  (this drops support for pem though)
from looks of it mbedtls_pk_load_file is just read file x and put it in data.

vszakats:
Using low-level crypto (or alternate internals?) would of course not be an ideal solution.

@gilles-peskine-arm suggested `mbedtls_pk_parse_key()`. Could this be useful?

orangepizza:
that returns mbedtls_pk_context, ecp.h needed to convert it into mbedtls_ecdsa_context (which is rename of mbedtls_ecp_keypair)
so if we manually touch mbedtls_ecdsa_context we need to handle ecp. 

https://github.com/Mbed-TLS/mbedtls/blob/f974bff38c958f1f3ca997253f002a7b179952ab/tf-psa-crypto/drivers/builtin/include/mbedtls/ecdsa.h#L67
https://github.com/Mbed-TLS/mbedtls/blob/development/tf-psa-crypto/drivers/builtin/include/mbedtls/ecp.h#L415

orangepizza:
tried mimicing what rsa function did, but ecdsa_context isn't something able to copied by public (no mbedtls_ecdsa_keypair_copy) 
I can just point *ctx into ec keypair inside pk_context structure, but doing so will cause leak on other members of pk_context structure (because no we can't clean pkey without nulling ecdsa

src/mbedtls.h
```diff
@@ -278,6 +278,12 @@
 
 #if LIBSSH2_ECDSA
 
+/* force expose internal mbedtls function for linking,
+   I expect this won't change internally until mbedtls 4.0 */
+#if MBEDTLS_VERSION_MAJOR == 3 && MBEDTLS_VERSION_MINOR >= 6
+int mbedtls_pk_load_file(const char *path, unsigned char **buf, size_t *n);
+#endif
+
```
vszakats, line 286:
```suggestion
/* Force-expose internal mbedTLS function */
#if MBEDTLS_VERSION_NUMBER >= 0x03060000
int mbedtls_pk_load_file(const char *path, unsigned char **buf, size_t *n);
#endif

```

Also move this block into `mbedtls.c` before the function where we are using this.

src/mbedtls.c
```diff
@@ -1292,7 +1292,7 @@ _libssh2_mbedtls_ecdsa_new_private(libssh2_ecdsa_ctx **ctx,
     unsigned char *data;
     size_t data_len;
 
-#if MBEDTLS_VERSION_NUMBER >= 0x03060000 && \
+#if MBEDTLS_VERSION_NUMBER >= 0x04000000 && \
     defined(_LIBSSH2_DISABLE_MBEDTLS36_PK_LOAD_FILE)
 
     /* FIXME: implement this functionality via a public API */
```
vszakats, line 1298: Wouldn't it be better to replace this hack with the FIXME comment, and let it fail when mbedTLS actually decides to remove this function completely?

If so, after a rebase on master, `_LIBSSH2_DISABLE_MBEDTLS36_PK_LOAD_FILE` references need to be deleted from  `.github/workflows/ci.yml`.

vszakats (changes requested):

vszakats:
I'm still bummed why mbedTLS is removing a function with no replacement. This combined with no notice or even line in the ChangeLog is causing unnecessary pain.

src/mbedtls.c
orangepizza: yes, I agree that it'd better fail to link when this function actually removed:

vszakats:
Thanks. Would you mind rebasing to master? And delete the `_LIBSSH2_DISABLE_MBEDTLS36_PK_LOAD_FILE` references I mentioned? those become unused after this patch.

orangepizza:
rebased: should I squash those too?

vszakats:
No need, thanks.

vszakats:
Merged now, thanks you for the PR!

PR #1423, opened 2024-07-11
Always check for null pointers before calling _libssh2_bn_set_word

MichaelBuckley:
Another small change. A while back, we submitted a change to check the return value of `_libssh2_bn_set_word` in the hope that it would fix a report from an open-source fuzzer. It did not. Our current theory is that the fuzzer is running out of memory, but we won't know for sure until this gets merged.

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -1054,12 +1054,12 @@ kex_method_diffie_hellman_group1_sha1_key_exchange(LIBSSH2_SESSION *session,
         key_state->g = _libssh2_bn_init();      /* SSH2 defined value (2) */
 
         /* Initialize P and G */
-        if(_libssh2_bn_set_word(key_state->g, 2)) {
+        if(!key_state->g || _libssh2_bn_set_word(key_state->g, 2)) {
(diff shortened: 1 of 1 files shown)


src/kex.c
```diff
@@ -1152,12 +1152,13 @@ kex_method_diffie_hellman_group14_key_exchange(LIBSSH2_SESSION *session,
...
         /* Initialize P and G */
-        if(_libssh2_bn_set_word(key_state->g, 2)) {
+        if(!key_state->g || _libssh2_bn_set_word(key_state->g, 2)) {
             ret = _libssh2_error(session, LIBSSH2_ERROR_ALLOC,
                                  "Failed to allocate key state g.");
             goto clean_exit;
         }
-        else if(_libssh2_bn_from_bin(key_state->p, 256, p_value)) {
+        else if(!key_state->p ||
+          _libssh2_bn_from_bin(key_state->p, 256, p_value)) {
```
vszakats, line 1161:
```suggestion
        else if(!key_state->p ||
                _libssh2_bn_from_bin(key_state->p, 256, p_value)) {
```

vszakats (approved):

PR #1426, opened 2024-07-14
Implement chacha20-poly1305@openssh.com

### LINKED ISSUE #584: Implement OpenSSH ChaCha20-Poly1305
**Describe the bug**
libssh2 does not implement `chacha20-poly1305@openssh.com`.

**To Reproduce**
libssh2 does not connect to a server implementing only `chacha20-poly1305@openssh.com`.

**Expected behavior**
libssh2 should support `chacha20-poly1305@openssh.com`.

Should be all systems with the latest `master`.  `git grep` shows no instances of `chacha` in the codebase.

**Additional context**
The only ciphers that libssh2 currently implements are CBC ciphers, which are known to be insecure (with a 2^18 attack effort), and AES modes.  Without hardware support for AES, it is either slow, or non-constant time (i.e., insecure) or both.  ChaCha20-Poly1305 runs at about 2 GB/s on an x86-64 processor and will perform similarly even on ARM processors without AES support, and it is usually extremely easy to implement in a constant-time way.  Modern versions of OpenSSL provide support for this primitive, as do many other crypto libraries.

MichaelBuckley:
Probably the biggest and potentially most controversial change we have to upstream.

Because earlier versions of OpenSSL implemented the algorithm before standardization, using an older version of OpenSSL can cause problems connecting to OpenSSH servers. Because of this, we use the public domain reference implementation instead of the crypto backends, just like OpenSSH does.

We've been holding this one for a few years. We were about to upstream it around the same time as aes128gcm landed upstream, and the two changes were completely incompatible. Honestly, it took me weeks to reconcile these two implementations, and it could be much better.

Our original implementation changed every crypt method to decrypt the entire message at once. the AESGCM implementation instead went with this firstlast design, where a firstlast paramater indicates whether this is the first or last call to the crypt method for each message. That added a lot of bookkeeping overhead, and wasn't compatible with the chacha public domain implementation.

As far as I could tell, OpenSSH uses the technique of decrypting the entire message in one go, and doesn't have anything like firstlast. However, I could not get out aes128gcm implementation to work that way, nor could I get the chacha implementation to work with firstlast, so I split it down the middle and let each implementation work differently. It's kind of a mess, and probably should be cleaned up, but I don't have the time to spend on it anymore, and it's probably better to have everything upstream.

Fixes #584
Closes #1426


MichaelBuckley:
Looks like the code is failing in a bunch of CI runners. I'll get it fixed up when I can.

vszakats:
I haven't looked at the logs, but using `__bounded__` (an OpenBSD gcc extension) will likely cause compiler warnings in other envs.

This might work to fix it:
```c
#if !defined(__bounded__)
#define __bounded__(x, y, z)
#endif
```
(if not, `!defined(__OpenBSD__)`, or more PP gymnastics)

Refs:
https://github.com/libressl/openbsd/commit/c42d1dfb3a73c033740c77876c3da35f1868ec0d
https://github.com/libressl/portable/blob/c8ca89425c255790a538c2e7a2520e69454c79ee/include/compat/stdint.h#L19-L21

MichaelBuckley:
It looks like there are issues beyond __bounded__ with the chacha reference implementation on some platforms.

I'm loathe to alter the reference implementation, because it's crypto code, but many of the changes look straightforward. I'll see what I can do.

src/poly1305.c
```diff
@@ -9,9 +9,7 @@
 /* $OpenBSD: poly1305.c,v 1.3 2013/12/19 22:57:13 djm Exp $ */
 
 #include <sys/types.h>
-#ifdef HAVE_STDINT_H
 # include <stdint.h>
```
vszakats: `stdint.h` is missing from earlier MSVC versions.
But `libssh2.h` already has the necessary compatilbility tricks added,
thus including it may fix this. It also includes `sys/types.h` already.

vszakats: Ah, you fixed already!

vszakats:
The distrocheck jobs miss the `chacha.h` headers from the `HHEADERS` list in `src/Makefile.inc`.

vszakats:
The bogus MSVC warnings in non-unity builds may go by including `libssh2_priv.h` into the offending sources.
(This includes `libssh2_setup.h` which has the pragma silencing this.)

vszakats:
@MichaelBuckley The distcheck and old-MSVC issues should be fixed now.

vszakats:
Green now.

vszakats (approved):

MichaelBuckley:
Thank you for fixing up the last problems. I've been swamped for time recently.

vszakats:
No worries, they were just some minor touches. Thanks for all your work with this.

vszakats:
@MichaelBuckley Thanks for this great addition! It's merged now.

PR #1427, opened 2024-07-22
packet: fix memory leak in `packet_x11_open()`

vszakats:
Reported-by: 尹麓鸣
Ref: https://github.com/LuMingYinDetect/libssh_defects
Closes #xxxx

---

/cc @LuMingYinDetect

### DIFF
--- a/src/packet.c
+++ b/src/packet.c
@@ -405,6 +405,7 @@ packet_x11_open(LIBSSH2_SESSION * session, unsigned char *data,
             rc = _libssh2_transport_send(session, x11open_state->packet, 17,
                                          NULL, 0);
             if(rc == LIBSSH2_ERROR_EAGAIN) {
+                LIBSSH2_FREE(session, channel);
                 return rc;
             }
             else if(rc) {


src/packet.c
```diff
@@ -405,6 +405,7 @@ packet_x11_open(LIBSSH2_SESSION * session, unsigned char *data,
             rc = _libssh2_transport_send(session, x11open_state->packet, 17,
                                          NULL, 0);
             if(rc == LIBSSH2_ERROR_EAGAIN) {
+                LIBSSH2_FREE(session, channel);
```
willco007, line 408: If we free `channel` here, what happens when it comes back in to this this function after EAGAIN is hit and the state is currently set to `libssh2_NB_state_created` which assumes channel is already allocated and usable?

willco007: Secondly, just freeing channel wouldn't be enough, you'd also need to free `channel->channel_type`. That said, I don't think this is a leak.

vszakats: Thanks @willco007!

vszakats:
Closing without merge.

PR #1428, opened 2024-07-22
Correctly initialize values

willco007:
Fix regression with commit from #1421

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -1294,8 +1294,8 @@ _libssh2_mbedtls_ecdsa_new_private(libssh2_ecdsa_ctx **ctx,
                                    const unsigned char *pwd)
 {
     mbedtls_pk_context pkey;
-    unsigned char *data;
-    size_t data_len;
+    unsigned char *data = NULL;
+    size_t data_len = 0;
 
     /* FIXME: Reimplement this functionality via a public API. */
     if(mbedtls_pk_load_file(filename, &data, &data_len))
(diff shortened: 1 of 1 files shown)


vszakats:
@willco007: It just dawned on me that these initialization were pretty much always missing from the code. The crashes you've seen _might_ have a different reason. Unless mbedTLS changed how the `mbedtls_pk_load_file` API behaves and now leaves these values uninitialized on failure.

This patch PR is fine either way.

Ref: 2e4c5ec4627b3ecf4b6da16f365c011dec9a31b4

willco007:
The crash I saw was reported by the fuzzer but marked as not reproducible, however it first appeared from the commit for #1421. It seems it just happened to hit the uninitialized variables on that run??

vszakats:
#1421 basically reverted the earlier 2e4c5ec4627b3ecf4b6da16f365c011dec9a31b4, with the manual declaration added for the most recent mbedTLS version.

It gets even more interesting because this was indeed wrong, but it was wrong all along before these commits: mbedTLS does indeed leaves these variables unitialized on failure, by looking at 3.5.0:
https://github.com/Mbed-TLS/mbedtls/blob/1ec69067fa1351427f904362c1221b31538c8b57/library/pkparse.c#L77-L94

One explamantion is that the manual declaration triggered the fuzzer to extend into this mbedTLS function.

Makes me think how many more of these may exist out of view of the fuzzer.

willco007:
I quickly scrubbed mbedtls.c and didn't see any more initialized variables, but I did find this error case #1430.

vszakats:
Nice!

PR #1433, opened 2024-07-25
test: drop ssh-dss openssh_server config

quaresmajose:
The ssh-dss was deprecated [1] in the OpenSSH 9.8p1

Support for ssh-dss, ssh-dss-cert-* host and user keys is disabled by default at run-time. These may be re-enabled using the instructions at https://www.openssh.com/legacy.html

[1] https://www.openssh.com/releasenotes.html

### DIFF
--- a/tests/openssh_server/sshd_config
+++ b/tests/openssh_server/sshd_config
@@ -1,4 +1,4 @@
 HostKeyAlgorithms +ssh-rsa
-PubkeyAcceptedKeyTypes +ssh-rsa,ssh-dss,ssh-rsa-cert-v01@openssh.com
+PubkeyAcceptedKeyTypes +ssh-rsa,ssh-rsa-cert-v01@openssh.com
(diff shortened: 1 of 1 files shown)


vszakats:
Sure, and this would fall over as soon as Debian testing upgrades to 9.8.
But then what to do with the CI tests that assume DSA support on the test server?

Shall we disable it by default in libssh2 and stop testing it _now_? Would you want to update the PR to that effect?

This was originally scheduled for September 2024, which is almost here, so doing it now probably doesn't make much difference. Also nobody ever commented on this schedule that I proposed, meaning either nobody cares, or everyone agrees, or nobody has seen it. The only issue I see this is that announcement never made it into a release, for people to see it and prepare for it.

quaresmajose:
This patch was mostly to fix one of the test we run on openembedded that start regressing with the update to OpenSSH 9.8p1 https://github.com/openembedded/openembedded-core/commit/1a2fef1d70c8083e65325dfa8a5cae7d8443951b

I believe we ran just a little of tests available in libssh2 and it was done in cross-compilation environment.
This is to say that I can try to fix and remove the remaining DSA support but it would be better to do that in some followup next step, because for that I need to become better understand the libssh2 code base and test infrastructure.

MichaelBuckley:
FWLIW, I didn't reply to the September 2024 schedule because it seemed fine to me. Dropping it earlier or later won't have any impact on the projects I work on.

vszakats:
@quaresmajose I'd prefer to handle this in a single commit, to avoid a broken CI disrupting other PRs.

vszakats:
An attempt to disable DSA by default: #1435

quaresmajose:
> @quaresmajose I'd prefer to handle this in a single commit, to avoid a broken CI disrupting other PRs.

@vszakats thanks for the fix

PR #1436, opened 2024-08-01
ssh2_exec: bugfix read loop when libssh2_channel_read() returns EAGAIN

hbujanda:
In a slow system, first call to libssh2_channel_read() returns EAGAIN, what was making the example exit without doing any read.


vszakats:
Wouldn't it also apply to these examples with a similar loop?:
scp_nonblock.c
sftp_RW_nonblock.c
ssh2_agent_forwarding.c

hbujanda:
> Member

Hi vszakats,
I guess yes, but I've just tested a limited ammount of examples... none of the ones you mention.

Do you want me to add to the pull-request the same fix without testing to the examples you mention?

vszakats:
It'd be best if you could patch them and run a test with the patch, would that be possible?

hbujanda:
I'm working with a device that is big endian, no filesystem and several other constrains... porting and testing an example is quite painfull.

example/ssh2_exec.c
```diff
@@ -262,7 +262,7 @@ int main(int argc, char *argv[])
                     fprintf(stderr, "libssh2_channel_read returned %ld\n",
                             (long)nread);
             }
-        } while(nread > 0);
+        } while(nread > 0 || nread == LIBSSH2_ERROR_EAGAIN);
```
vszakats, line 265: On a second look: This change makes the loop continue on `EAGAIN`. However the outer loop expects this value and the inner loop to exit on that value. On `EAGAIN` it calls `waitsocket()` and enters the inner loop again, otherwise exists.

Am I overlooking something?

vszakats:
Information is missing to decide on this PR. Closing this for now, please feel free to reopen.

PR #1443, opened 2024-08-12
session: support server banners up to 8192 bytes (was: 256)

### LINKED ISSUE #1442: session handshake fails on server banners larger than 256 bytes
**Describe the bug**
`libssh2_session_handshake()` returns -1 when server has banner message that exceeds the 256 characters allowed by `_LIBSSH2_SESSION.banner_TxRx_banner` as key exchange fails because the full banner hasn't been read.

**To Reproduce**
Configure OpenSSH server with very long banner message by adding
`VersionAddendum " Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit."`
to `sshd_config`

Connect to server using examples/ssh2

**Expected behavior**
`libssh2_session_handshake()` should be able to negotiate a connection. OpenSSH clients have no issues connecting to servers with very large banners.

 - OS and version: macOS 14.6.1 (23G93)
 - libssh2 version: 1.11.1_DEV (814a850c97b0d535341868b4aefdfe76179330e9)

**Additional context**
I have a [minimal fix](https://github.com/palmin/libssh2/commit/0b08e75d2010aa1ad7b8cb90743f3aacff6f4124) that keeps reading the server banner until reaching `\n` allocating new chunks of memory for session->remote.banner for each 256 bytes read.

Should I submit this as a PR?

palmin:
If server had banner exceeding 256 bytes there wasn't enough room in
`_LIBSSH2_SESSION.banner_TxRx_banner`. Only the first 256 bytes would be
read making the first packet read fail but also dooming key exchange as
`session->remote.banner` didn't include everything.

This change makes `banner_receive` keep reading until the banner
completes with `\n` by growing `session->remote.banner` in chunks of
256 bytes.

Fixes #1442

### DIFF
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -778,7 +778,7 @@ struct _LIBSSH2_SESSION
 
     /* State variables used in libssh2_banner_send() */
     libssh2_nonblocking_states banner_TxRx_state;
-    char banner_TxRx_banner[256];
+    char banner_TxRx_banner[8192];
     ssize_t banner_TxRx_total_send;
 
     /* State variables used in libssh2_kexinit() */


willco007:
This should probably be capped at some reasonable max length.

palmin:
RFCs mention no limits and I'm not sure if this should influence libssh2 but OpenSSH seems to have a 8192 character limit for each banner line with a maximum of 1024 lines before the `SSH-` one should appear.

https://github.com/openssh/openssh-portable/blob/1ec0a64c5dc57b8a2053a93b5ef0d02ff8598e5c/ssh.h#L103-L104

How should I proceed?

willco007:
Should we increase the buffer size to something like 1024 and call it a day? That gives a bunch of padding to RFC size and keeps it simple and capped?

MichaelBuckley:
I'd vote to make the buffer size the same as OpenSSH, otherwise we'll eventually get someone emailing support about why their 2 KB banner works in OpenSSH but not our apps. This already happened when a user asked for `SFTP_HANDLE_MAXLEN` to be increased beyond the RFC to match OpenSSH.

willco007:
@MichaelBuckley IDK, that is a very large amount over the RFC, and adding complexity, for seemingly no particular reason to do so. We could talk offline if you feel strongly about it. :)

palmin:
whether or not this ends up being
    `char banner_TxRx_banner[1024];`
or 
    `char banner_TxRx_banner[8192];`
it seems safe to close this PR, as it has been rejected for adding complexity, right?

src/session.c
```diff
@@ -107,20 +137,23 @@ static int
...
         session->banner_TxRx_state = libssh2_NB_state_created;
     }
     else {
         banner_len = session->banner_TxRx_total_send;
     }
 
-    while((banner_len < sizeof(session->banner_TxRx_banner)) &&
-          ((banner_len == 0)
-           || (session->banner_TxRx_banner[banner_len - 1] != '\n'))) {
+    while(1) {
```
vszakats, line 156:
```suggestion
    for(;;) {
```
seem preferred in src.

vszakats (dismissed):

vszakats:
Ah OK, missed the conversation. If bumping the buffer fixes the issue, it's a good alternative.

Could that still necessitate this patch, or such larger banners are too weird?

willco007:
> whether or not this ends up being `char banner_TxRx_banner[1024];` or `char banner_TxRx_banner[8192];` it seems safe to close this PR, as it has been rejected for adding complexity, right?

At the very least this PR would need max length truncation added to avoid DOS.

palmin:
> At the very least this PR would need max length truncation added to avoid DOS.

Made a change where the banner buffer is never grown above 8192 bytes making `banner_receive()` fail.

willco007:
Would it be enough to and simply change `banner_TxRx_banner[256]` to `banner_TxRx_banner[8192]`?

vszakats:
@palmin What do you think of "just" bumping `banner_TxRx_banner` size as @willco007 suggests?

palmin:
> @palmin What do you think of "just" bumping `banner_TxRx_banner` size as @willco007 suggests?

I like it and will adjust the PR to do this tomorrow.

vszakats (approved):

vszakats:
Thank you @palmin, merged!

PR #1450, opened 2024-09-06
cmake: always run the tests even when static libs are disabled

eli-schwartz:
The BUILD_STATIC_LIBS option is intended for the use case where users do not wish to build and install a static lib because it is not one of the artifacts they are interested in. The option should not be used to disable running the testsuite.

Instead, when tests are enabled, define the target. Just don't export or install it. Build systems build lots of things that aren't deliverable artifacts, and this becomes just one more of them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,11 @@ else()
   endif()
 endif()
 
+# Some tests are only built if BUILD_TESTING is deliberately passed on the command line.
+if(BUILD_TESTING)
+  set(BUILD_TESTING_EXPLICIT ON)
+endif()
+
 option(BUILD_EXAMPLES "Build libssh2 examples" ON)
 option(BUILD_TESTING "Build libssh2 test suite" ON)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -94,7 +94,7 @@ set(_sources ${CSOURCES} ${HHEADERS})
 # when building both static and shared library. On Windows, with certain
 # toolchains (e.g. MSVC) these libraries get the same by default, overwriting
 # each other. MinGW is not affected.
-if(WIN32 AND BUILD_STATIC_LIBS AND BUILD_SHARED_LIBS AND
+if(WIN32 AND BUILD_SHARED_LIBS AND
    NOT STATIC_LIB_SUFFIX AND NOT IMPORT_LIB_SUFFIX AND
    CMAKE_STATIC_LIBRARY_SUFFIX STREQUAL CMAKE_IMPORT_LIBRARY_SUFFIX)
   set(STATIC_LIB_SUFFIX "_static")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -103,7 +103,7 @@ endif()
 unset(_libssh2_export)
 
 # we want it to be called libssh2 on all platforms
-if(BUILD_STATIC_LIBS OR BUILD_TESTING)
+if(BUILD_STATIC_LIBS OR BUILD_TESTING_EXPLICIT)
   if(BUILD_STATIC_LIBS)
     list(APPEND _libssh2_export ${LIB_STATIC})
   endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -103,8 +103,10 @@ endif()
 unset(_libssh2_export)
 
 # we want it to be called libssh2 on all platforms
-if(BUILD_STATIC_LIBS)
-  list(APPEND _libssh2_export ${LIB_STATIC})
+if(BUILD_STATIC_LIBS OR BUILD_TESTING)
+  if(BUILD_STATIC_LIBS)
+    list(APPEND _libssh2_export ${LIB_STATIC})
+  endif()
   add_library(${LIB_STATIC} STATIC ${_sources})
   add_library(${PROJECT_NAME}::${LIB_STATIC} ALIAS ${LIB_STATIC})
   target_compile_definitions(${LIB_STATIC} PRIVATE ${PRIVATE_COMPILE_DEFINITIONS} ${_libssh2_definitions})
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -82,16 +82,12 @@ target_link_libraries(runner PRIVATE libssh2)
 foreach(_test IN LISTS DOCKER_TESTS STANDALONE_TESTS SSHD_TESTS)
   if(NOT ";${DOCKER_TESTS_STATIC};${STANDALONE_TESTS_STATIC};" MATCHES ";${_test};")
     set(_lib_for_tests ${LIB_SELECTED})
-  elseif(TARGET ${LIB_STATIC})
-    set(_lib_for_tests ${LIB_STATIC})
   else()
-    unset(_lib_for_tests)
-    message(STATUS "Skip test requiring static libssh2 lib: ${_test}")
+    set(_lib_for_tests ${LIB_STATIC})
   endif()
 
   # We support the same target as both Docker and SSHD test. Build those just once.
-  # Skip building tests that require the static lib when the static lib is disabled.
-  if(NOT TARGET ${_test} AND _lib_for_tests)
+  if(NOT TARGET ${_test})
     add_executable(${_test} "${_test}.c")
     target_compile_definitions(${_test} PRIVATE "${CRYPTO_BACKEND_DEFINE}")
     target_include_directories(${_test} PRIVATE
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -82,12 +82,17 @@ target_link_libraries(runner PRIVATE libssh2)
 foreach(_test IN LISTS DOCKER_TESTS STANDALONE_TESTS SSHD_TESTS)
   if(NOT ";${DOCKER_TESTS_STATIC};${STANDALONE_TESTS_STATIC};" MATCHES ";${_test};")
     set(_lib_for_tests ${LIB_SELECTED})
-  else()
+  elseif(TARGET ${LIB_STATIC})
     set(_lib_for_tests ${LIB_STATIC})
+  else()
+    unset(_lib_for_tests)
+    message(STATUS "Skip test requiring static libssh2 lib: ${_test}")
   endif()
 
   # We support the same target as both Docker and SSHD test. Build those just once.
-  if(NOT TARGET ${_test})
+  # Skip building tests that require the static lib if tests weren't explicitly
+  # enabled via the command line.
+  if(NOT TARGET ${_test} AND _lib_for_tests)
     add_executable(${_test} "${_test}.c")
     target_compile_definitions(${_test} PRIVATE "${CRYPTO_BACKEND_DEFINE}")
     target_include_directories(${_test} PRIVATE


vszakats:
This assumes that the only motivation/expectation to/for disabling static builds is to not install them. But isn't it also a common expection that `BUILD_STATIC_LIBS=OFF` actually disables building them? It saves a extra build pass, and some may use the results without install (and expect a static lib to not be there). Alao some may not necessarily want to run tests at all, or fine with enabling static if they want to. On the other hand it helps those who want to run tests but definitely don't want to publish a static lib.

(One downside could be eliminated by adding support for single-pass static/shared lib build.)

eli-schwartz:
> It saves a extra build pass,

I explicitly asked for tests which means I built a *lot* of passes that I don't need as a consumer. Building a static library needed by the tests is not a surprise if I'm already asking to build tons of test files.

> and some may use the results without install (and expect a static lib to not be there).

What's the meaning of that expectation? Why would it matter if there is an additional file that is not being used? The cmake config files already export separate shared vs static targets to link to, the presence of an additional file won't confuse cmake.

> Alao some may not necessarily want to run tests at all, or fine with enabling static if they want to.

Then if they don't run the tests, the static library won't be built?

I'm *not* fine with enabling static just to build tests. Most distros, including mine, have policies about forbidding static libraries. Unfortunately due to cmake being cmake, it's not possible to simply `rm` the static library as part of packaging because if you install a static library and annotate it with cmake's "export", then the *-cmake.config breaks and emits bizarre errors about not being able to find required files (it does this even if you try to link to libssh2_shared).

eli-schwartz:
> One downside could be eliminated by adding support for single-pass static/shared lib builds.

I'm not entirely certain what you mean by this. It's already possible to build libssh2 in either static-only or shared-only mode, I think. As long as you don't run the testsuite. Do you mean building just the one library file and not any other build artifacts? Maybe it suffices to add a "phony" target, in the style of `make shared-lib`?

eli-schwartz:
hi @vszakats, any news on whether this PR can be merged?

I can fix the merge conflicts if the changes themselves are reviewed and agreed to be useful improvements...

vszakats:
> I explicitly asked for tests which means I built a _lot_ of passes that I don't need as a consumer. Building a static library needed by the tests is not a surprise if I'm already asking to build tons of test files.

To be precise it's 3 tests out of the 44. The rest of tests don't need the static lib.

We have two settings here, with _defaults_ that are incompatible with each other
for these 3 tests.

Since these are default settings (tests = on, static libs = off), just reading their
values as in this PR will not detect user intent. In effect it means this PR force-
enables static libs for everyone, unless they start explicitly disabling tests (all of
them) to avoid this behaviour.

> > and some may use the results without install (and expect a static lib to not be there).
> 
> What's the meaning of that expectation? Why would it matter if there is an additional file that is not being used? The cmake config files already export separate shared vs static targets to link to, the presence of an additional file won't confuse cmake.

It means that if someone didn't ask for static libs it's unexpected to build and end
up with static libs.

> > Alao some may not necessarily want to run tests at all, or fine with enabling static if they want to.
> 
> Then if they don't run the tests, the static library won't be built?
> 
> I'm _not_ fine with enabling static just to build tests. Most distros, including mine, have policies about forbidding static libraries. Unfortunately due to cmake being cmake, it's not possible to simply `rm` the static library as part of packaging because if you install a static library and annotate it with cmake's "export", then the *-cmake.config breaks and emits bizarre errors about not being able to find required files (it does this even if you try to link to libssh2_shared).

Right, but someone else might not have these exact same requirements.

vszakats:
> > One downside could be eliminated by adding support for single-pass static/shared lib builds.
> 
> I'm not entirely certain what you mean by this. It's already possible to build libssh2 in either static-only or shared-only mode, I think. As long as you don't run the testsuite. Do you mean building just the one library file and not any other build artifacts? Maybe it suffices to add a "phony" target, in the style of `make shared-lib`?

I was thinking how to lessen the downsides of a static-libs=on scenario.

shared lib is on by default, and static lib is off. Static and shared lib are built in two
separate passes, so enabling static lib triggers and extra build pass. Building them
in a single pass (on platforms that supports it), can make building both "free".
Meaning no extra build pass, time, etc.

But, that's an aside and doesn't answer the question how to handle this situation.

eli-schwartz:
> Since these are default settings (tests = on, static libs = off), just reading their values as in this PR will not detect user intent. In effect it means this PR force-enables static libs for everyone, unless they start explicitly disabling tests (all of them) to avoid this behaviour.

It does detect user intent. 'Tests = on, static libs = off" means that you intend to not get static libs as an artifact.

> It means that if someone didn't ask for static libs it's unexpected to build and end up with static libs.

In my testing, they did not end up with static libs.

A static ***test*** lib was created but it had no effect on what the user ended up with. There is no static artifact anywhere that I can find. Can you describe to me how I can reproduce your results so I can fix it?

vszakats:
> I explicitly asked for tests which means I built a _lot_ of passes that I don't need as a consumer. Building a static library needed by the tests is not a surprise if I'm already asking to build tons of test files.

It's 3 tests out of the 44, we're talking about here. The rest of tests don't
need the static lib. 2 of them are also _simple_ tests, one named just that.

libssh2 has `BUILD_TESTING=ON` by default, so just by reading the value
as in this PR, isn't acting on user intent. What it's effectively doing is re-enabling
static libs for everyone by default, even if explicitly disabled, just not exporting it.

> > and some may use the results without install (and expect a static lib to not be there).
> 
> What's the meaning of that expectation? Why would it matter if there is an additional file that is not being used? The cmake config files already export separate shared vs static targets to link to, the presence of an additional file won't confuse cmake.

It means that if someone didn't ask for static libs it's unexpected to build and end
up with static libs.

> > Alao some may not necessarily want to run tests at all, or fine with enabling static if they want to.
> 
> Then if they don't run the tests, the static library won't be built?
> 
> I'm _not_ fine with enabling static just to build tests. Most distros, including mine, have policies about forbidding static libraries. Unfortunately due to cmake being cmake, it's not possible to simply `rm` the static library as part of packaging because if you install a static library and annotate it with cmake's "export", then the *-cmake.config breaks and emits bizarre errors about not being able to find required files (it does this even if you try to link to libssh2_shared).

Right, but someone else might not have these exact same requirements.

vszakats:
> > Since these are default settings (tests = on, static libs = off), just reading their values as in this PR will not detect user intent. In effect it means this PR force-enables static libs for everyone, unless they start explicitly disabling tests (all of them) to avoid this behaviour.
> 
> It does detect user intent. 'Tests = on, static libs = off" means that you intend to not get static libs as an artifact.

How does it detect user intent with `BUILD_TESTING=ON` being the default?

eli-schwartz:
> How does it detect user intent with `BUILD_TESTING=ON` being the default?

That default behavior currently builds 40+ tests before my PR.

I assumed that was intentional and haven't tried to change it. If users don't specify an intent, libssh2 guesses an intent for you (and that guess is that users want to build test programs).

Do you think I should change it?

vszakats:
Would this fix your use-case?:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 6dd8b57f..49cf97c6 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,10 @@ else()
   endif()
 endif()
 
+if(BUILD_TESTING)
+  set(BUILD_TESTING_EXPLICIT ON)
+endif()
+
 option(BUILD_EXAMPLES "Build libssh2 examples" ON)
 option(BUILD_TESTING "Build libssh2 test suite" ON)
 
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 78d18eaa..b58c0a70 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -99,7 +99,7 @@ set(_sources ${CSOURCES} ${HHEADERS})
 # when building both static and shared library. On Windows, with certain
 # toolchains (e.g. MSVC) these libraries get the same by default, overwriting
 # each other. MinGW is not affected.
-if(WIN32 AND BUILD_STATIC_LIBS AND BUILD_SHARED_LIBS AND
+if(WIN32 AND (BUILD_STATIC_LIBS OR BUILD_TESTING_EXPLICIT) AND BUILD_SHARED_LIBS AND
    NOT STATIC_LIB_SUFFIX AND NOT IMPORT_LIB_SUFFIX AND
    CMAKE_STATIC_LIBRARY_SUFFIX STREQUAL CMAKE_IMPORT_LIBRARY_SUFFIX)
   set(STATIC_LIB_SUFFIX "_static")
@@ -108,8 +108,10 @@ endif()
 unset(_libssh2_export)
 
 # we want it to be called libssh2 on all platforms
-if(BUILD_STATIC_LIBS)
-  list(APPEND _libssh2_export ${LIB_STATIC})
+if(BUILD_STATIC_LIBS OR BUILD_TESTING_EXPLICIT)
+  if(BUILD_STATIC_LIBS)
+    list(APPEND _libssh2_export ${LIB_STATIC})
+  endif()
   add_library(${LIB_STATIC} STATIC ${_sources})
   add_library(${PROJECT_NAME}::${LIB_STATIC} ALIAS ${LIB_STATIC})
   target_compile_definitions(${LIB_STATIC} PRIVATE ${PRIVATE_COMPILE_DEFINITIONS} ${_libssh2_definitions})
```

vszakats:
@eli-schwartz: Does https://github.com/libssh2/libssh2/pull/1450#issuecomment-2384054629 work for you?

eli-schwartz:
> Would this fix your use-case?:

This would work for me as we always pass -DBUILD_TESTING explicitly. I think it is an unnecessary complication but as a compromise option it is workable.

Do you want me to include such a change in my patch? Add it as a second patch?

vszakats:
> > Would this fix your use-case?:
> 
> This would work for me as we always pass -DBUILD_TESTING explicitly. I think it is an unnecessary complication but as a compromise option it is workable.
> 
> Do you want me to include such a change in my patch? Add it as a second patch?

Yes, please, feel free to push it to this PR.

vszakats:
Are you sure this is not necessary?
```diff
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -99,7 +99,7 @@ set(_sources ${CSOURCES} ${HHEADERS})
 # when building both static and shared library. On Windows, with certain
 # toolchains (e.g. MSVC) these libraries get the same by default, overwriting
 # each other. MinGW is not affected.
-if(WIN32 AND BUILD_STATIC_LIBS AND BUILD_SHARED_LIBS AND
+if(WIN32 AND (BUILD_STATIC_LIBS OR BUILD_TESTING_EXPLICIT) AND BUILD_SHARED_LIBS AND
    NOT STATIC_LIB_SUFFIX AND NOT IMPORT_LIB_SUFFIX AND
    CMAKE_STATIC_LIBRARY_SUFFIX STREQUAL CMAKE_IMPORT_LIBRARY_SUFFIX)
   set(STATIC_LIB_SUFFIX "_static")
```

This is necessary to avoid filename collision with MSVC between static and shared lib when both are built. I think it applies to this case, though haven't made tests.

eli-schwartz:
> This is used to avoid filename collision with MSVC between static and shared lib when both are built. I think it applies to this case, though haven't made tests.

That makes sense but I suggest a slightly shorter fix.

vszakats:
It is shorter, but IMO obscures the intent of that code snippet:
Why would we want to change the static lib suffix if we're not building
static libs? I'd stay explicit here for clarity.

eli-schwartz:
> It is shorter, but IMO obscures the intent of that code snippet: Why would we want to change the static lib suffix if we're not building static libs? I'd stay explicit here for clarity.

Because:
- it is more explicit to be shorter, as you spell out the actual conditions that need to be met in order to modify the naming style, rather than including the condition "and also, if the variable ends up being used"
- it is easier to read and understand what it's doing when you only calculate the things that matter
- it is less brittle to changes that have nothing to do with the conditions that are being met: you don't need to change logic in multiple places

The intent of the code snippet is only clear after my change. In git `master` the intent of the code snippet is confusing and difficult to read, and deceives the perception of the reader.

vszakats:
Hard disagree with introducing unrelated optimizations like this into this PR.

eli-schwartz:
Its not an optimization. I was mildly losing my mind at first, trying to figure out why in G-d's name that if block mattered.

vszakats:
Imagine finding out the issue in CI and coming up with a fix for that.

eli-schwartz:
I am not sure what you mean by that.

vszakats:
All in all this feels like the wrong/incomplete fix. Having reached the "G-d" moment above, I'm not motivated to deal with this, esp. a few days before a new release.

vszakats:
I mean by that that it took significant effort to figure out why some builds failed in odd ways in CI, then realizing it's the implib name == static libary name collision yet again, then coming up with that condition that you don't seem like. I'm sorry about that, and that `.lib` file are named how they are by the MSVC toolchain.

vszakats:
PR with https://github.com/libssh2/libssh2/pull/1450#issuecomment-2384054629 tweaked further: #1469

eli-schwartz:
I think I will just not report bugs at all anymore. The vast majority of the time I spent on this PR was actually fighting to convince you that there was even an issue in the first place, which was highly demotivating and set the tone for my future replies.

vszakats:
I'm aware of this naming rule, it's been so forever.

Yet it comes up unexpectedly at one point in every project. Here it happened while adding support to build both shared-static in one invocation. Needs to be traced, need to be identified. Then needs a unique fix every time, while considering backwards compatibility and all the rest. In nghttp2 this issue resulted in a multi-year long history of misunderstandings and back and forth until it settled. It's not fun for the first time, and it's not fun for the tenth time either.

I don't see a reason to touch it more than absolutely necessary. Making this `if()` shorter is not a goal, it's already ugly and it will likely stay ugly for a long time.

It's also not something to do with CMake. All build systems need to deal with this in some ways (if they support MSVC).

That said, if you feel like, try the proposed alternate PR. It should solve the issue, a little bit improved compared to the original proposal. Still not "clean". A clean solution would be to build an independent internal lib for tests only, that's never installed, but that's complex to implement and/or wasteful to run.

vszakats:
Feels the same from my end. 

It was a letdown that you did not recognize that we need to figure out intent, not just unexpectedly alter behaviour for those use-cases you did not seem to consider. Instead of hearing me out, you attack.

Then accept my suggestion anyway but with dropping a hunk for the critical condition for MSVC. When asking about it, you double down by adding an extra modification and fall into a bike-shed.

Perhaps more constructive and positive is to ask or propose or report wether the thing solves the issue you reported.

vszakats:
> I think I will just not report bugs at all anymore. The vast majority of the time I spent on this PR was actually fighting to convince you that there was even an issue in the first place, which was highly demotivating and set the tone for my future replies.

Can you point where I questioned the root issue?

Issue #1431, opened 2024-07-24
rcvd too much data: Occasional Duplicate Packet Sent Causing Corrupted Data


rolag:
**Describe the bug**
While writing various stress tests for a library I'm writing built on libssh2 I found numerous errors where the data
sent over ssh (scp, exec, direct-tcp) does not match the data sent.

Usually this is accompanied by the following error in the openssh-server logs:
```text
channel 0: rcvd too much data 2176, win 1836
```
The exact amount of data in each case varies and the number of times this error occurs per session varies too.

**To Reproduce**
I uploaded a [repo](https://github.com/rolag/debug_libssh2) containing a minimum example necessary to reproduce this reliably.

```bash
# Download libssh2
git submodule update --init --recursive
# Build docker image for openssh-server
cd docker && DOCKER_BUILDKIT=1 docker build -t debug_libssh2:latest -f sshd.dockerfile . && cd ..
cd bugs/
./compile.sh
while [ $? -eq 0 ]; do
    ./run.sh ./rcvd 2>&1 >log
done
```

The while loop will keep running the code in `bugs/rcvd.c` until the error occurs. Usually this happens within 10 minutes or so but on powerful machines I find I need to execute the following to simulate a slow machine and/or busy network:
```bash
sudo tc qdisc change dev docker0 root netem loss 2.0%
# To undo the above changes:
# sudo tc qdisc delete dev docker0 root netem loss 2.0%
```

I have example output logs of `rcvd.c` in the file `bugs/rcvd.log` and at the very end it shows channel 3 received more data than was sent (by exactly one libssh2 packet size `16809916 - 32700 = 16777216`).
```
channel 3: rcvd too much data 32700, win 4352/2097152 (excess 28348)
channel 3: rcvd too much data 4352, win 0/2097152 (excess 32700)
...
Channel 3 total_sent=16777216 total_read=16809916
```

**Expected behavior**
The data sent should match the data received.

**Version:**
 - OS and version: macOS (M1 sonoma), Linux x86_64 (Ubuntu, Debian)
 - libssh2 version: 1.11.0
 - crypto backend and version: OpenSSL (Linux: 1.1.1f-1ubuntu2.22, Linux: 3.3.1-r3, Mac: 3.3.0)
 - openssh-server: 9.2 (Debian Bookworm), 9.8 (Debian Sid)

**Additional context**
[...]

Issue #1397, opened 2024-05-28
Instable connections in case of SSH transfers over nonblocking sockets


klux21:
I faced instable TCP connections that are tunneled over SSH in case of transferring huge buffers using libssh2 1.10.0. 
I'm using libssh2 for SSH connections together with libressl and nonblocking sockets with possibly slow connections in usually single threaded applications. The problem was the usage of storage on the stack and relying on that data and it's address on the stack in subsequent calls once sending fails with EAGAIN or EWOULDBLOCK.
I did create a patch for version 1.10.0 and did see, that you have been working on a similar issue in 1.11.0 without resolving the reason of the problem. You'll find my patches of the sources in the attachment. I didn't yet verify the patch for version 1.11.0 and can't provide any test case. But it's just some few lines and not hard to understand. May be some of the maintainer of the library is able to have a look on the patches and to add those or better ones to the sources? Thanks in advance and for that library of course as well!

I have attached my patches and copies of the original files for version 1.10.0 and 1.11.0. I'm not insisting on any copyright of those little changes so everybody is free to use and modify that code according the copyright conditions of libssh2.
[libssh2-patches.tgz](https://github.com/libssh2/libssh2/files/15476613/libssh2-patches.tgz)

vszakats:
> [...], that you have been working on a similar issue in 1.11.0 without resolving the reason of the problem.

Can you say which existing commit and/or PR was meant to address this issue?

Also, would you mind opening a PR with your modifications so that we can review and get it tested by CI?

klux21:
I did only download the sources from https://libssh2.org only ( https://libssh2.org/download/libssh2-1.11.0.tar.gz ). 
I didn't checkout the sources from github and don't know git yet very well. A PR would take some time to learn git and 
I have no idea whether my changes are even suitable for the people here. That's why I attached the patches together with the original sources as a tgz. The sources should match the release tag libssh2-1.11.0  ( 1c3f1b7 ) here.

vszakats:
@klux21 Perhaps this?: 33dddd2f8ac3bc811719e6422e9bec6458f91de3 #720

vszakats:
Here's the suggested patch diffed from the tarball:
```diff
diff -r -u orig/src/libssh2_priv.h patch/src/libssh2_priv.h
--- orig/src/libssh2_priv.h	2023-05-04 08:32:38.000000000 +0200
[...]

PR #1454, opened 2024-09-12
Fix unstable connections over nonblocking sockets

rolag:
This is the patch described in #1397 

I've tested this branch and I cannot reproduce the bug in #1431 anymore with it.
I've provided a description of the bug in the commit message.

### DIFF
--- a/src/transport.c
+++ b/src/transport.c
@@ -919,7 +919,8 @@ send_existing(LIBSSH2_SESSION *session, const unsigned char *data,
            make the caller really notice his/hers flaw, we return error for
            this case */
         _libssh2_debug((session, LIBSSH2_TRACE_SOCKET,
-                       "Address is different, but will resume nonetheless"));
+                       "Address is different, returning EAGAIN"));
+        return LIBSSH2_ERROR_EAGAIN;
     }
 
     *ret = 1;                   /* set to make our parent return */


src/libssh2_priv.h
```diff
@@ -600,6 +600,8 @@ struct transportpacket
     unsigned char buf[PACKETBUFSIZE];
     unsigned char init[5];  /* first 5 bytes of the incoming data stream,
                                still encrypted */
+    unsigned char reply[5]; /* first 5 bytes of the outgoing data stream
```
willco007, line 603: This isn't used in this PR is it? Can be removed.

src/packet.c
```diff
@@ -1194,9 +1194,10 @@ _libssh2_packet_add(LIBSSH2_SESSION * session, unsigned char *data,
 
 
                 if(want_reply) {
-                    unsigned char packet[5];
+                    unsigned char *packet;
 libssh2_packet_add_jump_point4:
                     session->packAdd_state = libssh2_NB_state_jump4;
+                    packet = session->packet.reply;
```
willco007, line 1200: Where is `packet.reply` set?

src/transport.c
```diff
@@ -919,7 +919,8 @@ send_existing(LIBSSH2_SESSION *session, const unsigned char *data,
            make the caller really notice his/hers flaw, we return error for
            this case */
         _libssh2_debug((session, LIBSSH2_TRACE_SOCKET,
-                       "Address is different, but will resume nonetheless"));
+                       "Address is different, returning EAGAIN"));
+        return LIBSSH2_ERROR_EAGAIN;
```
willco007, line 923: Seems to me this is the only part of this PR that impacts this issue. I _believe_ the rest of the changes can be removed.

PR #1457, opened 2024-09-26
checksrc: update, check all sources, fix fallouts

vszakats:
update from curl:
https://github.com/curl/curl/blob/cff75acfeca65738da8297aee0b30427b004b240/scripts/checksrc.pl

---

w/o whitespace: https://github.com/libssh2/libssh2/pull/1457/files?w=1

### DIFF
--- a/.checksrc
+++ b/.checksrc
@@ -1,3 +1,0 @@
-disable FOPENMODE
-disable SNPRINTF
-disable TYPEDEFSTRUCT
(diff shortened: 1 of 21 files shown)


vszakats:
@monnerat: I touched OS/400 sources to make them pass checksrc. Can you take a look if it looks OK?

Also VMS, in case anyones sees who can check those.

vszakats:
@monnerat Thanks a lot for verifying, and your input. (Glad to not have broken anything!)

Agreed, writing wrappers is overkill. Adding definitions sounds like a good compromise. If the cutoff version can be figured out and checked for, even better. We can also live with it, and perhaps replace the FIXME with a comment, or your words above.

PR #1459, opened 2024-09-27
openssl: free allocated resources when using openssl3

bagder:
Reproduces consistently with curl test case 638


willco007:
There appears to be several other calls to `EVP_PKEY_get_bn_param()` which also need to be cleared.

PR #1461, opened 2024-09-27
ci/appveyor: reduce test runs (workaround for infrastructure permafails)

vszakats:
Jobs consistently fail to connect to the test server (run in GHA) since
2024-Aug-29:
https://ci.appveyor.com/project/libssh2org/libssh2/builds/50498393

There was an earlier phase of failures one month before that, that got
fixed by increasing the wait for the server in
bf3af90b3f1bb14cf452df7a8eb55cc9088f3e7f.

Thus, skip running tests in AppVeyor CI jobs, except: After some
experiments, it seems that running tests with the last OpenSSL job and
the last WinCrypt job _work_, which still leaves some coverage.
It remains to be seen how stable this is.

This is meant as a temporary fix till there is a solution to make all 
jobs run tests reliable like up until a few months ago.

---

Bumping up the timeout to 10 minutes doesn't help. But it turns out
by accident, that running the tests only in the last AppVeyor job _does_
pick up the server and run correctly. So it's not serving as a "canary"
or a signal after all.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,12 +110,6 @@ else()
   endif()
 endif()
 
-# Detect if testing was explicitly enabled and make sure to build the
-# static libssh2 lib in this case, to also build those tests that require it.
-if(BUILD_TESTING)
-  set(BUILD_TESTING_EXPLICIT ON)
-endif()
-
 option(BUILD_EXAMPLES "Build libssh2 examples" ON)
 option(BUILD_TESTING "Build libssh2 test suite" ON)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,12 @@ else()
   endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


mback2k:
Are the docker bridge containers CI jobs actually launched on GitHub in time?

vszakats:
It seemed to me they do launch, then remain idle. But my idea of "in time" is fuzzy and may be misinterpreting the visible logs.

vszakats:
After some trial and error it seems that the last OpenSSL and last WinCrypt job is enabled, they both finish fine. No idea what the logic is behind this, if any, or if this will stay this way or just a fluke.

vszakats:
Merged. This PR is not a solution to this problem, just an attempt to keep the CI usable till a real fix is found and we can re-enable tests for these Windows CI jobs.

PR #1462, opened 2024-10-02
Drop vsprintf() use on OS/400.

monnerat:
Follow-up to discussion in #1457

Plus e-mail address update.

Thanks in advance for merge.


vszakats:
Nice, thank you! It's merged.

PR #1469, opened 2024-10-13
cmake: build but don't install static lib in certain conditions

vszakats:
Building 3 tests require static libssh2 lib. Some may prefer not to
create the static lib, yet prefer to build all tests, including those
3 that require it.

Detect such intent by looking for an explicit `BUILD_TESTING=ON` and
`BUILD_STATIC_LIBS=OFF`, then build the static lib anyway but without
installing it.

Reported-by: Eli Schwartz
Fixes #1450

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,10 @@ else()
   endif()
 endif()
 
+if(BUILD_TESTING)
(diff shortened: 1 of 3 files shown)


vszakats:
On hold till after the release.

PR #1482, opened 2024-10-28
cmake: enable `pkg-config` for MinGW cross-builds

vszakats:


vszakats:
Not merging. Libgcrypt gets picked up in a default OS env. That's not helpful.

PR #1498, opened 2024-11-26
we parse/store method in agent identity list

elric1:
This makes it a little easier to work with, we can just strcmp(3) against a C string.


vszakats (changes requested):

vszakats:
This PR would still need addressing the issues raised, and perhaps find a way to avoid the ABI break.

Closing. Feel free to reopen if you'd like to continue working on it.

PR #1506, opened 2025-01-02
configure: fix `--without-lib*-prefix` when `lib*` is detected anyway

### LINKED ISSUE #1505: 1.11.1 fails to build with "--without-libssl-prefix"
**Describe the bug**

Updating to 1.11.1 resulted in the following build error:

```
    CCLD     libssh2.la
  ../libtool: line 7958: cd: no/lib: No such file or directory
  libtool:   error: cannot determine absolute directory name of 'no/lib'
  make[2]: *** [Makefile:593: libssh2.la] Error 1
  make[2]: Leaving directory '/C/_/libssh2/src/libssh2-1.11.1/src'
  make[1]: *** [Makefile:507: all] Error 2
  make[1]: Leaving directory '/C/_/libssh2/src/libssh2-1.11.1/src'
```

Removing `--without-libssl-prefix` from my build recipe makes things work again.

I assume some "no" value isn't checked in autotools and taken as the prefix?

Building 1.11.0 works fine in the same environment.

**To Reproduce**

```
  ./configure \
    --prefix=/usr \
    --enable-shared \
    --with-openssl \
    --with-libz \
    --without-libz-prefix \
    --without-libssl-prefix \
    --without-libgcrypt-prefix
  make
```

**Expected behavior**

build works

 - OS and version: Cygwin
 - libssh2 version: 1.11.1

vszakats:
Do not test the prefix when set to `no`.
(as with `--without-lib*-prefix`)

Before this patch this test was always made and when detected despite
the wrong prefix, the `no` prefix remained in `LDFLAGS` causing a build
failure later in `libtool`.

Fixes:
```
$ ../configure --without-libssl-prefix
[...]
../libtool: line 7756: cd: no/lib: No such file or directory
libtool:   error: cannot determine absolute directory name of 'no/lib'
make[2]: *** [libssh2.la] Error 1
```

Follow-up to d19b61907039554b8261a90898f2f31e1a706f38 #1384

Reported-by: Christoph Reiter
Fixes #1505

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -835,21 +835,21 @@ dnl
 dnl For convenience, $4 is expanded if [lib]$1 is found.
 
 AC_DEFUN([LIBSSH2_LIB_HAVE_LINKFLAGS], [
-  libssh2_save_CPPFLAGS="$CPPFLAGS"
-  libssh2_save_LDFLAGS="$LDFLAGS"
 
   if test "${with_lib$1_prefix+set}" = set; then
+    libssh2_save_CPPFLAGS="$CPPFLAGS"
+    libssh2_save_LDFLAGS="$LDFLAGS"
     CPPFLAGS="$CPPFLAGS${CPPFLAGS:+ }-I${with_lib$1_prefix}/include"
     LDFLAGS="$LDFLAGS${LDFLAGS:+ }-L${with_lib$1_prefix}/lib"
+    AC_LIB_HAVE_LINKFLAGS([$1], [$2], [$3])
+    CPPFLAGS="$libssh2_save_CPPFLAGS"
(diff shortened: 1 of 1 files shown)


vszakats:
Ah, still not good.

vszakats:
This fixes the `--without-*_prefix` case, but the same effect can still be
reproduced with `--with-libssl-prefix=bogus`. The issue is that when the
dependency check succeeds (because the dependency is on a system 
path or a path already added by a preceding check), the tested prefix gets
added to the `LDFLAGS` anyway, and `libtool` is breaking when it cannot
convert a relative, non-existing path an absolute one. But, an absolute path
is expected with `--with-*=` options, so this shouldn't be an issue in practice
outside of the `no` case fixed in this PR.

PR #1518, opened 2025-01-23
libssh2_trace.3: Update prototype

willco007:
Return value is int, not void.

Credit:
pyscripter


vszakats:
Thanks!

The nasty Cygwin breakage was fixed in d65c4b3987defa445b7ebe244da55d6faeb2a404.

willco007:
@vszakats I did notice the failure, that was a good one to get fixed!

PR #1534, opened 2025-02-02
cmake: make `libssh2-config` work with all TLS-backends

vszakats:
CMake:

- Find*: set `<modulename>_FOUND` for compatibility when found via
  `pkg-config`. E.g. `MbedTLS_FOUND`.
  `find_package_handle_standard_args()` sets both `<MODULENAME>_FOUND`
  and `<Modulename>_FOUND` when detecting the dependency. Some CMake
  code relies on this and 3rd-party code may rely on it too. Make sure
  to set the latter variant when detecting the dependency via
  `pkg-config`, where we don't call
  `find_package_handle_standard_args()`.

  CMake sets these variable to `TRUE` (not `ON` or `1`). Replicate this
  for compatibility.

- libssh2-config.cmake: inherit default `LIBSSH2_USE_PKGCONFIG`.
  Follow-up to a3aa6b4ca83bb8a3ed159612a44b25261dc9f537 #1525

- document variables consumed by `libssh2-config.cmake.in`.

- `libssh2-config.cmake`: fix to link to non-OpenSSL crypto backends.
  This is most likely not how this is supposed to be done, but better
  than failing.
  What's the canonical way to do this, and how OpenSSL and zlib does it
  is yet to be figured out.

- use `ZLIB::ZLIB` to reference zlib.
- use `IN ITEMS` where missed.
- harmonize variable dump output formats.

CMake `find_package` integration tests:

- extend to all crypto backends (was: OpenSSL).
- show libssh2 variables set by `find_package()`.
- stop building examples and tests for the consumed package.
  For performance.
- enable zlib, for coverage.
- be verbose when building the test targets.

ci/GHA:

- add packaged mbedTLS (2.x) build to Linux matrix.
- alphasort some tests.

Follow-up to d9c2e550caa4ad9052ad949d165ed85895ae5aee #1460
Follow-up to 82b09f9b3aae97f641fbcc2d746d2a6383abe857 #1322


vszakats:
/cc @talregev in case you're interested. I used your idea to inherit the pkg-config flag default from the build. It seems cleaner this way. I kept the option to override it at consumption-time. The namespace pollution remains unchanged with this PR, but the integration tests now do work.

CMake's built-in OpenSSL Find module implements this by keeping track of dependency details within `OpenSSL::*` target properties. 

For those, CMake has a mandatory `IMPORTED_LOCATION` property. This is a full path to a library. It's problematic for us, because we don't have such information when detecting dependencies via `pkg-config`. We have library paths and libnames instead. OpenSSL does it by feeding `pkg-config` detection results as input to the CMake-native-detection, but that way most of the `pkg-config` information is ignored and lost.

If targets supported link paths and lib name properties as first class citizens, it'd be a nice and clean way to store all detection results as properties, instead of rolling them in custom variables (that pollute the namespace). Though even with `TARGET::TARGET` names, pollution remains an issue, albeit a lesser one.

Maybe this mandatory property can be reconstructed with a fair chance of working most of the time, but it seems shaky.

```
CMake Warning (dev) in CMakeLists.txt:
  IMPORTED_LOCATION not set for imported target "WolfSSL::WolfSSL".
```

talregev:
@dg0yt
FYI

dg0yt:
> CMake has a mandatory `IMPORTED_LOCATION` property. This is a full path to a library. It's problematic for us, because we don't have such information when detecting dependencies via `pkg-config`. We have library paths and libnames instead. OpenSSL does it by feeding `pkg-config` detection results as input to the CMake-native-detection, but that way most of the `pkg-config` information is ignored and lost.

- `IMPORTED_LOCATION` isn't needed for `INTERFACE` targets.
- CMake would actually use `IMPORTED_LOCATION_<CONFIG>` in some cases.
- `IMPORTED_LOCATION` isn't the right property for Windows import libs of imported `SHARED` targets.

> pollute the namespace

Mitigated by prefixing with the package name, the only prefix the package may take ownership of.

vszakats:
I tried `INTERFACE`, but the manual says it's meant for header-only packages,
which wasn't reassuring. Then it turned out it doesn't support `::`-style target names
(temply worked around by using `__`).  Then got an odd error, worked around
with appending `IMPORTED` to `add_library`. After this it allows `::` again.
Then it still fails because it seems to ignore the libdir set via
`INTERFACE_LINK_DIRECTORIES`, and doesn't find `-lwolfssl`.

Do you have a hint where to look?

(`INTERFACE` is also not compatible with our `.pc` file assembly logic, but
that's hopefully a lighter issue.)

vszakats:
Continuing the `INTERFACE` experiments and this discussion thread in #1535.

PR #1535, opened 2025-02-03
cmake: make Find modules use `INTERFACE`

vszakats:
- move dependency properties (libs, libdirs, C flags, header dirs,
  pkg-config module names) from global lists to imported target
  `INTERFACE` properties. Rework FInd modules to return their results
  like this and update the libssh2 build process to use it. It makes
  Find modules re-usable from the cmake-config script by libssh2
  consumers, to integrate with libssh2 dependencies.

- define libssh2 dependencies as "imported targets" by the name:
  `libssh2::<depname>`, e.g. `libssh2::libgcrypt`.

- cmake-config: add fall-back logic for CMake without
  CMP0099 (v3.17 2020-03-20) to set lib directories.

- generate `libssh2.pc` based on imported target properties (instead of
  global lists).

- add target property dump debug function.

- ci/GHA: also test cmake integration on macOS.

Follow-up to 96d7f404e7d15425ae69e364a816a40d9ad8001e #1534

---

w/o ws https://github.com/libssh2/libssh2/pull/1535/files?w=1

Continuing #1534 from:
https://github.com/libssh2/libssh2/pull/1534#issuecomment-2629659875

- [x] move raw C flags into target props.
  - [ ] drop include dirs also passed as CFLAGS (as `-I`, not `-isystem`) (when detected via `pkg-config`)
- [x] figure out how to pass libdirs to consumer projects.
- [x] how to store dependency version as target prop?
- [x] how to store pc module name as target prop?
- [x] ideally make all Find result variables private (lowercase-underscore).
- [x] finalize target names. For now it's: `libssh2::<DependencyName>`.
- [x] assemble pc module list in a central place like libdirs/libs, for `libssh2.pc`.
  It can now be extracted from target props (TODO: OpenSSL, ZLIB).
- [x] add integration tests on macOS.
- [x] split off non-strictly related commits.


vszakats:
The fallout mentioned in https://github.com/libssh2/libssh2/pull/1534#issuecomment-2630699606 doesn't replicate in CI, presumably because it's on Linux where the dependency is at a default, system, location. While in my env it requires the `-L`, and it's not there.

The rest looks about fine. It's basically moving the Find module custom result variables into (mostly) standardized target properties. This makes it possible to let CMake apply these properties automatically to targets. It allows to set custom properties, which I used for the `pkg-config` module name.

dg0yt:
Wrt https://github.com/libssh2/libssh2/pull/1534#issuecomment-2630699606:

On target names and namespaces:  
AFAICT namespace can only be used for `IMPORTED` targets and for `ALIAS` targets, i.e. for targets which are not direct citizens of your project.
This also applies to `INTERFACE` libraries: If they are owned by your project, you can't add a namespace directly. (But you can export or alias them, adding namespace in *these* steps.)

On `INTERFACE_LINK_DIRECTORIES` for `INTERFACE` targets:
Interface targets must carry interface link *libraries* (*targets*) in order to resolve transitive usage requirements. So they must export them. Now the question is how much CMake really generalized the concept of carrying transitive usage requirements vs. specifically targeting header-only libraries. I admit that I don't know the answer.

vszakats:
I tried looking at how to export them, but so far no results. Probably the next is peeking into the source code. I don't have high hopes. It's sad because otherwise the concept makes dep handling on the whole more organized.

My impression is that CMake's concept of lib deps is rooted from Windows, and this abstraction doesn't work well elsewhere.

vszakats:
Extended integration tests to macOS, where the issue replicates.

A year 2017 thread complaining about this: https://cmake.org/pipermail/cmake/2017-August/066086.html
The reaction (or lack thereof) from CMake also isn't giving much hope.

Found this policy describing something close to what we experience here:
https://cmake.org/cmake/help/latest/policy/CMP0099.html
Requires CMake 3.17 (2020-03-20):

It needs to be set by the consumer project:
```cmake
if(POLICY CMP0099)  # new in CMake 3.17, defaults to OLD
  cmake_policy(SET CMP0099 NEW)
endif()
```
or opt-in via `...3.17` in `cmake_minimum_required()`.

Also it turns out `INTERFACE_LINK_DIRECTORIES` requires CMake 3.13 (2018-10-21).
(We aim for 3.7 (2016-11-11), as in curl, for old-linux compatibility.)
This makes this PR need more workarounds for old CMake versions.

I'll now push a terrible manual workaround to fill in for `CMP0099`.

vszakats:
I tend to consider the namespace pollution issue unresolvable.

CMake is a hot mess of namespace pollution. Avoiding it fully seems
impossible, at best impractical and unusual. Some of the modifications would
break compatibility with existing control variables, which had broken
namespacing rules a decade+ ago.

In theory `LIBSSH2_MBEDTLS_INCLUDE_DIR` sounds reasonable.
In practice, less so. Some name collisions are probably used in accidental,
but useful ways.

As a fan of namespacing, I'm calling defeat here. Instead of inflicting 
a lot of pain on everyone by spilling countless more hours into this
to break things, I'd rather go by fixing such issues if explicitly reported.

Other than that, do what's reasonably possible, without going overboard.

vszakats:
With the changes made after my last comment, that is moving module
properties under namespaced imported targets, I think this should do
the trick to not pollute the namespace more than necessary.

I plan to merge this soon.

PR #1560, opened 2025-03-17
Bug fix for gen_publickey_from_rsa

skuodi:
This PR fixes bugs in the current implementation so that a valid key is generated in SSH public key binary format.

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -727,10 +727,13 @@ gen_publickey_from_rsa(LIBSSH2_SESSION *session,
     _libssh2_htonu32(p, e_bytes);
     p += 4;
     mbedtls_mpi_write_binary(&rsa->MBEDTLS_PRIVATE(E), p, e_bytes);
+    p += e_bytes;   /* Increment write index after writing to buffer */
+    n_bytes++;      /* Add 1 to bignum size */
 
     _libssh2_htonu32(p, n_bytes);
     p += 4;
     mbedtls_mpi_write_binary(&rsa->MBEDTLS_PRIVATE(N), p, n_bytes);
+    p += n_bytes;   /* Increment write index after writing to buffer*/
 
     *keylen = (size_t)(p - key);
     return key;


src/mbedtls.c
```diff
@@ -727,10 +727,13 @@ gen_publickey_from_rsa(LIBSSH2_SESSION *session,
     _libssh2_htonu32(p, e_bytes);
     p += 4;
     mbedtls_mpi_write_binary(&rsa->MBEDTLS_PRIVATE(E), p, e_bytes);
+    p += e_bytes;   /* Increment write index after writing to buffer */
+    n_bytes++;      /* Add 1 to bignum size */
```
willco007, line 731: If you look at the openssl implementation both `e_bytes` and `n_bytes` size are incremented + 1.  They are incremented before `len` is calculated so we don't overflow the buffer.  This is incrementing `n_bytes` after `len` is calculated, it should be before for safety.

skuodi: An interesting observation is that adding 1 to both `e_bytes` and `n_bytes` generates a completely different public key.

I created a sample program to test this [rsa_genkey_ssh.c](https://gist.github.com/skuodi/e67c4f65149a23c41f948f0a01728545), placed it in the `programs/pkey` directory of [Mbed-TLS/mbedtls](https://github.com/Mbed-TLS/mbedtls), added it to the directory's [CMakeLists.txt](https://github.com/Mbed-TLS/mbedtls/blob/development/programs/pkey/CMakeLists.txt) and run the following command from the `mbedtls/build` directory  to compare the generated public key with that from OpenSSH's `ssh-keygen`:

```sh
make                                                              &&\
./programs/pkey/rsa_genkey_ssh 4096 id_rsa                        &&\
sudo chmod 400 id_rsa                                             &&\
ssh-keygen -t rsa -b 4096 -C '' -f id_rsa -N '' -y > keyfile.pub  &&\
diff id_rsa.pub keyfile.pub
```
Incrementing both `e_bytes` and `n_bytes` gives:
```

  . Seeding the random number generator.... ok
  . Generating the RSA key [ 4096-bit ].... ok
  . Checking public/private key validity.... ok
  . Exporting the RSA parameters.... ok
  . Writing public key to string(PEM format).... ok
  . Writing private key to string(PEM format).... ok
  . Storing keys to filesystem....
  . Private key 'id_rsa' exists, deleting.... ok
  . Wrote 'id_rsa' (3244 bytes).... ok
  . Public key 'id_rsa.pub' exists, deleting.... ok
  . Wrote 'id_rsa.pub' (725 bytes).... ok

1c1
< ssh-rsa AAAAB3NzaC1yc2EAAAAEAAEAAQAAAgEAxXYgPAvGrdR6GY1Vcns5HQqY0/DxnqrPQ9XUZ2WOQ3SafCI9/gwhrPN3ZzsuZCk3uVveqISAgwmCXucn6kGiAW4tZd4gJhl12iPLGigqe7bQ1SMAbUBY8AOxa7HTyp1pRMsiz2zZGcNLs9eoF+D584NyXEAT7MJ60XLRy+1pS+mfKH6yh6pALdiQM6YWMbrx3N1vW/fk+00VpvF7VybL/LWOha9LGELgg80TrxQvBzFYIhkfFj/nnz6w+g0ZInMEY1u1k+0J9DhRJee0leRh0d3WOz7kIOypScvQMbE4nGYpViNffQ5m84oXkDbOgdj6qTxhMgQy/dc3h4FQMLDplnaEFJzbJjKhzUOj5VVEnn8iPB2Sv1OpvCw5BD6TcsKo61pDp8NR2W/2dvZNSCBDq9RBP4Uz+630nn59qQXnZrM0jwUBv6eOO3SjDUBdQkrlJSdq9+Byd+G/p+TKasX6+vxe6xOzu7oMnd/0zlMuaH45CkGoETd2pnvBRCKp4fBWrY/61+tLRULaOr8KUfjFuz09d0UEb79K9G859n6O0HkFNxe6rA1K3z1KLFwmA4YrUZxZhuhHiimfxCDDnXs+mXNcom+VaY+34FO6DW35dnNtVVv/C/pBvfgUB5feRBSV5/2aj37Odvf2MNt2//sSg5tmByz/Wi3lC6O3gj9g4TM=
---
> ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDFdiA8C8at1HoZjVVyezkdCpjT8PGeqs9D1dRnZY5DdJp8Ij3+DCGs83dnOy5kKTe5W96ohICDCYJe5yfqQaIBbi1l3iAmGXXaI8saKCp7ttDVIwBtQFjwA7FrsdPKnWlEyyLPbNkZw0uz16gX4Pnzg3JcQBPswnrRctHL7WlL6Z8ofrKHqkAt2JAzphYxuvHc3W9b9+T7TRWm8XtXJsv8tY6Fr0sYQuCDzROvFC8HMVgiGR8WP+efPrD6DRkicwRjW7WT7Qn0OFEl57SV5GHR3dY7PuQg7KlJy9AxsTicZilWI199DmbziheQNs6B2PqpPGEyBDL91zeHgVAwsOmWdoQUnNsmMqHNQ6PlVUSefyI8HZK/U6m8LDkEPpNywqjrWkOnw1HZb/Z29k1IIEOr1EE/hTP7rfSefn2pBedmszSPBQG/p447dKMNQF1CSuUlJ2r34HJ34b+n5Mpqxfr6/F7rE7O7ugyd3/TOUy5ofjkKQagRN3ame8FEIqnh8Fatj/rX60tFQto6vwpR+MW7PT13RQRvv0r0bzn2fo7QeQU3F7qsDUrfPUosXCYDhitRnFmG6EeKKZ/EIMOdez6Zc1yib5Vpj7fgU7oNbfl2c21VW/8L+kG9+BQHl95EFJXn/ZqPfs529/Yw23b/+xKDm2YHLP9aLeULo7eCP2DhMw==
``` 

Incrementing only `n_bytes` gives:
```

  . Seeding the random number generator.... ok
  . Generating the RSA key [ 4096-bit ].... ok
  . Checking public/private key validity.... ok
  . Exporting the RSA parameters.... ok
  . Writing public key to string(PEM format).... ok
  . Writing private key to string(PEM format).... ok
  . Storing keys to filesystem....
  . Private key 'id_rsa' exists, deleting.... ok
  . Wrote 'id_rsa' (3244 bytes).... ok
  . Public key 'id_rsa.pub' exists, deleting.... ok
  . Wrote 'id_rsa.pub' (725 bytes).... ok

```

willco007: That is interesting, I wonder why OpenSSL is padding both of them out while this doesn't appear to need to. I'm not familiar enough with the format to understand why and the RFC isn't super helpful.

skuodi: I created a similar sample program for OpenSSL, [EVP_PKEY_RSA_keygen_ssh.c](https://gist.github.com/skuodi/88be74a75df4ffda57042451674e4091) placed it in the `/demos/pkey` directory of [openssl/openssl](https://github.com/openssl/openssl), added it to the directory's [Makefile](https://github.com/openssl/openssl/blob/master/demos/pkey/Makefile) and run the following command to verify the output against OpenSSH's `ssh-keygen`:

```
make                                                              &&\
./EVP_PKEY_RSA_keygen_ssh 4096 id_rsa                             &&\
sudo chmod 400 id_rsa                                             &&\
ssh-keygen -t rsa -b 4096 -C '' -f id_rsa -N '' -y > keyfile.pub  &&\
diff id_rsa.pub keyfile.pub
```

The output matches that of `ssh-keygen` when both `e_bytes` and `n_bytes` are incremented by 1.
The general structure of the SSH public key binary that `gen_publickey_from_rsa` generates consists of the type name, exponent and modulus as explained in this [brief blog post](https://coolaj86.com/articles/the-ssh-public-key-format/), `e` being our exponent (commonly 65537) and `n` being our modulus; two of the RSA _bignum_ s generated during key generation.

Both the calls to MbedTLS and OpenSSL functions return identical values at the assignment of `e_bytes` and `n_bytes` but the difference is in the implementation of how these bytes are serialized into the key buffer. In [openssl.c](https://github.com/libssh2/libssh2/blob/master/src/openssl.c) this is done in [write_bn](https://github.com/libssh2/libssh2/blob/df0563a85732fd9a33bbb116d8c07ca18ba6af38/src/openssl.c#L225) which first skips 4 bytes for the bignum length and 1 padding byte set to 0, writes the bytes of the bignum in big endian, then calculates and populates the value of bignum length.

 `write_bn` then checks the if the MSBit of the bignum (big endian encoding means MSByte is the first byte in the buffer) is set in which case it returns with the buffer pointer advanced by the number of bytes written, leaving the 1-byte padding between the length and the bignum MSByte. But if the MSBit of the bignum is not set, all the bytes of the bignum are `memove`d backward one byte within the buffer, effectively undoing the prepended padding. That is how `e_bytes` and `n_bytes` are assigned different values but both implementations of `gen_publickey_from_rsa` generate the same key.

I presume the padding is meant to handle a possible overflow but without any more information from official spec, it's no more than a guess. However, if the implementation of `gen_publickey_from_rsa` in [mbedtls.c](https://github.com/libssh2/libssh2/blob/master/src/mbedtls.c) was adapted from the one in [openssl.c](https://github.com/libssh2/libssh2/blob/master/src/openssl.c), I take it that `write_bn` should also have been adapter along with it to conform to the OpenSSL implementation.

willco007: I'm satisfied with your investigation, thanks! Seems like this is good to go.

PR #1563, opened 2025-03-22
tests: improve `SSHD` default value

vszakats:
Instead of going for a hardcoded `/usr/sbin/sshd`, use `command -v`
to figure out the default `sshd` executable.


pghmcfc:
Shouldn't that have been checking the SSHD variable rather than the SSH variable?

vszakats:
It very well should have. Thanks for spotting this! Fix → #1568

Issue #1580, opened 2025-04-22
Connection fails when using one encryption method with integrated MAC and one without


chrisgch:
**Describe the bug**
When setting one encryption method with libssh2_session_method_pref to
chacha20-poly1305@openssh.com (with external MAC)
and the other
aes256-gcm@openssh.com (with integrated MAC)
the library either fails to connect or crashes with an exception.

**To Reproduce**
Trying to connect with for example these methods fails:
Key exchange method: curve25519-sha256 (doesn't really matter)
Host key method: ecdsa-sha2-nistp256 (doesn't really matter)
Encryption method (client to server): chacha20-poly1305@openssh.com
Encryption method (server to client): aes256-gcm@openssh.com
MAC (message authentication code) method (client to server): hmac-sha2-256-etm@openssh.com
MAC (message authentication code) method (server to client): INTEGRATED-AES-GCM
(also fails when encryption methods reversed)

**A recipe or example code that reproduces the problem?**
Code called:
```
session = libssh2_session_init_ex(myalloc,myfree,myrealloc,ConnectSettings);
libssh2_session_method_pref(session,LIBSSH2_METHOD_CRYPT_CS,"chacha20-poly1305@openssh.com");
libssh2_session_method_pref(session,LIBSSH2_METHOD_CRYPT_SC,"aes256-gcm@openssh.com");
```
Then try to connect to the server with libssh2_session_startup.

**A stack trace from a crash**
Microsoft Visual C++ Runtime Library
Assertion failed!

Program: libssh2.dll
src\transport.c
Line: 139

Expression: (len % blocksize) == 0

**Expected behavior**
Connection should work even when using different encryption methods.

Reason for error: CRYPT_FLAG_L is used while receiving in functions decrypt() and fullpacket(), should be CRYPT_FLAG_R (remote method) instead because the functions are used for receiving packets.
And in function send_existing() both calls to CRYPT_FLAG_R should be replaced with CRYPT_FLAG_L.

**Version (please complete the following information):**
 - OS and version: Windows 11
 - libssh2 version: 1.11.1 or master
 - crypto backend and version: OpenSSL

**Additional context**
Here is the code that needs to be changed in transport.c:
Receive side:
```
    if(!CRYPT_FLAG_L(session, PKTLEN_AAD))
       assert((len % blocksize) == 0);
```
-> CRYPT_FLAG_R

```
[...]

PR #1582, opened 2025-04-24
fix for issue 1580

chrisgch:
This is a fix for my bug report here:
https://github.com/libssh2/libssh2/issues/1580
Connection fails when using one encryption method with integrated MAC and one without


MichaelBuckley:
I have tested this fix and found that it indeed solves the connection problem. The example in the original report showed that there was a problem when the client to server method had an integrated MAC and the server to client method did not. I also tested in the case where the client to server method does not have an integrated MAC but the server to client method does. Both cases were able to connect successfully with this patch. I found no cases where this patch broke any other configuration.

PR #1591, opened 2025-04-28
cmake: test integration with old cmake (v3.11.4 2018-03-28)

vszakats:
It was an exercise to run old cmake versions in CI and in the test suite.

It also revealed that 3.7.2 2017-01-13 is too old to consume libssh2 via
`find_package()` due to:
```
CMake Error at bld-libssh2/_pkg/lib/cmake/libssh2/libssh2-config.cmake:35 (add_library):
  add_library cannot create ALIAS target "libssh2::libssh2" because target
  "libssh2::libssh2_shared" is IMPORTED.
Call Stack (most recent call first):
  CMakeLists.txt:27 (find_package)
```
The mitigation for this issue requires 3.11.

Also:
- rename a few existing envs to use the `TEST_` prefix.
- make the `find_package` test provider stage verbose.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,7 +123,7 @@ jobs:
           export CMAKE_CONSUMER="$(cat "${HOME}/old-cmake-path.txt")"
           [[ '${{ matrix.image }}' = *'macos'* ]] && export CFLAGS='-arch arm64'
           [[ '${{ matrix.image }}' = *'windows'* ]] && options+=' -DCMAKE_C_COMPILER=x86_64-w64-mingw32-gcc'
-          [[ '${{ matrix.image }}' = *'windows'* ]] && PATH='/usr/bin;/c/Program Files/Git/bin'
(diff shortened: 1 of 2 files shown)


vszakats:
Re the 32-bit Ubuntu libgcrypt build fallout:

Maybe related?:

libgcrypt20 (1.10.3-2build1) noble; urgency=high
```
  * No change rebuild for 64-bit time_t and frame pointers.

 -- Julian Andres Klode <juliank@ubuntu.com>  Mon, 08 Apr 2024 18:03:56 +0200
```
https://changelogs.ubuntu.com/changelogs/pool/main/libg/libgcrypt20/libgcrypt20_1.10.3-2build1/changelog

Logs say that this revision before and after, suggesting this is also unrelated:
good: https://github.com/libssh2/libssh2/actions/runs/14691706850/job/41227891749
bad: https://github.com/libssh2/libssh2/actions/runs/14708495487/job/41274562010

PR #1592, opened 2025-04-28
cmake: fix copy-paste typo in integration tests for old cmake

vszakats:
Follow-up to 77df767784f6336dfea166b9202b6be98595b833 #1589


vszakats:
For no apparent reason, `CI / linux (clang, i386, Libgcrypt, autotools, ON, --disable-static` started failing today.
There was no related in the repo. Runner image is the same, installed package versions are the same.

Same job passes with amd64, and also passes with OpenSSL i386.
It suggests something libgcrypt 32-bit-related.

PR #1597, opened 2025-05-07
Fix: Null-dereference in ssh2_client_fuzzer

Shubham7-1:
This fix ensures that the ssh2_client_fuzzer target no longer crashes due to null-dereference errors, improving the stability and reliability 

Issue Summary:
The vulnerability occurs due to improper handling of the session pointer in the EXIT_LABEL section of the ssh2_client_fuzzer.cc file. Specifically, when libssh2_session_init fails or libssh2_session_handshake encounters an error, the session pointer remains NULL. Subsequent calls to libssh2_session_disconnect or libssh2_session_free result in a null-dereference.

issue link: https://issues.oss-fuzz.com/issues/393411531

### DIFF
--- a/tests/ossfuzz/ssh2_client_fuzzer.cc
+++ b/tests/ossfuzz/ssh2_client_fuzzer.cc
@@ -97,6 +97,5 @@ EXIT_LABEL:
         close(socket_fds[1]);
     }
 
-
     return 0;
 }
--- a/tests/ossfuzz/ssh2_client_fuzzer.cc
+++ b/tests/ossfuzz/ssh2_client_fuzzer.cc
@@ -90,8 +90,13 @@ EXIT_LABEL:
 
     libssh2_exit();
 
-    if(socket_fds[0] != -1) close(socket_fds[0]);
-    if(socket_fds[1] != -1) close(socket_fds[1]);
+    if (socket_fds[0] != -1) {
+        close(socket_fds[0]);
+    }
+    if (socket_fds[1] != -1) {
+        close(socket_fds[1]);
+    }
+
 
     return 0;
 }
--- a/tests/ossfuzz/ssh2_client_fuzzer.cc
+++ b/tests/ossfuzz/ssh2_client_fuzzer.cc
@@ -90,10 +90,10 @@ EXIT_LABEL:
 
     libssh2_exit();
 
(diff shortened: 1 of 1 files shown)


tests/ossfuzz/ssh2_client_fuzzer.cc
```diff
@@ -90,8 +90,13 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 
     libssh2_exit();
 
-    close(socket_fds[0]);
-    close(socket_fds[1]);
+    if (socket_fds[0] != -1) {
+        close(socket_fds[0]);
+    }
+    if (socket_fds[1] != -1) {
```
vszakats, line 96:
```suggestion
    if(socket_fds[0] != -1) {
        close(socket_fds[0]);
    }
    if(socket_fds[1] != -1) {
```
formatting nit

vszakats (changes requested):

tests/ossfuzz/ssh2_client_fuzzer.cc
```diff
@@ -90,8 +90,13 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
...
+        close(socket_fds[1]);
+    }
+
```
vszakats, line 99:
```suggestion
```
delete extra empty line

vszakats (changes requested):

tests/ossfuzz/ssh2_client_fuzzer.cc
Shubham7-1: Yes done changes

vszakats:
Can you explain how is this patch addressing the NULL-dereference in the linked OSS FUZZ issue?

Also, how can invalid socket handles get past through `FUZZ_ASSERT()`? If they can, wouldn't it be better to bail out anyway because all the code depends on valid sockets, not just `close()` at the end.

vszakats:
This fix isn't fixing a NULL dereference, contrary to the title. Instead it adds 
to invalid socket checks for socket close calls, but it's unclear if those are
useful, because the code already check for a failed socketpair open, and 
no other socket uses check for an invalid socket either.

Let me know if I missed something.

PR #1599, opened 2025-05-11
kex: fix DH-GEX-sha256 bignum initialization

jbrobst:
In `kex_method_diffie_hellman_group_exchange_sha256_key_exchange`,
`p` and `g` are later initialized with `_libssh2_bn_from_bin`, so they
should be initially created using `_libssh2_bn_init_from_bin` rather
than `_libssh2_bn_init`, as is done in
`kex_method_diffie_hellman_group_exchange_sha1_key_exchange`.

Fixing memory leaks when using the libgcrypt backend.

Follow-up to 09c5e59933daf67b833f34b8c388766abc038483
Ref: https://web.archive.org/web/trac.libssh2.org/ticket/168

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -1590,8 +1590,8 @@ kex_method_diffie_hellman_group_exchange_sha256_key_exchange(
     int rc;
 
     if(key_state->state == libssh2_NB_state_idle) {
-        key_state->p = _libssh2_bn_init();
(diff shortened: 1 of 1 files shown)


vszakats (approved):

vszakats:
Thanks @jbrobst!

PR #1603, opened 2025-05-14
fix byte amount copied for method in libssh2_agent_sign

palmin:
It looks like the incorrect length is used to copy the public key method into the session in `libssh2_agent_sign()` and while the public key type at the start of a identity blob is often identical to the public key method it might not always be such as when method is `rsa-sha2-256` for `ssh-rsa` keys.

### DIFF
--- a/src/agent.c
+++ b/src/agent.c
@@ -1255,7 +1255,7 @@ libssh2_agent_sign(LIBSSH2_AGENT *agent,
     agent->session->userauth_pblc_method = LIBSSH2_ALLOC(agent->session,
                                                          method_len);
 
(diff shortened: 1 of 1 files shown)


MichaelBuckley:
This change is correct. However, I think it could be improved by renaming `methodLen` to something more descriptive. Having `methodLen` and `method_len` is liable to cause more confusion in the future. Could I ask you to also rename `methodLen` to `identity_method_len` (or something else, if you have a better idea) as a part of this change? Thanks in advance.

MichaelBuckley:
This looks great. Thank you very much!

vszakats:
Thank you @palmin, merged!

PR #1607, opened 2025-05-22
Buffer overread vulnerability fix

vivek4434:
If Address Sanitization (asan) is enabled when building libssh2 and any encapsulating code, attempts to call the vulnerable functions using mismatched lengths will cause the application to panic and crash. If asan is not enabled during the build process, the application will continue running with the malicious buffer. The affected functions send the data to the connected SSH server, leaking memory from the server running libssh2. Due to the nature of libssh2, and code using the library, there is a high probability of sensitive information such as authentication material being leaked to the remote server.

### DIFF
--- a/src/channel.c
+++ b/src/channel.c
@@ -2337,6 +2337,13 @@ ssize_t
 _libssh2_channel_write(LIBSSH2_CHANNEL *channel, int stream_id,
                        const unsigned char *buf, size_t buflen)
 {
+    if (buflen > strlen((const char *)buf)) {
+        /* Length mismatch */
+        return _libssh2_error(channel->session,
+                                  LIBSSH2_ERROR_INVAL,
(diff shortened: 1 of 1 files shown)


willco007:
buf may not be null terminated so you can't call strlen() on it, that is why buflen is passed in.

vivek4434:
> buf may not be null terminated so you can't call strlen() on it, that is why buflen is passed in.

if buflen > input_buf then it is vulnerable, so there should be a check for that

vivek4434:
> The check needs to be done before calling `_libssh2_channel_write()`, not in this function. This function requires the buffer to be the length it says it is. If you have a solution that is workable to check in this function I'd be interested in an MR for it, but it can't be done like this.

understood. i'll get back.

vszakats:
@vivek4434 Are you still working on this PR?

PR #1611, opened 2025-06-13
Update FindMbedTLS.cmake

winterheart:
This module is a compatibility layer for the library version below 3.6.0, where full support of cmake configuration modules appeared. It is recommended to use the library 3.6 and higher, for which this module is redundant.

The module provides the same targets as the MbedTLS 3.6 modules, due to which the migration process to 3.6 will be very simple - just delete the local copy of this file and reinitialize the project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -478,10 +478,10 @@ endif()
 
 if(CRYPTO_BACKEND STREQUAL "mbedTLS" OR NOT CRYPTO_BACKEND)
   find_package(MbedTLS ${_specific_crypto_requirement})
-  if(MBEDTLS_FOUND)
+  if(MbedTLS_FOUND)
     set(CRYPTO_BACKEND "mbedTLS")
     set(CRYPTO_BACKEND_DEFINE "LIBSSH2_MBEDTLS")
-    list(APPEND LIBSSH2_LIBS libssh2::mbedcrypto)
+    list(APPEND LIBSSH2_LIBS MbedTLS::mbedcrypto)
   endif()
 endif()
 
--- a/cmake/FindMbedTLS.cmake
+++ b/cmake/FindMbedTLS.cmake
@@ -1,89 +1,131 @@
 # Copyright (C) The libssh2 project and its contributors.
 # SPDX-License-Identifier: BSD-3-Clause
-#
-###########################################################################
-# Find the mbedTLS library
-#
-# Input variables:
-#
-# - `MBEDTLS_INCLUDE_DIR`:  The mbedTLS include directory.
-# - `MBEDCRYPTO_LIBRARY`:   Path to `mbedcrypto` library.
-#
-# Defines:
-#
-# - `MBEDTLS_FOUND`:        System has mbedTLS.
-# - `MBEDTLS_VERSION`:      Version of mbedTLS.
-# - `libssh2::mbedcrypto`:  mbedcrypto library target.
 
-set(_mbedtls_pc_requires "mbedcrypto")
+#[=======================================================================[.rst:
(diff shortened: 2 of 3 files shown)


winterheart:
Hello.
I'm the maintainer of the mbedtls package on the Gentoo Linux system. We have ongoing migrations of all package base from mbedtls-2 to mbedtls-3 as the 2.28.x branch reached EOL and will not receive any updates (https://bugs.gentoo.org/956763). At the same time we have to keep both versions in the package tree because not all packages are ready for a smooth transition.

My part here is to prepare each dependent package to be ready for mbedtls-3 integration. While libssh2's source code supports mbedtls-3, custom FindMbedTLS.cmake does not allow easy to do it. In this PR I've updated module in a way that allows re-use of the configuration module from MbedTLS 3.6 branch which provides better ways for linking.

Main issue to integrate with existing mbedts-3 installation on Gentoo is that libraries have custom naming (like mbedtls-3.so, mbedcrypto-3.so and mbedx509-3.so). Custom FindMbedTLS.cmake not aware of these names, so fails to find the library. With my changes we can remove this file on package building, and cmake happily find configuration module that made by mbedtls itself and provide proper MbedTLS::mbedtls, MbedTLS::mbedcrypto and MbedTLS::mbedx509 targets (effectively, this is like `-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON`, but in Gentoo there we have more convenient way by just deleting specified module). Still, I don't break any other builds as they may still rely to updated FindMbedTLS.cmake file as long as it finds targets in their systems.

vszakats:
Thanks for the explanation.

The issue I see with this proposal is that it flatly overwrites this module from another project (?),
misaligning it with the other modules in this repo, its features, formatting, naming conventions.
It also overwrites code that was working just fine (version detection). It removes mbedTLS 2
support, which I think we're not ready to do yet (e.g. 3.x support is lacking in some regards),
and some libssh2 users may rely on it. It also removes `pkg-config` support. Touching the 
`MbedTLS::` namespace also feels risky.

As I understand the issue is not recognizing certain lib naming schemes, what do you think of
fixing it in the existing module, with the minimal necessary changes?

vszakats:
I'm not sure I agree to alter libssh2 to prepare it for a build case where some of its
parts are modified in incompatible ways, or its Find module deleted. Nor that it
should export a 3rd-party namespace. If it's patched, isn't it simpler to patch it in
a way that remains compatible with the rest of libssh2?

If such patch, or a subset of it, can be tamed to merge here, it can work too, e.g.?:
```diff
--- a/cmake/FindMbedTLS.cmake
+++ b/cmake/FindMbedTLS.cmake
@@ -33,7 +33,7 @@ else()
   set(_mbedtls_pc_requires "")
 
   find_path(MBEDTLS_INCLUDE_DIR NAMES "mbedtls/version.h")
-  find_library(MBEDCRYPTO_LIBRARY NAMES "mbedcrypto" "libmbedcrypto")
+  find_library(MBEDCRYPTO_LIBRARY NAMES "mbedcrypto" "libmbedcrypto" "mbedcrypto-3")
 
   unset(MBEDTLS_VERSION CACHE)
   if(MBEDTLS_INCLUDE_DIR)
```

Exporting a 3rd-party namespace would assume that libssh2's private Find module behaves
exactly the same as any other Find module (e.g. future built-in CMake one) exporting this
target, which is impossible to guarantee.

vszakats:
Closing with the thought that libssh2's Find modules should remain
a private matter to libssh2, in sync with its other internals and Find 
modules.

If needs fixing to detect dependencies in certain envs, we welcome
another PR for that.

PR #1617, opened 2025-07-08
INSTALL_CMAKE: fixes build result directories

lucasqueiroz23:
In the CMake installation instructions, the resulting static library,
examples and tests are actually found in `bld`, not in `bin`.


vszakats (approved):

vszakats:
Thank you!

PR #1632, opened 2025-07-22
security: add bounds checking in _libssh2_store_str to prevent heap overflow

LkkkLxy:


LkkkLxy:
> I'm not sure this is necessary, the validation should be done on the caller side.

Generally I agree with you,but I believe validation should be implemented for potentially dangerous data that can be easily protected. You may refer to my submitted Issue #1632 for reference.

PR #1638, opened 2025-08-07
CMake: Make installation optional; use `NEW` behavior for `CMP0077`

trustytrojan:
- Enables `NEW` behavior of CMake policy [`CMP0077`](https://cmake.org/cmake/help/latest/policy/CMP0077.html) so that existing variables set by dependent projects are **respected and not overridden** by the last argument to `option(...)`.
- Makes the installation phase **optional** by adding an `ENABLE_INSTALL` option variable.
  - This was breaking the build of one of my projects that uses MbedTLS. The error was along the lines of `install(...) required target "mbedcrypto" but it was not found in any export set`, same with `export(...)` calls. I don't need installation for that...

I hope this can be backported to the latest stable release 1.11.1 :slightly_smiling_face: 

---

w/o sp https://github.com/libssh2/libssh2/pull/1638/files?w=1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,10 +50,6 @@ include(FeatureSummary)
 include(CheckFunctionExistsMayNeedLibrary)
 include(CheckNonblockingSocketSupport)
 
-if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.13)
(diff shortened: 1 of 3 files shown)


vszakats:
Thank you, merged!

PR #1663, opened 2025-09-14
REUSE: drop unnecessary exception

vszakats:


vszakats:
Closing without merge.

PR #1669, opened 2025-09-19
mbedtls: drop support for 2.x and older

vszakats:


vszakats:
2.x still on Bookworm and on the Ubuntu runner.

To retry later.

PR #1674, opened 2025-09-24
cmake: define _DEFAULT_SOURCE for POSIX extensions

tambry:
If the compiler defaults to non-GNU mode `_DEFAULT_SOURCE` needs to be explicitly defined for Linux headers to expose definitions for various POSIX extensions.


tambry:
> only LibreSSL sets it, when `CMAKE_SYSTEM_NAME MATCHES "Linux"`

I could probably scope similarly here.
I just don't have any non-Linux systems to verify if it's necessary there too (LibreSSL's check would suggest not).
Though then again it probably just has no effect elsewhere.

vszakats:
Thanks for the details! I wonder if it'd be a better approach to update the code to not use POSIX extensions? The chacha code uses types could be replaced with standard ones.

In case of `popen()`, perhaps defining `_DEFAULT_SOURCE` in-source would be better, with a comment saying why it's necessary. Same for any specific source in `examples`, that may need it.

tambry:
> I wonder if it'd be a better approach to update the code to not use POSIX extensions?
> In case of popen(), perhaps defining _DEFAULT_SOURCE in-source would be better, with a comment saying why it's necessary. Same for any specific source in examples, that may need it.

Absolutely, done!
I was trying to keep the changes minimal, but then again actually cleaning up the code to avoid needing the extensions is fairly simple too.

tambry:
I'll open a separate PR to try bump maximum CMake range and add to a `CMAKE_C_EXTENSIONS=OFF` combination to CI.

example/x11.c
```diff
@@ -7,6 +7,11 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+/* For fileno(), tcsetattr() */
+#if defined(__linux__) && !defined(_DEFAULT_SOURCE)
+#define _DEFAULT_SOURCE
+#endif
```
vszakats, line 13:
```suggestion
#if defined(__linux__) && !defined(_DEFAULT_SOURCE)
#define _DEFAULT_SOURCE  /* for fileno(), tcsetattr() */
#endif
```
for greppability

PR #1675, opened 2025-09-24
Add SFTP input bounds checking

willco007:
Notes:
When being passed in invalid path lengths to various public SFTP API, `uint32_t` packet length can overflow resulting in heap corruption.  This MR will compute packet length as `size_t` and bounds check it before using it.

Credit:
[Oblivionsage](https://github.com/Oblivionsage)

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -47,6 +47,7 @@
 #include "sftp.h"
 
 #include <assert.h>
+#include <stdint.h>
 #include <stdlib.h>  /* strtol() */
 
 /* This release of libssh2 implements Version 5 with automatic downgrade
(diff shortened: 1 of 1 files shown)


src/sftp.c
```diff
@@ -1155,19 +1155,30 @@ sftp_open(LIBSSH2_SFTP *sftp, const char *filename,
...
     ssize_t rc;
+    size_t packet_len;
     int open_file = (open_type == LIBSSH2_SFTP_OPENFILE) ? 1 : 0;
 
+    /* packet_len(4) + packet_type(1) + request_id(4) + filename_len(4) +
+       flags(4) */
+    packet_len = (filename_len + 13 +
+                 (open_file ? (4 + sftp_attrsize(attrs.flags)) : 0));
+
+    if(packet_len > UINT32_MAX || packet_len == 0) {
```
Michael-Panic, line 1166: I don't think packet_len can ever be > UINT32_MAX unless filename_len is nearly that large. However, I'm not aware of any filesystem that allows filenames that long.

Nonetheless, if filename_len is that large, this won't catch overflows. Say for example that filename_len is UINT32_MAX - 1 and that open_file is false. We'll end up with (UINT32_MAX - 1) + 13, which on a 32-bit system would wrap around to 11.

So the proper check would be UINT32_MAX - (13 + (open_file ? (4 + sftp_attrsize(attrs.flags)) : 0)) > filename_len

Michael-Panic (commented):
I left a very pedantic comment that I don't expect would ever be an issue, at least not in any clients that validate their filenames. But technically, it could lead to an overflow.

PR #1700, opened 2025-10-06
publickey: guard attribute allocation overflow in list_fetch v2

MegaManSec:


src/publickey.c
```diff
@@ -1109,9 +1109,16 @@ libssh2_publickey_list_fetch(LIBSSH2_PUBLICKEY * pkey, unsigned long *num_keys,
                 }
 
                 if(list[keys].num_attrs) {
+                    size_t count = (size_t)list[keys].num_attrs;
+                    if(count > ((size_t)-1) /
```
Michael-Panic: We should use `SIZE_MAX` here instead. libssh2_priv.h, which is included by this file, makes sure it's defined on platforms that don't define it themselves.

(The definition itself isn't very exhaustive. It's 0xFFFFFFFFFFFFFFFF for Win64 and 0xFFFFFFFF for any other system that doesn't define it, but it's part of stdint.h. But I'm not aware of any platforms we support that both have a 64-bit size_t and don't have stdint.h.)

Michael-Panic (changes requested):

PR #1702, opened 2025-10-06
packet: bound-safe logging for SSH_MSG_DEBUG/DISCONNECT

MegaManSec:
Use %.*s with parsed lengths and NULL guards when printing message/language in _libssh2_packet_add(). These fields are not NUL-terminated and may be absent; previous %s usage risked OOB reads or NULL deref (debug builds).


Michael-Panic:
This looks fine to me, but the CI failed some tests. It looks like it might be a temporary infrastructure issue, but I don't have access to run these tests myself.

PR #1708, opened 2025-10-06
sftp: readdir() fix off-by-one when NUL-terminating filename/longentry

MegaManSec:


src/sftp.c
```diff
@@ -1867,13 +1867,13 @@ static ssize_t sftp_readdir(LIBSSH2_SFTP_HANDLE *handle, char *buffer,
...
-            if(filename_len >= buffer_maxlen) {
+            if(filename_len + 1 > buffer_maxlen) {
                 filename_len = (size_t)LIBSSH2_ERROR_BUFFER_TOO_SMALL;
                 goto end;
             }
 
-            if(buffer_maxlen >= filename_len && names_packet_len >=
-               filename_len) {
+            if(buffer_maxlen >= (filename_len + 1) &&
+               names_packet_len >= real_filename_len) {
```
Michael-Panic: This one's a real potential problem, but again, the +1 isn't what we want to be doing here because of the potential of an overflow. Instead, changing this to `buffer_maxlen > filename_len` should be sufficient.

PR #1711, opened 2025-10-06
agent: pageant backend, bound reply copy, handle missing reply

MegaManSec:
The Pageant transact path trusted the 32-bit length in the shared memory
mapping and could memcpy past the mapped view. It also treated
a non-positive SendMessage(WM_COPYDATA) result as success.

Changes:

Reject replies when SendMessage returns ≤ 0 and report
LIBSSH2_ERROR_AGENT_PROTOCOL.

Bound the copy by validating response_len <= PAGEANT_MAX_MSGLEN - 4
(accounting for the length prefix) to avoid OOB reads.

Impact: prevents potential out-of-bounds read and use of uninitialized
mapping contents when Pageant misbehaves or is malicious.

### DIFF
--- a/src/agent.c
+++ b/src/agent.c
@@ -721,7 +721,7 @@ agent_transact_pageant(LIBSSH2_AGENT *agent, agent_transaction_ctx_t transctx)
         UnmapViewOfFile(p);
         CloseHandle(filemap);
         return _libssh2_error(agent->session, LIBSSH2_ERROR_AGENT_PROTOCOL,
-                              "pagent did not handle message");
+                              "pageant did not handle message");
     }
 
     UnmapViewOfFile(p);


src/agent.c
```diff
@@ -716,6 +716,11 @@ agent_transact_pageant(LIBSSH2_AGENT *agent, agent_transaction_ctx_t transctx)
                                   "agent malloc");
         }
         memcpy(transctx->response, p + 4, transctx->response_len);
+    } else {
```
vszakats, line 719:
```suggestion
    }
    else {
```
checksrc

src/agent.c
```diff
@@ -717,6 +717,12 @@ agent_transact_pageant(LIBSSH2_AGENT *agent, agent_transaction_ctx_t transctx)
         }
         memcpy(transctx->response, p + 4, transctx->response_len);
     }
+    else {
+        UnmapViewOfFile(p);
+        CloseHandle(filemap);
+        return _libssh2_error(agent->session, LIBSSH2_ERROR_AGENT_PROTOCOL,
+                              "no response from pageant");
```
Michael-Panic: More accurately, I might go with "pagent didn't handle message," but that's super pedantic, so I'm going to approve this one as-is.

vszakats:
```suggestion
                              "pagent did not handle message");
```
We can go with that, also to have another go with the CI.
For pedandic++, I'd suggest dropping the contraction:
"pagent did not handle message"

src/agent.c
```diff
@@ -717,6 +717,12 @@ agent_transact_pageant(LIBSSH2_AGENT *agent, agent_transaction_ctx_t transctx)
         }
         memcpy(transctx->response, p + 4, transctx->response_len);
     }
+    else {
+        UnmapViewOfFile(p);
+        CloseHandle(filemap);
+        return _libssh2_error(agent->session, LIBSSH2_ERROR_AGENT_PROTOCOL,
+                              "pagent did not handle message");
```
vszakats:
```suggestion
                              "pageant did not handle message");
```
spellcheck

PR #1727, opened 2025-10-16
mbedtls: require 3.1.0 (2.x no longer supported upstream)

vszakats:
The mbedTLS API is not one API but distinctly incompatible APIs
depending on its version. 2.x is largely incompatible with 3.x, and 4.x
is largely incompatible with 3.x, also dropping certain legacy crypto
primitives.

Following the release of 3.x in 2021, libssh2 added support for 3.x in
2022. 4.x was released in October 2025, and it makes using the PSA
Crypto API mandatory, after adding it in 3.x.

Supporting 2.x while also supporting the PSA Crypto API would mean
supporting 3 different crypto backends meshed together in a single
source file, with various overlapping and per-version parts, with
different features and separate needs to test.

2.x is also no longer maintained, supported, or security patched, as
written in its latest, and final, release notes on 2025-03-24:
https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-2.28.10
The mbedTLS project recommends upgrading to an upper major version.

To allow adding support for the PSA Crypto API and the 4.x series, and
to avoid an unmaintained 2.x major version, this patch drops support for
mbedTLS 2.x.

Also:
- GHA: update CI jobs to either use Linuxbrew mbedtls 3.x, or build from
  source and cache it.
- Require 3.1.0 minimum (instead of 3.0.0), because libssh2 never built
  with 3.0.0 due the use of API `mbedtls_cipher_info_get_key_bitlen()`.
  Nobody complained since the initial commit introducing it.

Refs:
https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/tf-psa-crypto-1.0.0/docs/1.0-migration-guide.md
https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/tf-psa-crypto-1.0.0/docs/psa-transition.md#asymmetric-cryptography

---

w/o sp: https://github.com/libssh2/libssh2/pull/1727/files?w=1

mbedTLS 4.0.0 was released yesterday with a new round of breaking changes.
Drop support for 2.x to give room for the upcoming complexity.

- [x] fix CI. GHA's ubuntu-latest (24.04) is shipping with 2.x. Also Debian Bookworm.
  https://sources.debian.org/src/mbedtls/

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -85,7 +85,7 @@ jobs:
             cmake --build . --verbose
           else
             eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
-            export PKG_CONFIG_PATH; PKG_CONFIG_PATH="$(brew --prefix mbedtls)/lib/pkgconfig"
+            export PKG_CONFIG_PATH; PKG_CONFIG_PATH="$(brew --prefix mbedtls@3)/lib/pkgconfig"
             for crypto in OpenSSL Libgcrypt mbedTLS wolfSSL; do
               cmake -B _bld-"${crypto}" -G Ninja -DCRYPTO_BACKEND="${crypto}" -DENABLE_ZLIB_COMPRESSION=ON
               cmake --build _bld-"${crypto}" --verbose


vszakats:
mbedTLS breaks compatibility in big ways on each major release,
and it did it again with 4.x. If you have a patch adding support
for 4.x while keeping support for 2.x, we're happy to review the PR.

Issue #1691, opened 2025-09-30
Memory Leak in libssh2_session_free


LkkkLxy:
## Describe the bug

libssh2 contains a memory leak in `session_free()` where approximately 82KB of memory is not freed when sessions fail to complete handshake. The root cause is state-dependent cleanup logic that skips memory deallocation when `LIBSSH2_STATE_NEWKEYS` flag is not set.

## Build Information

**libssh2 version**: Current git master (commit from fuzzing environment)

**System Information**:
- Building on and for Linux x86_64
- Tested in Docker container with AddressSanitizer/LeakSanitizer

**Toolchain**:
- Compiler: clang with AddressSanitizer (`-fsanitize=address -fsanitize=leak`)

## To Reproduce

The leak occurs when:
1. A session is created with `libssh2_session_init()`
2. A handshake is attempted but fails or is incomplete (common with invalid sockets)
3. `libssh2_channel_open_session()` is called (fails but allocates memory)
4. Proper cleanup is performed with `libssh2_session_free()` and `libssh2_exit()`

### Minimal Reproduction Code

```c
#include <libssh2.h>
#include <sys/socket.h>
#include <unistd.h>

int main() {
    libssh2_init(0);

    LIBSSH2_SESSION* session = libssh2_session_init();
    libssh2_session_set_blocking(session, 0);

    int sv[2];
    if (socketpair(AF_UNIX, SOCK_STREAM, 0, sv) == 0) {
        libssh2_session_handshake(session, sv[0]);  // Fails, state stays INITIAL_KEX
    }

    LIBSSH2_CHANNEL* chan = libssh2_channel_open_session(session);
    if (chan) {
        libssh2_channel_close(chan);
        libssh2_channel_wait_closed(chan);
        libssh2_channel_free(chan);
    }

    libssh2_session_disconnect(session, "test");
    libssh2_session_free(session);  // Memory leak occurs here

    close(sv[0]); close(sv[1]);
    libssh2_exit();
    return 0;
}
```

### Root Cause 
Analysis Through detailed source code analysis, the memory leak stems from a **state-driven cleanup logic design flaw** in session_free() (session.c:966):
```
if(session->state & LIBSSH2_STATE_NEWKEYS) {
    // Most memory cleanup logic only executes when this condition is true
    // Including hostkey, crypto components, session_id cleanup
}
```

## Additional Context

This bug was discovered through fuzzing.

[...]

Issue #623, opened 2021-09-02
We have a memory leak at   src/transport.c:474


timelyslot:
we find libssh2-1.9.0/src/transport.c:474(p->payload = LIBSSH2_ALLOC(session, total_num);), always LIBSSH2_ALLOC without LIBSSH2_FREE it, and memory leak. 
    Our program is used for SFTP transmission with 20 threads. Due to the demand, we will not close the session after uploading the file and maintain a long-time connection with the server by keeping it alive. After running for a long time, it is found that libssh2_ sftp_ open_ Ex has a memory leak, and finally locate p - > payload = libsh2_ ALLOC(session, total_ num); We want to know whether there is a wrong method or something wrong

 - OS: linux
 - libssh2 version 1.10.0

willco007:
There are a couple places where `p->payload` is released, the main one is in the function `fullpacket`. Can you step through your code and make sure it's being hit as expected.

timelyslot:
Through the test, it is found that when the keepalive  is configured with want_reply = 1, the memory leak will occur at the next time which _libssh2_transport_read(libssh2_sftp_open_ex) called. I guess the message from the server is not processed, while allocaing space  mistakenly during the next reading. Can you tell us how to deal with this situation

emiliopastormira:
Hi,
I could solve the leak as we see it with the following patch
--- a/src/session.c 2021-10-11 15:29:54.768257700 +0100
+++ b/src/session.c 2021-10-13 06:51:50.788974100 +0100
@@ -1053,7 +1053,7 @@
}



/* Free payload buffer */
- if(session->packet.total_num) {
+ if(session->packet.payload) {
LIBSSH2_FREE(session, session->packet.payload);
}

As I see it in closing a session, payload would need definitely to be freed, and a pointer should be NOT null, only if in use.

However, I would think that the line may need a change from LIBSSH2_ALLOC to LIBSSH2_REALLOC if that part of the code is going to be used multiple times (as my analysis indicated at first glance. 
I could not fully test my second suggestion, sorry.

cloud4t0r:
I encounter the same memory leak in mly company with version 1.11.0 ...

willco007:
May be fixed with https://github.com/libssh2/libssh2/pull/1412, please test it and let us know.

LarsNordin-LNdata:
PR #1804 fails with double free, the fix is not that easy and I need to dig into how it is supposed to work.

obhor:
[...]

PR #1745, opened 2025-11-24
Add check for valid socket in session_startup()

LarsNordin-LNdata:
Insert check for valid socket

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -43,6 +43,8 @@
 
 #ifdef _WIN32
 #include <ws2tcpip.h>  /* for socklen_t */
+#else
+#include <netinet/in.h>
 #endif
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
@@ -739,6 +741,8 @@ static int
 session_startup(LIBSSH2_SESSION *session, libssh2_socket_t sock)
 {
     int rc;
+    struct sockaddr_in serv_addr;
+    socklen_t len = sizeof(serv_addr);
 
     if(!session) {
         fprintf(stderr, "Session is NULL, error: %i\n",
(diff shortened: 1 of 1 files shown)


LarsNordin-LNdata:
This fix was a reaction to issue #1691. The issue is about a possible memory leak but the code is using a local socket and also non-blocking. The non-existing leak is because the session_free() is returning EAGAIN but is not looping until session_free() is finished) and this because it not a real server on the other side. If the usage of libssh2 would be blocking, it would block foreever.

I OK with that you do not take this PR, but please close #1691 with a comment about bad usage and non-blocking should always loop until not EAGAIN to avoid memory leak (I'm about to updating allnon-blocking examples so no memory leak is in the examples.

The examples should be like this: (I wanted to check my fix for #623 and tested with example/sftp_write_sliding and got a memory leak, but it was from not handling EAGAIN in closa and free).
 ```
while(libssh2_sftp_close(sftp_handle) == LIBSSH2_ERROR_EAGAIN)
        waitsocket(sock, session);
    while(libssh2_sftp_shutdown(sftp_session) == LIBSSH2_ERROR_EAGAIN)
        waitsocket(sock, session);

shutdown:

    fclose(local);

    if(session) {
        while(libssh2_session_disconnect(session, "Normal Shutdown") ==
              LIBSSH2_ERROR_EAGAIN)
                waitsocket(sock, session);
        while (libssh2_session_free(session) == LIBSSH2_ERROR_EAGAIN)
            waitsocket(sock, session);
    }
```

vszakats:
I am not a fan of doing this within libssh2, for portability and
robustness reasons. Deciding what is a valid socket seems
non-trivial. The libssh2 API user is resposible for providing
a working socket.

vszakats:
Closing as something not planned.

PR #1758, opened 2025-12-05
Fix mbedTLS AES-CBC backend: avoid BAD_INPUT_DATA during cipher_finish() in SSH handshake

### LINKED ISSUE #793: Hitting error -44 "Unable to ask for ssh-userauth service" with mbedtls

AdrianMoranMontes:
# Description

This PR fixes a reproducible failure in the mbedTLS backend of libssh2 when negotiating AES-CBC ciphers with certain SSH servers.

When libssh2 is built with mbedTLS (tested with mbedTLS 3.6.2), the library fails to encrypt the very first SSH packet after NEWKEYS when the chosen cipher is AES-XXX-CBC. The failure manifests as:

```shell
Failure establishing SSH session: -44
libssh2 last error: Unable to ask for ssh-userauth service
```

Tracing the mbedTLS backend reveals the root cause:

```shell
mbedTLS cipher_finish failed: ret = -24832 (MBEDTLS_ERR_CIPHER_BAD_INPUT_DATA)
mbedTLS cipher_crypt failed
crypt name = aes256-cbc
```

This happens during the encryption of SSH_MSG_SERVICE_REQUEST (seqno=0), right after the KEX exchange succeeds.
The server → client traffic is irrelevant at that point; this is a pure client-side mbedTLS cipher failure.

# Reproduction steps

The issue can be reproduced with:

   - libssh2 master
   - mbedTLS 3.6.2
   - any SSH server that negotiates AES-128-CBC or AES-256-CBC
(confirmed with OpenSSH and a proprietary embedded SSH server)

Steps:

   1. Build libssh2 with mbedTLS backend.
   2. Connect to a server offering CBC ciphers.
   3. Let the negotiation select AES-CBC.
   4. Observe handshake failure right after NEWKEYS.

# Root cause analysis

1. Wrong interpretation of the `encrypt` flag

The code used:

```c
op = encrypt == 0 ? MBEDTLS_ENCRYPT : MBEDTLS_DECRYPT;
```

which inverted the intended direction.
The fix correctly maps the flag:

```c
op = encrypt ? MBEDTLS_ENCRYPT : MBEDTLS_DECRYPT;
```

2. Missing explicit padding configuration for mbedTLS

libssh2 handles SSH packet padding internally, so the cipher must operate with no padding.
Without explicitly setting this, mbedTLS rejects CBC operations during `cipher_finish()`.

The fix adds:

```c
mbedtls_cipher_set_padding_mode(ctx, MBEDTLS_PADDING_NONE);
```

This ensures that mbedTLS does not attempt to validate or add padding at the cipher layer.

# Related issues

Probably, this PR fixes #793

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -135,10 +135,10 @@ _libssh2_mbedtls_cipher_init(_libssh2_cipher_ctx *ctx,
     if(!ret) {
         mbedtls_cipher_type_t ctype =
             mbedtls_cipher_info_get_type(cipher_info);
-        if( ctype == MBEDTLS_CIPHER_AES_128_CBC ||
-            ctype == MBEDTLS_CIPHER_AES_192_CBC ||
-            ctype == MBEDTLS_CIPHER_AES_256_CBC ||
-            ctype == MBEDTLS_CIPHER_DES_EDE3_CBC) {
+        if(ctype == MBEDTLS_CIPHER_AES_128_CBC ||
+           ctype == MBEDTLS_CIPHER_AES_192_CBC ||
+           ctype == MBEDTLS_CIPHER_AES_256_CBC ||
+           ctype == MBEDTLS_CIPHER_DES_EDE3_CBC) {
             ret = mbedtls_cipher_set_padding_mode(ctx, MBEDTLS_PADDING_NONE);
         }
     }
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -133,12 +133,10 @@ _libssh2_mbedtls_cipher_init(_libssh2_cipher_ctx *ctx,
      * set_padding_mode for CBC ciphers since other modes (CTR, stream)
      * are not applicable and will cause an error. */
     if(!ret) {
-        mbedtls_cipher_type_t ctype =
-            mbedtls_cipher_info_get_type(cipher_info);
-        if(ctype == MBEDTLS_CIPHER_AES_128_CBC ||
-           ctype == MBEDTLS_CIPHER_AES_192_CBC ||
-           ctype == MBEDTLS_CIPHER_AES_256_CBC ||
-           ctype == MBEDTLS_CIPHER_DES_EDE3_CBC) {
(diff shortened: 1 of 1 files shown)


vszakats (dismissed):

src/mbedtls.c
```diff
@@ -119,14 +119,28 @@ _libssh2_mbedtls_cipher_init(_libssh2_cipher_ctx *ctx,
     if(!ctx)
         return -1;
 
-    op = encrypt == 0 ? MBEDTLS_ENCRYPT : MBEDTLS_DECRYPT;
+    op = encrypt ? MBEDTLS_ENCRYPT : MBEDTLS_DECRYPT;
```
vszakats: Makes me wonder how could libssh2 + mbedTLS work correctly before if this condition
was reversed all along?

PR #1762, opened 2025-12-12
example/x11.c: free session 

LarsNordin-LNdata:
To be able to verify memory-leak the session has to be free'd

### DIFF
--- a/example/x11.c
+++ b/example/x11.c
@@ -270,7 +270,7 @@ int main(int argc, char *argv[])
     libssh2_socket_t sock = LIBSSH2_INVALID_SOCKET;
     struct sockaddr_in sin;
     LIBSSH2_SESSION *session = NULL;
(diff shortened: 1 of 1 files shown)


example/x11.c
```diff
@@ -506,6 +506,19 @@ int main(int argc, char *argv[])
         libssh2_channel_free(channel);
         channel = NULL;
     }
+
+    if(session) {
+        libssh2_session_disconnect(session, "Normal Shutdown");
+        libssh2_session_free(session);
+    }
```
vszakats:
```suggestion
    if(session)
        session_shutdown(session);
```
Use the local function doing the above two calls?

vszakats:
What do you think of adding a `shutdown:` label to the cleanup code, and
goto there from error branches, to do a full cleanup in those cases too?

This may also make the internal `session_shutdown()` function unnecessary,
but doing those two calls.

PR #1765, opened 2025-12-25
cmake: rename `libssh2-config.cmake.in` to `libssh2-config.in.cmake`

vszakats:
To:
- match filename suffix with file content.
- simplify recognizing CMake sources.
- ensure syntax highlighters use the correct file type.

Also:
- cmakelint.sh: simplify, alpha sort the filelist.


vszakats:
Unfortunately the AppVeyor CI + ssh Docker setup decayed further, so that the last remaining job is broken too.

PR #1775, opened 2025-12-27
Create checksrc.ps1

LarsNordin-LNdata:
Powershell-script to run checksrc on Windows.

Rus as:
powershell -File .\checksrc.ps1

or just rigth-click and "run with powershell"
On error exit-code is not 0

### DIFF
--- a/ci/checksrc.ps1
+++ b/ci/checksrc.ps1
@@ -1,0 +1,41 @@
+# Copyright (C) The libssh2 project and its contributors.
+# SPDX-License-Identifier: BSD-3-Clause
+
+if(-Not(Get-Command -ErrorAction Ignore perl.exe)) {
+    Write-Host "perl required, abort"
+    pause
+    exit
+}
+
+$currentDir = Split-Path -Path $MyInvocation.MyCommand.Definition -Parent
+cd $currentDir/..
+
+$RC = 0
+git ls-files '*.[ch]' '*.cc' | ForEach-object {
+perl ci/checksrc.pl -i4 -m79 -AFIXME -AERRNOVAR -AFOPENMODE -ATYPEDEFSTRUCT `
+  -aaccept `
+  -aatoi `
+  -acalloc `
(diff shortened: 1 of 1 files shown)


vszakats:
This would mean maintining double the copies of these rules after
this PR, reverting an earlier effort of deduplicating them. (And another
effort to reduce the amount of PowerShell in the repo.)

Would it be possible instead to pick a global configuration automatically
or via an option, and perhaps even the list of files, to avoid an extra
starter script (or any starter script)?

vszakats:
I'd prefer to not add more PowerShell code, unless absolutely unavoidable.
It's painful to maintain from other platforms. It also duplicates efforts. In this
case the checksrc script is already in Perl that's required for this on all
platforms, moving missing logic into it, also allowing to drop checksrc.sh,
would be the best IMO.

LarsNordin-LNdata:
But checksrc.pl looks lika a generic script, all libssh2 unique options are added in the shell-script when calling the perl-script.

the perl-script can be rewritten to hardcoded do what the shell-script+perl-script is doing, but then loosing the generic ascpect.

I didn't know how you maintainers feel about powershell, just skip it then...

vszakats:
I think `checksrc.pl` should remain generic, while allowing to configure
it for both libssh2 and curl to not need starter scripts and continue
to work as they did before.

In theory these may work:
- if no input file specified, it may use `git ls-files` output
- it may automatically pick up a root `.checksrc` for defaults
- it may automatically pick up per-directory `.checksrc` configs
for files in those directories.
- could read indentation and line-width config from `.checksrc`
too.

At the end a bare `checksrc.pl` invocation could check files in
current dir and below. A `checksrc.pl` with individual files would
continue checking those specifically.

Till someone cuts into the above, I made #1778 which replaces
`checksrc.sh` with a `checksrc-all.pl`. It removes the shell
dependency and also syncs with curl which did this earlier.

PR #1780, opened 2026-01-12
kex: missing `free()` when random data cannot be generated

LarsNordin-LNdata:
If random can't be generated the allocated data must be FREE


MichaelBuckley:
Looks good to me.

PR #1784, opened 2026-01-13
scp.c: handle scpSend_respons as buffer with length and not string in debug message

LarsNordin-LNdata:
A buffer with a length should never be handled as a string (even if it's the case this time)

### DIFF
--- a/src/scp.c
+++ b/src/scp.c
@@ -1,4 +1,4 @@
-scpSend/* Copyright (C) Daniel Stenberg
+/* Copyright (C) Daniel Stenberg
  * Copyright (C) Sara Golemon <sarag@libssh2.org>
  * All rights reserved.
  *


src/scp.c
```diff
@@ -994,8 +994,9 @@ scp_send(LIBSSH2_SESSION * session, const char *path, int mode,
                 snprintf((char *) session->scpSend_response,
                          LIBSSH2_SCP_RESPONSE_BUFLEN, "T%ld 0 %ld 0\n",
                          (long)mtime, (long)atime);
-            _libssh2_debug((session, LIBSSH2_TRACE_SCP, "Sent %s",
-                           session->scpSend_response));
+            _libssh2_debug((session, LIBSSH2_TRACE_SCP, "Sent %.*s",
+                (int)session->scpSend_response_len,
+                session->scpSend_response));
```
vszakats: Can you say why you think this is necessary?

My understanding is that `snprintf()` is always nul-terminating the result it writes into
`session->scpSend_response`, making existing usage safe.

There is one potential caveat with older MSVC's `_snprintf()` which does not always
nul terminate. libssh2 catches this case and remaps `snprintf()` to a safe local wrapper
that always nul-terminates.

LarsNordin-LNdata: OK, I think is bad usage but I will not figth over this

PR #1785, opened 2026-01-13
agent: replace `strncpy` with `memcpy`

LarsNordin-LNdata:


vszakats:
Thanks for the patch, and review!

PR #1786, opened 2026-01-13
packet: honor buffer length when passing to `_libssh2_debug()`

LarsNordin-LNdata:
A buffer with a length should never be handled as a string (even if it's the case this time because the next element in data-buffer is the length of language )


LarsNordin-LNdata:
God finding, I updated the PR

PR #1792, opened 2026-01-22
Add QNX support

nnarayanamurthy:


example/scp.c
```diff
@@ -29,8 +29,13 @@
 
 static const char *pubkey = "/home/username/.ssh/id_rsa.pub";
 static const char *privkey = "/home/username/.ssh/id_rsa";
+#ifdef __QNX__
+static const char *username = "qnxuser";
```
dfandrich: The actual username and password are intended to be examples for people to change as required. This change is not needed for the example to be correct on QNX.

vszakats:
@nnarayanamurthy I merged two fixes for missing includes #1688 and #1863.
Please retry on QNX and report if there remains any build failure.

PR #1811, opened 2026-03-01
fix: initializing keepalive_last_sent to now

### LINKED ISSUE #1810: To avoid interfering with auth, send keepalives at the end of the interval instead of the beginning
## Context

I was having a problem where ssh auth to github worked ok on wifi, but failed on mobile data, likely due to a lack of keepalives. So I patched libgit2 to enable keepalives. Doing so introduced a new problem which (both on WIFI and mobile data), caused github to reject auth. This issue is for that problem, which I believe this is related to `SSH_MSG_GLOBAL_REQUEST` being sent too aggressively.

[RFC4254](https://www.rfc-editor.org/rfc/rfc4254) suggests that these requests can be sent "at any time", but later [this draft](https://datatracker.ietf.org/doc/html/draft-ssh-global-requests-ok-00) attempted to amend it because:
>  A server that chooses to handle SSH_MSG_GLOBAL_REQUEST before it has
>  sent SSH_MSG_USERAUTH_SUCCESS MUST apply precautions which take into
>  account that the client has not yet authenticated.

One such precaution is to just reject auth if the client tries to engage in any non-auth activity, such as keepalives. Github is not the only one to do this, other services are also offended by receiving `SSH_MSG_GLOBAL_REQUEST` during auth:
- [SSHJ #810](https://github.com/hierynomus/sshj/issues/810)
- [Tabby #10684](https://github.com/Eugeny/tabby/issues/10684)

I managed to work around this problem by patching libgit2 to enable keepalives only after authentication had completed.
```diff
--- a/src/libgit2/transports/ssh_libssh2.c
+++ b/src/libgit2/transports/ssh_libssh2.c
@@ -875,6 +875,9 @@
        if (error < 0)
                goto done;

+       /* Send SSH keepalive every 5s to survive mobile NAT timeouts */
+       libssh2_keepalive_config(session, 1, 5);
+
        channel = libssh2_channel_open_session(session);
        if (!channel) {
                error = -1;

```
Ideally, libraries like libgit2 could just turn on keepalives generally without needing to worry about doing so too early.

## Scope

Currently, libssh2 sends a keepalive **immediately** (derailing auth in the cases described above).

Maybe libssh2 should avoid **ever** sending a keepalive until auth completes. That would be a larger change.  The small change that I'm proposing here is that libssh2 should send keepalives at the end of the keepalive interval instead of the beginning..

Following this change, the race condition between auth and keepalive might still exist, but auth now has a better shot at winning that race.

An easy way to achieve this would be to initialize `keepalive_last_sent` to the current time, rather than the epoch.  This way the first time that libssh2 subtracts to decide if the time has come to send a keepalive, it comes up with a very short time (indicating that it is not yet time for a keepalive) instead of 56 years (indicating that a keepalive long overdue).

## To reproduce

Connect to GitHub (or another strict SSH server) with keepalive configured before auth:
```c
session = libssh2_session_init();
libssh2_session_handshake(session, sock);
libssh2_keepalive_config(session, 1, 5);
/* Auth fails — the keepalive fires immediately during this call */
libssh2_userauth_publickey_fromfile(session, "git", pubkey, privkey, NULL);
```

I confirmed this against `github.com:22` (`SSH-2.0-babeld-...`):

| keepalive_last_sent | Result |
|---|---|
| 0 (unfixed) | `Authentication FAILED: -9 — Timed out waiting on socket` |
| `time(NULL)` (fixed) | `Authentication SUCCEEDED` |

## Expected behavior

The first keepalive should not be sent until `interval` seconds after `libssh2_keepalive_config()` is called, giving authentication time to complete.

## Version
 - OS and version: macOS 14.4.1 (Darwin 23.4.0 arm64)
 - libssh2 version: HEAD 704299e9
 - crypto backend and version: OpenSSL 3.6.1

MatrixManAtYrService:
Fixes: https://github.com/libssh2/libssh2/issues/1810

### DIFF
--- a/src/keepalive.c
+++ b/src/keepalive.c
@@ -52,6 +52,7 @@ libssh2_keepalive_config(LIBSSH2_SESSION *session,
     else
         session->keepalive_interval = interval;
     session->keepalive_want_reply = want_reply ? 1 : 0;
+    session->keepalive_last_sent = time(NULL);
 }
 
 LIBSSH2_API int


vszakats:
This looks like a workaround to me by postponing the next keepalive
in the hope the auth fits before it happens, it may fix some case, but
not all.

What do you think of this alternative?:
https://github.com/libssh2/libssh2/issues/1810#issuecomment-5012306806

it aims to hold off keepalive till the session finishes authenticating.

Issue #1497, opened 2024-11-21
Millisecond timeout precision not respected


Trzik:
Documentation of `libssh2_session_set_timeout` claims millisecond precision. This is never respected due to the fact that internally timestamps are obtained via `time` function that returns seconds. This results in 3 orders of magnitude difference.

Attached patch (for version 1.11.1) tries to address this issue. Feel free to use it or adapt it as you see fit.
[timeout_ms_precision_1.11.1.patch](https://github.com/user-attachments/files/17841566/timeout_ms_precision_1.11.1.patch)

PR #1818, opened 2026-03-12
Millisecond timeout precision fix

doorsdown:
Adapted patch provided by @Trzik in #1497


src/libssh2_priv.h
```diff
@@ -59,6 +59,7 @@
 #include <string.h>
 #include <time.h>
 #include <limits.h>
+#include <stdint.h>
```
vszakats: it's already included via libssh2.h, why include it again?

PR #1820, opened 2026-03-18
Fix returning a pointer to stack memory

eXpl0it3r:
[Upstream patch](https://github.com/SFML/SFML/pull/3675) to fix returning of a pointer to stack memory by having libssh2 make a copy of `buf`.

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -831,7 +831,10 @@ _libssh2_mbedtls_pub_priv_keyfile(LIBSSH2_SESSION *session,
     if(ret) {
         mbedtls_strerror(ret, (char *)buf, sizeof(buf));
         mbedtls_pk_free(&pkey);
-        return _libssh2_error(session, LIBSSH2_ERROR_FILE, buf);
+        return _libssh2_error_flags(session,
+                                    LIBSSH2_ERROR_FILE,
+                                    buf,
+                                    LIBSSH2_ERR_FLAG_DUP);
     }
 
     ret = _libssh2_mbedtls_pub_priv_key(session, method, method_len,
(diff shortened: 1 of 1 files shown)


vszakats:
~~FTR clang-tidy is reporting what looks like a similar issue in `agent_list_identities()`~~:
```
src/agent.c:956:9: warning: Address of stack memory associated with local variable 'c' is still referred to by the caller variable 'agent' upon returning to the caller.  This will
  956 |         return _libssh2_error(agent->session, LIBSSH2_ERROR_BAD_USE,
      |         ^
```

edit: nevermind, it's a different kind.

vszakats:
+1 to @willco007 suggestion to use `libssh2_session_set_last_error()`. IMO it makes
the code more readable and avoids repeating this pattern in favor of this public
function already covering it.

@eXpl0it3r What do you think of applying this change?

eXpl0it3r:
For me the responses just seem like a lot of guess-work, like "oh we found some public function in our API that we don't even know existed, let's just use that" and much less of an authoritative answer along the lines of "while these public functions were designed for external consumers, it's okay to use them internally as well, since we ensure to not create cyclic dependencies via XYZ".

Also kind of missing the proper acknowledgement of it being a bug. 😄 

Anyways, I can apply the fix. Should the documentation of `libssh2_session_set_last_error` be updated to indicate that this isn't just for external bindings?

vszakats:
> For me the responses just seem like a lot of guess-work, like "oh we found some public function in our API that we don't even know existed, let's just use that" and much less of an authoritative answer along the lines of "while these public functions were designed for external consumers, it's okay to use them internally as well, since we ensure to not create cyclic dependencies via XYZ".
> 
> Also kind of missing the proper acknowledgement of it being a bug. 😄
> 
> Anyways, I can apply the fix. Should the documentation of `libssh2_session_set_last_error` be updated to indicate that this isn't just for external bindings?

It's clearly a bug. We're conversing on the best way to fix it.

I don't think we need to explicitly say that a public API function
may be used internally.

willco007:
I don't use the mbedTLS backend nor the extended error API. I'm just one of the very few (two?) people looking at MRs here. So if you think I'm figuring it out as I go well then you're right. I'm not sure how that makes my advice more or less right.

vszakats:
@eXpl0it3r Cyclic dependency is not an issue when using the public API. It's fine to use.

When you have the opportunity to apply this update, I merge.

vszakats:
Merged as-is, thanks!

eXpl0it3r:
Thank you! Got busy with IRL stuff.

PR #1839, opened 2026-04-08
cmake: drop support for CMake 3.17 and older

vszakats:
Require CMake 3.18 (2020-07-15) or newer, up from 3.7 (2016-11-11) prior
to this patch.

This requirement also applies to the distributed `libssh2-config.cmake`.

To allow dropping compatibility code maintained for old versions, and to
use features which were unpractical in separate code paths. Also to make
testing, documentation and development easier, and builds faster due
CMake performance improvements over time.

CMake offers pre-built binaries for major platforms. They work without
an install step, just by unpacking and pointing the cmake command to
them. Making upgrades easy in many cases: 
https://cmake.org/download/
https://cmake.org/files/
https://github.com/Kitware/CMake/releases

CMake 3.18 brings these feature as generally available when building or
consuming curl/libcurl:

LTO support, improved performance, `pkg-config` and interface target
support, `OBJECT` target (for faster libcurl builds), modern invocation
with `-S`/`-B` options, better support for custom linker options,
FetchContent, `--verbose` and `--install` options, `CMAKE_GENERATOR`
env, last but not least unity mode and Ninja generator.
                                                      
Ref: https://github.com/curl/curl/discussions/18704   

---

https://github.com/libssh2/libssh2/pull/1839/files?w=1

- [x] rebase on #1841


vszakats:
unrelated flaky fail:
```
The following tests FAILED:
	 42 - test_read-hmac-sha2-256-etm@openssh.com (Failed)
Error: Process completed with exit code 8.
```
https://github.com/libssh2/libssh2/actions/runs/24162184088/job/70515870880?pr=1839

vszakats:
Another flaky:
```
The following tests FAILED:
	  6 - test_auth_password_fail_username (Failed)
```

PR #1842, opened 2026-04-09
tidy-up: miscellaneous

vszakats:
- configure: fix indent, whitespace, drop empty comments.
- .gitignore: drop `.DS_Store` (again)
  Keep this list for files created by libssh2. For other files, the
  per-machine global Git ignore file is a better place, and it fixes
  this for all repos, not just libssh2.

---

https://github.com/libssh2/libssh2/pull/1842/files?w=1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+# for files created by the libssh2 build process
 .deps
 .libs
 *-*-*
@@ -49,4 +50,3 @@ tags
 TAGS
 tap-driver.sh
 test-driver
-.DS_Store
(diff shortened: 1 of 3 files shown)


willco007:
Please don't remove.ds_store from git ignore.

vszakats:
> Please don't remove.ds_store from git ignore.

Can you add it to your local `ignore` file? Why add to libssh2?

willco007:
It's fine, I'll do that. I'd prefer if libssh2 had 'logical defaults' out of the box however.

vszakats:
That may be a PR for Git itself, fixing it for all repos. (Not even dreaming Apple fixing this in addition.)

Thanks for fixing locally.

PR #1856, opened 2026-04-11
ci/appveyor: try running tests

vszakats:


vszakats:
Tried switching back to RSA 2048, switching back to MS-DOS batch,
it worked in one run, keeps failing in other runs.

vszakats:
Continued in #1857

PR #1857, opened 2026-04-12
ci/appveyor: try fixing windows tests [WITHOUT SUCCESS]

vszakats:


vszakats:
There was one more successful SSH connection from AppVeyor CI.
It seems to be unrelated to these changes, and rather related to timing
or other external factors.

Maybe Windows CI tests should be reworked to use a local,
non-dockerized sshd server.

Or, perhaps consider dropping the WinCNG backend, because
without any tests it's a blind run. Predating that, WinCNG is also
unstable, so e.g. SCP/SFTP tests need to be kept disabled in curl,
because they randomly fail. It was reproducible in libssh2 CI too,
and worked around by retrying failing tests, which isn't really
a solution.

PR #1859, opened 2026-04-14
Improve bounds checking in exit-signal sig name handling

willco007:
Notes:
- Improve sig name reading to insure it doesn't return allocated but initialized memory in out of bounds case.
- Also use `string_buf` accessor in exit-status case to retrieve status code.

Credit:
Kit Knox

### DIFF
--- a/src/packet.c
+++ b/src/packet.c
@@ -1195,7 +1195,6 @@ libssh2_packet_add_jump_point1:
                                         channelp->local.id,
                                         channelp->remote.id));
                     }
-
                 }
                 else if(len == strlen("exit-signal") &&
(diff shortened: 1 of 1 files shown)


kitknox:
Was it intentional to remove the error handling on LIBSSH2_ALLOC channelp->exit_signal?  Doesn't this introduce the chance of NULL deref on the subsequent memcpy()?

willco007:
@kitknox It was not, I will get it added back in. Thanks for the heads up.

willco007:
Fixed here https://github.com/libssh2/libssh2/pull/1861

PR #1873, opened 2026-04-23
INSTALL_CMAKE: update minimum version and links, drop old cmake instructions

vszakats:
Follow-up to ac7b9b214afaea60862a8648681889b7f58eafeb #1839


vszakats:
GitHub API broken with 503, merging via web UI.

PR #1878, opened 2026-04-23
mbedtls: stop using `mbedtls_pk_load_file()`

vszakats:
This mbedTLS API is no longer exported by mbedTLS 3.6.0+. Replace with
local implementation.

Limit size of loaded keys to 1 MiB.

Follow-up to 1628f6cab8d2ffde0a02e2110ca33e99675c7afc #1421
Follow-up to e973493f992313b3be73f51d3f7ca6d52e288558 #1393
Follow-up to 2e4c5ec4627b3ecf4b6da16f365c011dec9a31b4 #1349

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -1250,11 +1250,6 @@ cleanup:
     return *ctx ? 0 : -1;
 }
 
-/* Force-expose internal mbedTLS function */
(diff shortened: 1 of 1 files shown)


willco007:
@vszakats  A max file length check should also be added to `_libssh2_pem_parse()` if you have a moment, thanks!

PR #1913, opened 2026-05-01
userauth: limit input sizes to prevent allocation size overflow

karnakarreddi:
Several functions sum caller- or key-file-derived size_t lengths to size
the packet allocation, with no overflow check. On 32-bit builds, crafted
oversized lengths wrap the sum to a small value, producing an undersized
heap buffer that is then overflowed by the subsequent
_libssh2_store_str() calls. The post-alloc null check doesn't help - the
wrap happens before LIBSSH2_ALLOC() is called, so the allocator returns
a valid (but undersized) buffer.

Fix: reject any of the input lengths greater than 64 KiB before the size
arithmetic. SSH usernames, hostnames, method names, and public-key blobs
are bounded well below this in legitimate use, so the cap rejects only
malicious or corrupt input while making the arithmetic unconditionally
safe to fits comfortably in a 32-bit size_t. Cleanup on the new error
path mirrors the existing LIBSSH2_ERROR_ALLOC branch immediately below.

Assisted-by: Viktor Szakats


src/userauth.c
```diff
@@ -1075,6 +1075,28 @@ userauth_hostbased_fromfile(LIBSSH2_SESSION *session,
...
+         * cap prevents the size_t additions below from wrapping on
+         * 32-bit (or any) platforms, which would otherwise lead to an
+         * undersized heap allocation followed by out-of-bounds writes
+         * in the _libssh2_store_str() calls that follow.
+         */
+        if(username_len                       > 65536 ||
+           session->userauth_host_method_len  > 65536 ||
+           hostname_len                       > 65536 ||
+           local_username_len                 > 65536 ||
+           pubkeydata_len                     > 65536) {
```
vszakats:
```suggestion
        #define MAX_INPUT_LEN 0x10000  /* 64 KiB */
        if(username_len > MAX_INPUT_LEN ||
           session->userauth_host_method_len > MAX_INPUT_LEN ||
           hostname_len > MAX_INPUT_LEN ||
           local_username_len > MAX_INPUT_LEN ||
           pubkeydata_len > MAX_INPUT_LEN) {
```
What do you think of moving the limit to a macro?

Then it and the explanatory comment can be moved to the top of the file
for reuse for other similar cases.

Also I suggest dropping the vertical alignment as its difficult to maintain.
To retain alignment maybe `MAX_INPUT_LEN < var` could be used but it
reduces readability in a different way, so I'm unsure if it'd be worth it.

karnakarreddi: make sense!

karnakarreddi: Pushed an update — moved the limit to a MAX_INPUT_LEN macro at the top of the file with the explanatory comment, dropped the vertical alignment, and applied the same check to the other affected spots in userauth.c.  PTAL @vszakats  🙏

src/userauth.c
```diff
@@ -1615,6 +1634,15 @@ _libssh2_userauth_publickey(LIBSSH2_SESSION *session,
...
         }
 
+        if(username_len > MAX_INPUT_LEN ||
+           session->userauth_pblc_method_len > MAX_INPUT_LEN ||
+           pubkeydata_len > MAX_INPUT_LEN) {
+            LIBSSH2_FREE(session, session->userauth_pblc_method);
+            session->userauth_pblc_method = NULL;
+            return _libssh2_error(session, LIBSSH2_ERROR_INVAL,
+                                  "Input lengths too large");
+        }
```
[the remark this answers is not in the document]
karnakarreddi: @vszakats  This make sense to me too. so i update the condition and pushed the changes

willco007:
Looks good to me, are you satisfied with it @vszakats?

src/userauth.c
```diff
@@ -1576,7 +1608,8 @@ _libssh2_userauth_publickey(LIBSSH2_SESSION *session,
         if(!session->userauth_pblc_method) {
             session->userauth_pblc_method_len = _libssh2_ntohu32(pubkeydata);
 
-            if(session->userauth_pblc_method_len > pubkeydata_len - 4)
+            if(session->userauth_pblc_method_len > MAX_INPUT_LEN ||
+                session->userauth_pblc_method_len > pubkeydata_len - 4)
```
vszakats:
```suggestion
            if(session->userauth_pblc_method_len > MAX_INPUT_LEN ||
               session->userauth_pblc_method_len > pubkeydata_len - 4)
```
Let's fix this indent.

vszakats:
@willco007, @karnakarreddi LGTM, thanks for the updates.

vszakats:
Thank you @karnakarreddi, merged now.

PR #1930, opened 2026-05-12
GHA: use shorthand 'AM', 'CM' in job names

vszakats:
AM = autotools/automake
CM = cmake


vszakats:
Job names still long and often difficult to read, but it's a small step.

PR #1944, opened 2026-05-16
GHA: explicitly `brew update` before `brew install` with Linuxbrew

vszakats:


vszakats:
Rather not, due to speed concerns.

PR #1945, opened 2026-05-16
transport: set `OUTBOUND` on `send_existing` `EAGAIN` exits

### LINKED ISSUE #1672: The write stucks at the transport layer.
I'm using channel, created with libssh2_channel_direct_tcpip2(). Sometimes when the program writes a lot of data, writing process stops and all calls of libssh2_channel_write() returns LIBSSH2_ERROR_EAGAIN.

After some debug I found the reason and fix the problem:
There is key exchange procedure, that starts by [_libssh2_transport_send()](https://github.com/libssh2/libssh2/blob/a30ea29bc36dbb0fd2995f40b3d8aa7b4f04f2b2/src/transport.c#L1030).
Also the _libssh2_transport_send tryes to continue sending existing packet if any by calling [send_existing()](https://github.com/libssh2/libssh2/blob/a30ea29bc36dbb0fd2995f40b3d8aa7b4f04f2b2/src/transport.c#L1041C10-L1041C23). 
send_existing() checks that we are trying to send the same packet that allready exists (if any). If packets are not the same, send_existing() exits with [LIBSSH2_ERROR_EAGAIN](https://github.com/libssh2/libssh2/blob/a30ea29bc36dbb0fd2995f40b3d8aa7b4f04f2b2/src/transport.c#L926).
But what happens if we need key exchange and send existing at the same time?
1. _libssh2_transport_send() calls _libssh2_kex_exchange()
2. _libssh2_kex_exchange() makes it's own packet and tryies to send it with _libssh2_transport_send()
3. _libssh2_transport_send() calls send_existing()
4. send_existing() compares new packet address and existing packet address. Test fails and we get LIBSSH2_ERROR_EAGAIN
5. All subsequent write calls exits with LIBSSH2_ERROR_EAGAIN by this scheme

My suggestion is simply call send_existing() before _libssh2_kex_exchange() in _libssh2_transport_send(). That solves the problem.

**To Reproduce**
The bug is rather random. It's hard to reproduce. You need to create channel with libssh2_channel_direct_tcpip2() and write data as many as possible.

**Expected behavior**
Succesful write attempts sooner or latter.

**Version:**
 - OS: Ubuntu 22.04 jammy
 - Kernel: x86_64 Linux 6.8.0-83-generic
 - libssh2 version: 1.11.1
 - crypto backend and version: OpenSSL from repository

**Additional context**
Also I'm using read/write callbacks to run SSH session through custom IO layer, but it doesnt matter I gues.

alexanderadam:
I also fell into this and initially I assumed other issues in my stack. :laughing: 

`send_existing()` in `src/transport.c` has three `EAGAIN` exits. Two of them correctly `OR` `LIBSSH2_SESSION_BLOCK_OUTBOUND` into `session->socket_block_directions` before returning.
However, the "Address is different" branch does not. When this fires, the caller gets `EAGAIN` with `bd == 0`.

The branch is fired whenever something else on the same session calls `_libssh2_transport_send` with a different `data` pointer while `p->olen != 0` from an unfinished write.
The obvious cases are interleaved channel writes, but it also fires from inside libssh2 itself when `_libssh2_kex_exchange` builds its own packet during a write, which is exactly what #1672 describes from a different angle.

There is also a 'twin defect' at the partial-send return at the end of the same function: if `LIBSSH2_SEND` accepted some bytes but not all. More are still pending, but `bd` doesn't get updated.
This is usually already set from a prior iteration, but if the prior transport call was a clean read, `bd` is `0` and you get the same hang.

Fixes #1672

Related #1397, #1431, #1454 the regression chain that introduced the `EAGAIN` return on this branch in the first place. The fix there was correct for the data-corruption symptom but left `bd` unset, which is what #1672 reports and what this PR addresses.

**PS:** Thank you so much for your work! :pray: So many people and projects are depending on libssh2 and it's a typical standing-on-the-shoulders-of-giants project.
**PPS:** huh? How is the CI green now? I'm pretty sure that I got an email that it failed!? :thinking: 
**PPPS:** I'm looking for a new adventure in case anybody is looking to hire someone or in case someone would like to work with a (Ruby/Rails/Crystal) dev and I'm also open to move to a PO/PM role.


willco007:
Sorry for the delay, could you pull up and resolve the conflicts and we'll get this integrated.

alexanderadam:
@willco007 done 😊

src/channel.c
```diff
@@ -1941,6 +1941,20 @@ ssize_t ssh2_channel_read(LIBSSH2_CHANNEL *channel, int stream_id,
...
+     * never enters; transport_read then sets BLOCK_INBOUND on its way
+     * back, and the terminal !bytes_read branch returns EAGAIN. The
+     * caller is then told to wait for inbound bytes they never asked
+     * for, and spins until they give up. Common trigger: an HTTP client
+     * over a forwarded channel where a bounded-body reader ends up
+     * asking for 0 bytes once the body length is fully accounted for. */
+    if(buflen == 0) {
+        if(channel->remote.eof || channel->remote.close)
+            return 0;
+        return 0;
```
vszakats, line 1955:
```suggestion
        return 0;
```
was there something meant to be done under the nested if? if not, let's delete it.

tests/test_transport_block_directions.c
```diff
@@ -0,0 +1,202 @@
+/* Copyright (C) The libssh2 project and its contributors.
+ *
+ * regression test for #1672: every EAGAIN exit in send_existing() must
+ * leave block_directions non-zero so the caller knows what to wait on.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
```
vszakats, line 8:
```suggestion
/* Copyright (C) The libssh2 project and its contributors.
 *
 * SPDX-License-Identifier: BSD-3-Clause
 */

/* Regression test for #1672: every EAGAIN exit in send_existing() must
   leave block_directions non-zero so the caller knows what to wait on. */

```
Let's keep the copyright header and description separate.

Also could we drop the reference to #1672 and make this comment
self-explanatory, yet short?

tests/test_transport_block_directions.c
```diff
@@ -0,0 +1,202 @@
...
+        fprintf(stderr, "libssh2_session_init() failed\n");
+        libssh2_exit();
+        return 1;
+    }
+
+    if(test_addr_mismatch(session) != 0)
+        failures++;
+    if(test_send_eagain(session) != 0)
+        failures++;
+    if(test_partial_send(session) != 0)
```
vszakats, line 195:
```suggestion
    if(test_addr_mismatch(session))
        failures++;
    if(test_send_eagain(session))
        failures++;
    if(test_partial_send(session))
```
let's omit `!= 0` to match rest of codebase.

tests/test_transport_block_directions.c
```diff
@@ -0,0 +1,202 @@
...
+ */
+
+#include "libssh2_priv.h"
+#include "transport.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
```
vszakats, line 16:
```suggestion
#include <assert.h>
```
drop headers already included via libssh2_priv.h and libssh2.h.
also drop `stdlib.h` unless it's required, in which case I suggest
adding a comment saying `/* for funcname() */`.

PR #1954, opened 2026-05-29
Extend event callbacks

d3x0r:
Adds callbacks to channels for EOF and data events; previously close was the only channel callback; and that callback was only used internally by the sftp module.

``` c
    sftp_handle->channel->abstract = sftp_handle;
    sftp_handle->channel->close_cb = libssh2_sftp_dtor;
```

Added similar registration of callbacks that sessions have.  Added a connect event on the listener, which dispatches a newly connected channel to a callback, which can then set the callbacks on that channel for itself.

Added markdown document pages for the new interfaces.


d3x0r:
I need to clean this better.....

PR #2044, opened 2026-06-11
Check for lang_cs and lang_sc values #2016

### LINKED ISSUE #2016: kex_agree_methods() skips lang_cs/lang_sc name-lists, reads burn_optimistic_kexinit from wrong offset
## Description

`kex_agree_methods()` in `kex.c` reads only 8 of the 10 SSH KEXINIT name-lists, skipping `lang_cs` (client-to-server languages) and `lang_sc` (server-to-client languages). The language parsing is inside a `#if 0` block (lines 4173-4180) and is never compiled.

After reading 8 name-lists, the code reads `burn_optimistic_kexinit` from the buffer position that actually corresponds to `lang_cs`'s 4-byte big-endian length field, not the actual `first_kex_packet_follows` boolean.

## Location

`src/kex.c`, lines 4123-4148 (`kex_agree_methods()`)

## Impact

For servers sending **empty** language lists (virtually all real servers), `lang_cs_len = 0`, and the first byte of its big-endian length encoding is `0x00`. This accidentally produces the correct value (`burn_optimistic_kexinit = 0`), masking the bug.

For servers sending language lists with `len >= 0x01000000` (unusual but RFC-compliant), the first byte would be non-zero, causing `burn_optimistic_kexinit` to be set incorrectly. This could cause the client to skip or discard the first KEXINIT packet, leading to key exchange state machine confusion.

## Root Cause

```c
// Lines 4123-4138: reads 8 name-lists
_libssh2_get_string(&buf, &kex, &kex_len);
_libssh2_get_string(&buf, &hostkey, &hostkey_len);
// ... 6 more ...

// Line 4143: reads what it thinks is first_kex_packet_follows
// but buf.dataptr is actually at the lang_cs length field!
if(_libssh2_check_length(&buf, 1)) {
    session->burn_optimistic_kexinit = *(buf.dataptr++);
}
```

The skipped parsing is at lines 4173-4180:
```c
#if 0
    if(libssh2_kex_agree_lang(session, &session->local, lang_cs, lang_cs_len) ||
       libssh2_kex_agree_lang(session, &session->remote, lang_sc, lang_sc_len)) {
        return -1;
    }
#endif
```

## Suggested Fix

Either:
1. Remove the `#if 0` and implement `libssh2_kex_agree_lang()`, or
2. Skip the two language name-lists explicitly before reading `burn_optimistic_kexinit`:
```c
/* skip language name-lists (not negotiated) */
{
    unsigned char *tmp;
    size_t tmp_len;
    _libssh2_get_string(&buf, &tmp, &tmp_len); /* lang_cs */
    _libssh2_get_string(&buf, &tmp, &tmp_len); /* lang_sc */
}

if(_libssh2_check_length(&buf, 1)) {
    session->burn_optimistic_kexinit = *(buf.dataptr++);
}
```

## Reproduction

Verified with a test program that constructs a KEXINIT message with non-empty language lists and feeds it through `_libssh2_packet_add()`. Confirmed that `burn_optimistic_kexinit` reads from the `lang_cs` offset.

willco007:
Read lang_cs and lang_sc key exchange values according to [RFC]( https://datatracker.ietf.org/doc/html/rfc4253#section-7.1) even if they are empty from most servers. #2016.

Credit: afldl from github
Fixes #2016

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -4045,7 +4045,7 @@ static int kex_agree_methods(LIBSSH2_SESSION *session, unsigned char *data,
                              size_t data_len)
 {
     unsigned char *kex, *hostkey, *crypt_cs, *crypt_sc, *comp_cs, *comp_sc,
-        *mac_cs, *mac_sc;
+        *mac_cs, *mac_sc, *tmp;
     size_t kex_len, hostkey_len, crypt_cs_len, crypt_sc_len, comp_cs_len;
     size_t comp_sc_len, mac_cs_len, mac_sc_len;
     struct string_buf buf;
@@ -4077,6 +4077,12 @@ static int kex_agree_methods(LIBSSH2_SESSION *session, unsigned char *data,
     if(ssh2_get_string(&buf, &comp_cs, &comp_cs_len))
         return -1;
     if(ssh2_get_string(&buf, &comp_sc, &comp_sc_len))
+        return -1;
+
+    /* read lang_cs and lang_sc but are unused */
+    if(_libssh2_get_string(&buf, &tmp, NULL))
+        return -1;
+    if(_libssh2_get_string(&buf, &tmp, NULL))
         return -1;
 
     /* If the server sent an optimistic packet, assume that it guessed wrong.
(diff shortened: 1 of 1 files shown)


vszakats:
The linter fail is unrelated and goes away with a rebase on master.

vszakats:
@willco007 I've added `Fixes #2016` so that it auto-closes and tracks the reference. Let's make this line into the commit msg on merge.

PR #2046, opened 2026-06-11
agent: fix bounds checks, and potential OOB access on 32-bit platforms in `agent_sign()`

### LINKED ISSUE #2021: agent_sign() integer overflow on 32-bit platforms leads to heap buffer overflow
## Description

In `agent_sign()` (agent.c:766), the request buffer length is computed as:

```c
ssize_t len = 1 + 4 + identity->external.blob_len + 4 + data_len + 4;
```

On **32-bit platforms**, `identity->external.blob_len` and `data_len` are both `size_t` (32-bit). If their sum plus 13 exceeds `UINT32_MAX`, the expression wraps to a small value. The subsequent `LIBSSH2_ALLOC(session, len)` allocates only that small amount, but `_libssh2_store_str()` at lines 783-784 writes `blob_len` bytes into the tiny buffer, causing a **heap buffer overflow**.

Additionally, at line 891, the signed/unsigned mismatch in the bounds check:
```c
len -= *sig_len;   // ssize_t -= size_t
if(len < 0) {      // always false when *sig_len > INT32_MAX
```
When `*sig_len > INT32_MAX`, the subtraction wraps in unsigned arithmetic and the `< 0` check never fires, allowing the subsequent `memcpy(*sig, s, *sig_len)` at line 902 to read out of bounds from the response buffer.

## Location

`src/agent.c`, lines 766, 891-902 (`agent_sign()`)

## Impact

- **CWE-190**: Integer Overflow or Wraparound
- **CWE-787**: Out-of-bounds Write (heap)

**Severity**: Medium (32-bit only, requires malicious/compromised ssh-agent)

### Concrete Example (32-bit)

```
blob_len = 0xFFFFFF00 (~4GB - 256)
data_len = 0x200 (512)

len = 13 + 0xFFFFFF00 + 0x200
    = 0x10000010D  (wraps to 0x10D = 269 on 32-bit)

LIBSSH2_ALLOC(session, 269)  // allocates 269 bytes
_libssh2_store_str(&s, blob, 0xFFFFFF00)  // writes ~4GB into 269 bytes → heap overflow
```

### On 64-bit

Not exploitable — `size_t` is 64-bit and `blob_len` is bounded by the agent response's 32-bit length field.

## Attack Scenario

1. Local attacker controls or compromises the ssh-agent (e.g., via agent forwarding, Unix socket access)
2. Agent returns a crafted identity with `blob_len` designed to cause overflow
3. When client calls `agent_sign()`, heap buffer overflow occurs
4. Note: controlling the agent socket typically already implies local code execution, but the overflow could enable further exploitation (e.g., bypassing agent restrictions)

## Suggested Fix

Use `size_t` for the length calculation and add overflow checks:

```c
size_t len = 1 + 4 + 4 + 4;  /* fixed parts */
if(identity->external.blob_len > SIZE_MAX - len ||
   data_len > SIZE_MAX - len - identity->external.blob_len) {
    return _libssh2_error(session, LIBSSH2_ERROR_OUT_OF_BOUNDARY,
                          "Agent sign request too large");
}
len += identity->external.blob_len + data_len;
```

Also fix the signed/unsigned mismatch at line 891:
```c
if(*sig_len > (size_t)len) {  /* use unsigned comparison */
    rc = LIBSSH2_ERROR_AGENT_PROTOCOL;
    goto error;
}
```

## Environment

- **Affected**: 32-bit builds (ILP32) of libssh2
- **Not affected**: 64-bit builds (LP64)

vszakats:
Also limit payload to `UINT32_MAX`, since the protocol uses 32-bit value
to store the size. (I'd guess a lower, sane limit may be even better.)

Reported-by: afldl on github
Fixes #2021

### DIFF
--- a/src/agent.c
+++ b/src/agent.c
@@ -753,8 +753,8 @@ static int agent_sign(LIBSSH2_SESSION *session,
     uint32_t sign_flags = 0;
 
     len = 1 + 4 + 4 + 4;  /* fixed parts */
-    if(identity->external.blob_len > SIZE_MAX - len ||
-       data_len > SIZE_MAX - len - identity->external.blob_len) {
+    if(identity->external.blob_len > UINT32_MAX - len ||
+       data_len > UINT32_MAX - len - identity->external.blob_len) {
         return ssh2_err(session, LIBSSH2_ERROR_OUT_OF_BOUNDARY,
                         "Agent sign request too large");
     }
@@ -800,11 +800,6 @@ static int agent_sign(LIBSSH2_SESSION *session,
     /* if no agent has been connected, bail out */
     if(!agent->ops)
         return ssh2_err(session, LIBSSH2_ERROR_BAD_USE, "agent not connected");
-
-    /* cannot pass more than 32-bit size on the wire */
-    if(transctx->request_len > UINT32_MAX)
-        return ssh2_err(session, LIBSSH2_ERROR_BAD_USE,
-                        "agent request too large");
 
     rc = agent->ops->transact(agent, transctx);
     if(rc) {
--- a/src/agent.c
+++ b/src/agent.c
@@ -746,13 +746,11 @@ static int agent_sign(LIBSSH2_SESSION *session,
     LIBSSH2_AGENT *agent = (LIBSSH2_AGENT *)(*abstract);
     struct agent_transaction_ctx *transctx = &agent->transctx;
     struct agent_publickey *identity = agent->identity;
-    size_t len;
-    size_t method_len;
+    size_t len, method_len, plain_len;
     unsigned char *s;
     int rc;
     unsigned char *method_name = NULL;
     uint32_t sign_flags = 0;
-    size_t plain_len;
 
     len = 1 + 4 + 4 + 4;  /* fixed parts */
     if(identity->external.blob_len > SIZE_MAX - len ||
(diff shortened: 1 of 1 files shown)


src/agent.c
```diff
@@ -746,14 +746,22 @@ static int agent_sign(LIBSSH2_SESSION *session,
     LIBSSH2_AGENT *agent = (LIBSSH2_AGENT *)(*abstract);
     struct agent_transaction_ctx *transctx = &agent->transctx;
     struct agent_publickey *identity = agent->identity;
-    ssize_t len = 1 + 4 + identity->external.blob_len + 4 + data_len + 4;
+    ssize_t len;
```
vszakats:
```suggestion
    size_t len;
```

src/agent.c
```diff
@@ -746,14 +746,22 @@ static int agent_sign(LIBSSH2_SESSION *session,
...
     ssize_t plain_len;
 
+    len = 1 + 4 + 4 + 4;  /* fixed parts */
+    if(identity->external.blob_len > SIZE_MAX - len ||
+       data_len > SIZE_MAX - len - identity->external.blob_len) {
+        return ssh2_err(session, LIBSSH2_ERROR_OUT_OF_BOUNDARY,
+                        "Agent sign request too large");
+    }
+    len += identity->external.blob_len + data_len;
+
```
vszakats, line 762: Bail out if len > 32-bit. The maximum on wire

vszakats:
@afldl This needed more patching than in the report #2021. Care to review and retest if it fixes the original issues?

PR #2064, opened 2026-06-15
hostkey.c: fix ed25519-cert host key initialization

willco007:
When using `ssh-ed25519-cert-v01@openssh.com` host key the init function didn't parse the cert correctly leading to a `Unable to initialize hostkey importer` error on log in.

Credit: Will Cosgrove

### DIFF
--- a/src/hostkey.c
+++ b/src/hostkey.c
@@ -1115,7 +1115,7 @@ hostkey_method_ssh_ed25519_cert_init(LIBSSH2_SESSION * session,
 {
     size_t key_len, nonce_len;
     unsigned char *key;
-    libssh2_ed25519_ctx *ctx = NULL;
+    ssh2_ed25519_ctx *ctx = NULL;
     struct string_buf buf;
     unsigned char *nonce;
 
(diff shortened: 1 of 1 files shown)


src/hostkey.c
```diff
@@ -1102,6 +1102,63 @@ static int hostkey_method_ssh_ed25519_init(LIBSSH2_SESSION *session,
...
+/*
+ * hostkey_method_ssh_ed25519_cert_init
+ *
+ * Initialize the server hostkey cert
+ */
+static int
+hostkey_method_ssh_ed25519_cert_init(LIBSSH2_SESSION * session,
+                                     const unsigned char *hostkey_data,
+                                     size_t hostkey_data_len,
+                                     void **abstract)
```
vszakats:
```suggestion
static int hostkey_method_ssh_ed25519_cert_init(
    LIBSSH2_SESSION *session,
    const unsigned char *hostkey_data,
    size_t hostkey_data_len,
    void **abstract)
```
to sync with the clang-format style now used consistently in rest of code.

(that's the folded layout when the line cannot fit within the line width limit,
as in this case)

PR #2128, opened 2026-06-28
publickey: fix potential multiplication overflow in 32-bit `libssh2_publickey_list_fetch()`

vszakats:
Cap list size at 1024 elements.

Reported-and-initial-patch-by: Mateusz Gierblinski
Reported-and-initial-patch-by: Behzod Abdullayev
Reported-by: Sharique Raza

Follow-up to e15f5d97a04cc676ce117dd324fef85b046207a9

### DIFF
--- a/src/publickey.c
+++ b/src/publickey.c
@@ -1048,15 +1048,14 @@ int libssh2_publickey_list_fetch(LIBSSH2_PUBLICKEY *pkey,
                 }
 
                 if(list[keys].num_attrs) {
-                    size_t num_attrs = list[keys].num_attrs;
-                    if(num_attrs > 1024) {
+                    if(list[keys].num_attrs > 1024) {
                         ssh2_err(session, LIBSSH2_ERROR_OUT_OF_BOUNDARY,
                                  "Too many publickey attributes");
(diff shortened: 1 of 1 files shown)


src/publickey.c
```diff
@@ -1047,17 +1047,24 @@ int libssh2_publickey_list_fetch(LIBSSH2_PUBLICKEY *pkey,
                 }
 
                 if(list[keys].num_attrs) {
+                    size_t num_attrs = list[keys].num_attrs;
+                    if(num_attrs >
+                           SIZE_MAX / sizeof(libssh2_publickey_attribute)) {
```
willco007: Instead of using `SIZE_MAX / sizeof(libssh2_publickey_attribute))` which is still very large I would prefer if we just cap num_attrs at some reasonable value. Perhaps 1024 or even a smaller value.

vszakats: Agreed!

eamanu:
it is related to CVE-2026-58051?

vszakats:
To me that CVE description rather matches a9758da45a52bc8c630ec9493804d0c6ea30b24a #2127.

PR #2146, opened 2026-06-30
channel: fix state machine in `ssh2_channel_extended_data()`

vszakats:
The worker function permanently ignored
`ignore_mode == LIBSSH2_CHANNEL_EXTENDED_DATA_IGNORE`, and never called
`ssh2_channel_flush()` via this codepath.

"The state check at line 1876 should be `ssh2_NB_state_created` instead
of `ssh2_NB_state_idle`. The state is set to `created` on line 1873, so
checking for `idle` means the flush operation will never execute. This
appears to be a logic error in the state machine."

Reported by GitHub Code Quality

### DIFF
--- a/src/channel.c
+++ b/src/channel.c
@@ -1874,7 +1874,7 @@ int ssh2_channel_extended_data(LIBSSH2_CHANNEL *channel, int ignore_mode)
         channel->extData2_state = ssh2_NB_state_created;
     }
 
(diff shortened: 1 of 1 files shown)


vszakats:
This is so old, I wonder how it never came up, if it's indeed a problem.
It seems that the ignore flag was ignored and channel never flushed.

called from public API `libssh2_channel_handle_extended_data2()`,
the publickey API init, and SFTP init.

PR #2153, opened 2026-06-30
kex: verify `ssh2_dh_secret()` success

vszakats:


vszakats:
Tests will not say much because an error may only be reported by WinCNG
and OS400.

PR #626, opened 2021-09-02
RSA SHA2 256/512 key upgrade support libssh2 #536

willco007:
Files: openssl.h, openssl.c, crypto.h, hostkey.c, HACKING-CRYPTO

Notes:
Added rsa-sha2-256 and rsa-sha2-512 as described in RFC 8332 to OpenSSL backend. To implement in other backends implementers need to define LIBSSH2_RSA_SHA2 and implement _libssh2_rsa_sha2_sign and _libssh2_rsa_sha2_verify.

Credit:
Will Cosgrove

PR #2199, opened 2026-07-03
kex: skip signaling-type kex entries when negotiating

vszakats:
Reported-by: Haruki Oyama (Waseda University)

Follow-up to 00e2a07e824db8798d94809156e9fb4e70a42f89 #1327 #1326
Follow-up to 64a555d6f5aafed504a10e5b756e85c91b1d56ce #626

---

https://github.com/libssh2/libssh2/pull/2199/files?w=1

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -3732,12 +3732,8 @@ int ssh2_kex_exchange(LIBSSH2_SESSION *session, int reexchange,
     else
         key_state->state = ssh2_NB_state_sent2;
 
-    if(rc == 0 && session->kex) {
-        if(!session->kex->exchange_keys)
-            /* This must not happen for a real KEX exchange */
-            rc = ssh2_err(session, LIBSSH2_ERROR_KEX_FAILURE,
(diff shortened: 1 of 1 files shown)


vszakats:
The initial patch seems to trip CIFuzz, pushing it to a timeout,
perhaps due to an endless loop introduced?:
https://github.com/libssh2/libssh2/actions/runs/28673642941/job/85122565390?pr=2199

vszakats:
The fuzzer seems to hang right on start with this change in,
without ever showing the `printf()` traces. Or maybe not right
on start but indeed an infinite loop combined with CI capture/
logging issue.

vszakats:
The fuzz hanger happened in another PR touching the `kex_agree*`
logic. It says to me that this isn't stricly related to this patch, or rather
that both patches touch the same weak point either in the agree code,
the fuzzer or the two interacting with each other.

Ref: https://github.com/libssh2/libssh2/pull/2444

PR #1644, opened 2025-08-28
Hybrid ML-KEM KEX algorithms support with OpenSSL 3.5

yaroslavros:
This PR adds mlkem768x25519-sha256, mlkem768nistp256-sha256 and mlkem1024nistp384-sha384 according to [draft-ietf-sshm-mlkem-hybrid-kex](https://datatracker.ietf.org/doc/draft-ietf-sshm-mlkem-hybrid-kex/) if libssh2 built against OpenSSL 3.5 or newer.

Tested against OpenSSH 10 (mlkem768x25519-sha256) and against AsyncSSH (all three hybrids).

I also reduced code duplication in `kex.c` by moving code blocks that are common for all KEXes into `process_host_key` and `finish_kex`.

PR #2200, opened 2026-07-03
kex: fix mismatched ECDSA private key free in ML-KEM cleanup

vszakats:
Replace `SSH2_FREE()` with `ssh2_ecdsa_free()` in
`kex_method_mlkem_nistp_cleanup()`. This buffer is allocated with the
crypto backend's allocator, and may also require backend-specific 
cleanup beforehand.

Follow-up to 3ba252e2a6fe04d17ea13ad870698a4028c8ab4c #1644
Cherry-picked from #2196

### DIFF
--- a/src/kex.c
+++ b/src/kex.c
@@ -1968,7 +1968,7 @@ static void kex_method_mlkem_nistp_cleanup(
     }
 
     if(key_state->private_key) {
-        SSH2_FREE(session, key_state->private_key);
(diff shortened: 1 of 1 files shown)


vszakats:
FTR this did not fix the ML-KEM flakiness seen in CI, ref : https://github.com/libssh2/libssh2/pull/1644#issuecomment-4761871304

PR #2206, opened 2026-07-04
transport: flush partial send before key re-exchange redirect

### LINKED ISSUE #2110: Deadlock in `_libssh2_transport_send` when a key re-exchange starts while a channel packet is partially sent
## Affected versions

libssh2 **1.11.1** and **current `master`** (the relevant block in `src/transport.c` has been in this order since commit cc4f9d56, 2012).

## Summary

On a **non-blocking** session doing a sustained channel write, if the peer initiates a key re-exchange (e.g. OpenSSH `RekeyLimit`) at a moment when a channel-data packet is only **partially** written to the socket, `_libssh2_transport_send()` deadlocks permanently:

- `libssh2_channel_write()` returns `LIBSSH2_ERROR_EAGAIN` forever;
- `libssh2_session_block_directions()` returns `LIBSSH2_SESSION_BLOCK_OUTBOUND` even though the socket is writable;
- the `LIBSSH2_CALLBACK_SEND` callback **stops being invoked** — libssh2 says it wants to write but never calls `send()`;
- the channel write window is `> 0` (not window-limited).

It is intermittent because it requires a partial socket write to be in flight at the instant the rekey is processed.

## Root cause

`_libssh2_transport_send()` runs the `EXCHANGING_KEYS` redirect **before** flushing a pending partial via `send_existing()`:

```c
/* src/transport.c, _libssh2_transport_send() */

if(session->state & LIBSSH2_STATE_EXCHANGING_KEYS &&
    !(session->state & LIBSSH2_STATE_KEX_ACTIVE)) {
    /* "Redirecting into the key re-exchange from _libssh2_transport_send" */
    rc = _libssh2_kex_exchange(session, 1, &session->startup_key_state);
    if(rc)
        return rc;                 /* <-- returns here while a partial is pending */
}

debugdump(...);

/* FIRST, check if we have a pending write to complete. ... */   /* <-- comment says FIRST, runs SECOND */
rc = send_existing(session, data, data_len, &ret);
...
```

`send_existing()` will only flush the pending partial if the caller passes the **same** `(data, data_len)` as the in-flight packet; otherwise it refuses:

```c
/* src/transport.c, send_existing() */
if((data != p->odata) || (data_len != p->olen)) {
    /* "Address is different, returning EAGAIN" */
    return LIBSSH2_ERROR_EAGAIN;
}
```

So when a rekey lands mid-partial, the system wedges:

1. We have a half-written **channel-data** packet pending (`p->odata` = channel data, `p->olen` > 0).
2. `_libssh2_transport_send(channel_data)` sees `EXCHANGING_KEYS` and redirects into `_libssh2_kex_exchange()` **before** reaching its own `send_existing()`, then returns its `EAGAIN`. The channel partial is never offered to `send_existing()` by this call.
3. `_libssh2_kex_exchange()` tries to send the KEXINIT via `_libssh2_transport_send(kex_data)` → `send_existing(kex_data)`. But the pending packet is the **channel** data, so `(data != p->odata)` → `send_existing()` returns `EAGAIN` ("Address is different"). The KEX packet cannot go out until the channel partial is flushed.
4. Nothing ever flushes the channel partial: `transport_send(channel_data)` always redirects first; `transport_send(kex_data)` (from the KEX) is rejected by `send_existing`. `_libssh2_transport_read` has the same redirect and likewise never flushes the outbound channel partial.

Deadlock. `block_directions` stays `OUTBOUND` (the KEX wants to send) but no `send()` is ever attempted, because the one packet that *could* be sent (the channel partial) is gated behind the redirect, and the packet the code *wants* to send (KEXINIT) is gated behind `send_existing`'s identity check.

### History

This is the unfixed remainder of [Red Hat bug #804156](https://bugzilla.redhat.com/show_bug.cgi?id=804156) ("libssh2 fails key re-exchange when write channel is saturated", virt-p2v, >64 GB transfers). Commit [`cc4f9d56`](https://github.com/libssh2/libssh2/commit/cc4f9d5679278ce41cd5480fab3f5e71dba163ed) (2012) **added** the `EXCHANGING_KEYS` redirect to `_libssh2_transport_send` so that *new* data packets are not emitted mid-rekey (which corrupts the stream). But it placed the redirect **ahead of** `send_existing`, so a packet already in transmission can no longer be completed. The fix at the time was sufficient for the common case (blocking mode never has a partial pending across calls), but the non-blocking + partial-in-flight case remained broken — and still is on `master`.

## Reproduction

Three conditions are required, all forced deterministically by the program below:

1. **Non-blocking** session (so a `send()` can return a partial and resume on a later call).
2. **A partial socket write in flight** — a `LIBSSH2_CALLBACK_SEND` callback that caps every write at 1 KB, so a ~32 KB SSH packet always leaves as a sequence of partials.
3. **A rekey landing mid-partial** — an OpenSSH server with a low `RekeyLimit` (e.g. `RekeyLimit 256K`), so rekeys are frequent. A bidirectional pump (read each iteration) ensures the server's `KEXINIT` is processed while a channel partial is pending.

```sh
cc -O2 -Wall libssh2_rekey_partial_deadlock.c -lssh2 -o rekey_repro
# server: any OpenSSH sshd with `RekeyLimit 256K` in sshd_config (+ reload); `cat` on PATH
./rekey_repro <host> <port> <user> [<privkey> [<pubkey>]]
```

Observed (localhost OpenSSH 8.7p1, `RekeyLimit 256K`, libssh2 built from the 1.11.1 release tarball):

- **stock 1.11.1**:
  ```
  DEADLOCK: no channel progress for 15s after 392400 bytes.
    block_directions     = 2   (1=INBOUND 2=OUTBOUND)
    channel write-window = 1770152  (>0 => not window-limited)
    SEND-callback calls during the stall = 1   (libssh2 wants OUTBOUND but never calls send)
  ```
  (deadlocks just past the first 256 KB rekey boundary; exit 1)
- **1.11.1 + the patch below**: `OK: sent 64 MiB with no deadlock` (exit 0)

Same source, same server, only `src/transport.c` differs.

<details>
<summary><code>libssh2_rekey_partial_deadlock.c</code> (full, ~210 lines, public API only)</summary>

```c
/*
 * libssh2 rekey deadlock reproducer (standalone -- only the libssh2 public API).
 * See the surrounding issue for the analysis. Build: cc -O2 -Wall this.c -lssh2 -o rekey_repro
 * Run against an OpenSSH server with a low RekeyLimit (e.g. `RekeyLimit 256K`).
 * STOCK libssh2 -> DEADLOCK (exit 1); FIXED -> OK (exit 0).
 */
#include <libssh2.h>
#include <netdb.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/select.h>
#include <sys/socket.h>
#include <sys/time.h>
#include <time.h>
#include <unistd.h>

#define SEND_CAP   1024    /* cap each send() -> a ~32 KB SSH packet leaves as ~32 partials */
#define TARGET_MIB 64
#define STALL_SECS 15

static unsigned long g_send_calls = 0;

static ssize_t cap_send(libssh2_socket_t sock, const void *buf, size_t len, int flags, void **abstract)
{
    (void)abstract;
    g_send_calls++;
    if(len > SEND_CAP)
        len = SEND_CAP;
    return send(sock, buf, len, flags);
}

static void wait_socket(int sock, LIBSSH2_SESSION *session)
{
    struct timeval tv = { 1, 0 };
    fd_set fd, *rd = NULL, *wr = NULL;
    int dir = libssh2_session_block_directions(session);
    FD_ZERO(&fd);
    FD_SET(sock, &fd);
    if(dir & LIBSSH2_SESSION_BLOCK_INBOUND)
        rd = &fd;
    if(dir & LIBSSH2_SESSION_BLOCK_OUTBOUND)
        wr = &fd;
    select(sock + 1, rd, wr, NULL, &tv);
}

static int tcp_connect(const char *host, const char *port)
{
    struct addrinfo hints, *res, *p;
    int sock = -1;
    memset(&hints, 0, sizeof hints);
    hints.ai_family = AF_UNSPEC;
    hints.ai_socktype = SOCK_STREAM;
    if(getaddrinfo(host, port, &hints, &res) != 0)
        return -1;
    for(p = res; p; p = p->ai_next) {
        sock = socket(p->ai_family, p->ai_socktype, p->ai_protocol);
        if(sock < 0)
            continue;
        if(connect(sock, p->ai_addr, p->ai_addrlen) == 0)
            break;
        close(sock);
        sock = -1;
    }
    freeaddrinfo(res);
    return sock;
}

int main(int argc, char **argv)
{
    if(argc < 4) {
        fprintf(stderr, "usage: %s <host> <port> <user> [<privkey> [<pubkey>]]\n"
                        "  the server's sshd_config must set a low RekeyLimit, e.g. `RekeyLimit 256K`\n", argv[0]);
        return 2;
    }
    const char *host = argv[1], *port = argv[2], *user = argv[3];
    const char *privkey = argc > 4 ? argv[4] : NULL;
    const char *pubkey = argc > 5 ? argv[5] : NULL;
    int rc;

    int sock = tcp_connect(host, port);
    if(sock < 0) { fprintf(stderr, "tcp connect to %s:%s failed\n", host, port); return 2; }
    if(libssh2_init(0)) { fprintf(stderr, "libssh2_init failed\n"); return 2; }
    LIBSSH2_SESSION *session = libssh2_session_init();
    if(!session) { fprintf(stderr, "session_init failed\n"); return 2; }

    /* partial-forcing SEND callback BEFORE the handshake, non-blocking session -- both essential */
    libssh2_session_callback_set(session, LIBSSH2_CALLBACK_SEND, (void *)cap_send);
    libssh2_session_set_blocking(session, 0);

    while((rc = libssh2_session_handshake(session, sock)) == LIBSSH2_ERROR_EAGAIN)
        wait_socket(sock, session);
    if(rc) { fprintf(stderr, "handshake failed: %d\n", rc); return 2; }

    if(privkey) {
        while((rc = libssh2_userauth_publickey_fromfile(session, user, pubkey, privkey, NULL)) == LIBSSH2_ERROR_EAGAIN)
            wait_socket(sock, session);
        if(rc) { fprintf(stderr, "publickey auth failed: %d\n", rc); return 2; }
    }
    else {
        LIBSSH2_AGENT *agent = libssh2_agent_init(session);
        struct libssh2_agent_publickey *id = NULL, *prev = NULL;
        if(!agent || libssh2_agent_connect(agent) || libssh2_agent_list_identities(agent)) {
            fprintf(stderr, "ssh-agent unavailable; pass a privkey path instead\n"); return 2;
        }
        rc = LIBSSH2_ERROR_AUTHENTICATION_FAILED;
        while(libssh2_agent_get_identity(agent, &id, prev) == 0) {
            int r;
            while((r = libssh2_agent_userauth(agent, user, id)) == LIBSSH2_ERROR_EAGAIN)
                wait_socket(sock, session);
            if(r == 0) { rc = 0; break; }
            prev = id;
        }
        libssh2_agent_disconnect(agent);
        libssh2_agent_free(agent);
        if(rc) { fprintf(stderr, "ssh-agent auth failed for user %s\n", user); return 2; }
    }

    LIBSSH2_CHANNEL *channel;
    while((channel = libssh2_channel_open_session(session)) == NULL &&
          libssh2_session_last_errno(session) == LIBSSH2_ERROR_EAGAIN)
        wait_socket(sock, session);
    if(!channel) { fprintf(stderr, "channel_open failed\n"); return 2; }

    /* fast stdin sink -- keeps the window open so the only possible stall is the deadlock */
    while((rc = libssh2_channel_exec(channel, "cat >/dev/null")) == LIBSSH2_ERROR_EAGAIN)
        wait_socket(sock, session);
    if(rc) { fprintf(stderr, "exec failed: %d\n", rc); return 2; }

    fprintf(stderr, "connected + exec'd; pushing %d MiB (SEND_CAP=%d)...\n", TARGET_MIB, SEND_CAP);

    char buf[32768];
    memset(buf, 'x', sizeof buf);
    size_t total = 0, target = (size_t)TARGET_MIB * 1024 * 1024;
    time_t last_progress = time(NULL);
    unsigned long calls_at_progress = g_send_calls;

    while(total < target) {
        /* Pump INBOUND every iteration (real clients do). cat has no stdout -> EAGAIN, but the side effect
           drains the transport and processes a pending server KEXINIT WHILE our channel-data packet is a
           partial in flight -- the coincidence that wedges _libssh2_transport_send. */
        char rbuf[4096];
        (void)libssh2_channel_read(channel, rbuf, sizeof rbuf);

        ssize_t n = libssh2_channel_write(channel, buf, sizeof buf);
        if(n == LIBSSH2_ERROR_EAGAIN) {
            if(time(NULL) - last_progress >= STALL_SECS) {
                fprintf(stderr,
                        "\nDEADLOCK: no channel progress for %ds after %zu bytes.\n"
                        "  block_directions     = %d   (1=INBOUND 2=OUTBOUND)\n"
                        "  channel write-window = %lu  (>0 => not window-limited)\n"
                        "  SEND-callback calls during the stall = %lu  (libssh2 wants OUTBOUND but never calls send)\n",
                        STALL_SECS, total, libssh2_session_block_directions(session),
                        libssh2_channel_window_write_ex(channel, NULL), g_send_calls - calls_at_progress);
                return 1;
            }
            wait_socket(sock, session);
            continue;
        }
        if(n < 0) { fprintf(stderr, "channel_write error: %ld\n", (long)n); return 2; }
        if(total / (1024 * 1024) != (total + (size_t)n) / (1024 * 1024))
            fprintf(stderr, "  ... %zu MiB\n", (total + (size_t)n) / (1024 * 1024));
        total += (size_t)n;
        last_progress = time(NULL);
        calls_at_progress = g_send_calls;
    }

    fprintf(stderr, "OK: sent %zu MiB with no deadlock.\n", total / (1024 * 1024));
    libssh2_channel_send_eof(channel);
    libssh2_channel_close(channel);
    libssh2_channel_free(channel);
    libssh2_session_disconnect(session, "done");
    libssh2_session_free(session);
    close(sock);
    libssh2_exit();
    return 0;
}
```
</details>

## Proposed fix

Run `send_existing()` (flush any pending partial) **before** the `EXCHANGING_KEYS` redirect — i.e. swap the order of the two blocks. This is what the existing `/* FIRST, check if we have a pending write to complete. */` comment already says should happen.

This is correct per RFC 4253 §7.1: after sending `SSH_MSG_KEXINIT` a party must not send *new* non-KEX packets, but it must still finish transmitting the packet already in flight (you cannot splice a `KEXINIT` into the middle of another packet). New packets are still correctly withheld, because the redirect — now reached only once `send_existing` reports nothing pending (`ret == 0`) — still runs before any new packet is marshalled.

```diff
--- a/src/transport.c
+++ b/src/transport.c
@@ _libssh2_transport_send
+    debugdump(session, "libssh2_transport_write plain", data, data_len);
+    if(data2)
+        debugdump(session, "libssh2_transport_write plain2", data2, data2_len);
+
+    /* Finish flushing any partially-sent packet BEFORE redirecting into a key re-exchange. A packet that
+       is already in transmission can only be completed by a transport_send carrying that same packet
+       (send_existing rejects a different data pointer with EAGAIN), so the redirect must not run ahead of
+       it -- otherwise a packet caught mid-send when a rekey starts can never be flushed, and the rekey
+       deadlocks. RFC 4253 7.1 requires completing the in-flight packet; only NEW packets are withheld,
+       which the redirect (reached only once nothing is pending) still does. */
+    rc = send_existing(session, data, data_len, &ret);
+    if(rc)
+        return rc;
+    session->socket_block_directions &= ~LIBSSH2_SESSION_BLOCK_OUTBOUND;
+    if(ret)
+        return rc;
+
     /*
      * If the last read operation was interrupted in the middle of a key
      * exchange, we must complete that key exchange before continuing to write
      * further data.
      *
      * See the similar block in _libssh2_transport_read for more details.
      */
     if(session->state & LIBSSH2_STATE_EXCHANGING_KEYS &&
         !(session->state & LIBSSH2_STATE_KEX_ACTIVE)) {
         _libssh2_debug((session, LIBSSH2_TRACE_TRANS, "Redirecting into the"
                        " key re-exchange from _libssh2_transport_send"));
         rc = _libssh2_kex_exchange(session, 1, &session->startup_key_state);
         if(rc)
             return rc;
     }
-
-    debugdump(session, "libssh2_transport_write plain", data, data_len);
-    if(data2)
-        debugdump(session, "libssh2_transport_write plain2", data2, data2_len);
-
-    /* FIRST, check if we have a pending write to complete. send_existing
-       only sanity-check data and data_len and not data2 and data2_len! */
-    rc = send_existing(session, data, data_len, &ret);
-    if(rc)
-        return rc;
-
-    session->socket_block_directions &= ~LIBSSH2_SESSION_BLOCK_OUTBOUND;
-
-    if(ret)
-        /* set by send_existing if data was sent */
-        return rc;
```

With this change the reproducer above prints `OK: sent 64 MiB` and exits 0; without it, it deadlocks. Happy to open a PR.

SebTardif:
## Summary

Fixes a permanent non-blocking deadlock when a key re-exchange starts while a channel packet is only partially written to the socket.

## Problem

In `ssh2_transport_send()`, the `SSH2_STATE_EXCHANGING_KEYS` redirect ran **before** `send_existing()`. If a partial packet was already in flight:

1. Rekey runs and may return `EAGAIN` or complete while the partial remains pending.
2. Later calls with different `data` pointers hit `send_existing()`'s "Address is different" path and return `EAGAIN` forever.
3. Callers see `LIBSSH2_SESSION_BLOCK_OUTBOUND` with no progress on `send()`.

## Fix

Flush any pending partial via `send_existing()` **first**, then run the rekey redirect only for **new** packets. This matches RFC 4253 §7.1 (finish the in-flight packet; withhold new packets during rekey).

The comment above the old `send_existing` call already said "FIRST"; the rekey block had been ordered ahead of it.

## Testing

- `cmake --build` (full tree including tests) succeeds on macOS arm64.
- Manual review of control flow against the reproducer described in the issue.

## References

- Closes https://github.com/libssh2/libssh2/issues/2110
- Maintainer invited a PR on that issue (2026-06-21).

### DIFF
--- a/src/transport.c
+++ b/src/transport.c
@@ -996,12 +996,35 @@ int ssh2_transport_send(LIBSSH2_SESSION *session,
     size_t orgdata_len = data_len;
     size_t crypt_offset, etm_crypt_offset;
 
+    debugdump(session, "ssh2_transport_send() plain", data, data_len);
+    if(data2)
+        debugdump(session, "ssh2_transport_send() plain2", data2, data2_len);
+
+    /* Finish flushing any partially-sent packet BEFORE redirecting into a key
+     * re-exchange. A packet already in transmission can only be completed by a
+     * transport_send call with that same packet (send_existing rejects a
+     * different data pointer with EAGAIN). If rekey runs first, a packet
+     * caught mid-send when rekey starts can never be flushed and the session
+     * deadlocks. RFC 4253 7.1 requires completing the in-flight packet; only
+     * NEW packets are withheld, which the rekey redirect (reached only once
+     * nothing is pending) still does.
+     *
+     * send_existing only sanity-checks data and data_len, not data2/data2_len.
+     */
+    rc = send_existing(session, data, data_len, &ret);
+    if(rc)
+        return rc;
+
+    session->socket_block_directions &= ~LIBSSH2_SESSION_BLOCK_OUTBOUND;
+
+    if(ret)
+        /* set by send_existing if data was sent */
+        return rc;
+
     /*
      * If the last read operation was interrupted in the middle of a key
-     * exchange, we must complete that key exchange before continuing to write
-     * further data.
-     *
-     * See the similar block in ssh2_transport_read() for more details.
+     * exchange, we must complete that key exchange before writing further
+     * *new* data. See the similar block in ssh2_transport_read().
      */
     if(session->state & SSH2_STATE_EXCHANGING_KEYS &&
        !(session->state & SSH2_STATE_KEX_ACTIVE)) {
@@ -1013,22 +1036,6 @@ int ssh2_transport_send(LIBSSH2_SESSION *session,
         if(rc)
             return rc;
     }
-
-    debugdump(session, "ssh2_transport_send() plain", data, data_len);
-    if(data2)
-        debugdump(session, "ssh2_transport_send() plain2", data2, data2_len);
-
-    /* FIRST, check if we have a pending write to complete. send_existing
-       only sanity-check data and data_len and not data2 and data2_len! */
-    rc = send_existing(session, data, data_len, &ret);
-    if(rc)
-        return rc;
-
-    session->socket_block_directions &= ~LIBSSH2_SESSION_BLOCK_OUTBOUND;
-
-    if(ret)
-        /* set by send_existing if data was sent */
-        return rc;
 
     encrypted = (session->state & SSH2_STATE_NEWKEYS) ? 1 : 0;
 


vszakats:
Thanks!

PR #2252, opened 2026-07-08
userauth: try allowing `ssh-rsa-cert-v01@openssh.com` for all backends

vszakats:
The PR implementing this did not add anything OpenSSL-specific, yet it 
gated the feature OpenSSL only. This may be due to OpenSSL supporting a
necessary feature built-in, while others don't. Yet, I wonder if enabling
it everywhere could cause issues?

Ref: https://github.com/libssh2/libssh2/pull/1314#discussion_r3542830756
Follow-up to 1410efad7ab2f11b47bf16affaa9862a721d33fd #2249
Follow-up to 3a6ab70dcfeb87a3acbb5af7ea6fc783c3981e04 #1314

### DIFF
--- a/src/userauth.c
+++ b/src/userauth.c
@@ -1311,12 +1311,9 @@ static const char *userauth_supported_key_sign_algs(LIBSSH2_SESSION *session,
 
 #if LIBSSH2_RSA_SHA2
     if((key_method_len == 7 &&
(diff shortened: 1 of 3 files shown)


vszakats:
Nope, this looks wrong, I'm not sure, but this seems to advertise
the client support this method, while it's only true with OpenSSL.

Closing without merge.

PR #2277, opened 2026-07-12
mbedtls: drop unused Blowfish, RC4, make 3DES conditional

vszakats:
RC4 and Blowfish not supported by mbedTLS 3+.
3DES not supported by mbedTLS 4+.

Follow-up to f63f68b3500ed924115d0dc39bae212cde42a87f #1727

---

https://github.com/libssh2/libssh2/pull/2277/files?w=1

### DIFF
--- a/src/mbedtls.c
+++ b/src/mbedtls.c
@@ -118,8 +118,12 @@ int ssh2_cipher_init(ssh2_cipher_ctx *ctx, SSH2_CIPHER_T(algo),
     if(!ret &&
        (algo == MBEDTLS_CIPHER_AES_128_CBC ||
         algo == MBEDTLS_CIPHER_AES_192_CBC ||
-        algo == MBEDTLS_CIPHER_AES_256_CBC ||
-        algo == MBEDTLS_CIPHER_DES_EDE3_CBC)) {
+        algo == MBEDTLS_CIPHER_AES_256_CBC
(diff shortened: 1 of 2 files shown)


vszakats:
Regarding 4.0 support: It's a sad transition story; almost all things seem to be
supported by the PSA API, but parsing key files in a way compatible with it
requires 3.6.0+. I cannot spot a way to remedy this without keeping basically
two mbedtls implementations, the existing (non-PSA) one for 3.x and a PSA
one for 4.x. I find this a subpar option, because maintaining 5 backends is already
an enourmous strain, and making that 6, with two mbedTLS ones is not helping.

Another option is to stop relying on mbedTLS' built-in key parser, and use `pem.c`
instead. This reduces the number of supported key formats with mbedTLS, but
restores compatibility with mbedTLS 2.x, which already supports the PSA API
bits necessary for libssh2. The key formats lost this way are the legacy PKCS8
ones and PEM EC, which is probably fine, as it offer bad or worse security and
have never been tested in CI either. But also may break compatibility for those
relying on them in combination with mbedTLS.

Though a precise assessment of which backends/versions support which
formats exactly would be nice.

Regardless I tend to like this option, because the missing parsing bits may be
implemented in libssh2, and most other backends would benefit from it too.

(Another option is to bump the minimum from 3.1 to 3.6.)

PR #2280, opened 2026-07-12
sftp: reject bad readdir name counts and clear stuck dir state

### LINKED ISSUE #2018: sftp_readdir() uses server-supplied num_names without validation — stuck state and memory leak
## Description

In `sftp_readdir()` (`sftp.c`, line 2006), the `num_names` field from the server's `SSH_FXP_NAME` response is stored directly without validation:

```c
num_names = _libssh2_ntohu32(data + 5);
// ...
handle->u.dir.names_left = num_names;
```

## Impact

A malicious SFTP server can send `num_names = 0xFFFFFFFF`. Since each call to `sftp_readdir()` pops one name and decrements `names_left`, if the actual packet data contains fewer names than declared:

1. **Memory leak**: `names_packet` data is only freed when `names_left == 0` (line 1922). With `names_left` never reaching zero, it leaks until the SFTP handle is closed.
2. **Stuck state**: After real data is exhausted, every subsequent `sftp_readdir()` call enters the `names_left` branch, fails a bounds check, and returns an error — **forever**. The caller cannot recover without closing the handle.

## Location

`src/sftp.c`, lines 2006-2014 (`sftp_readdir()`)

## Suggested Fix

Validate `num_names` against the remaining packet data size. Each name entry requires at minimum 4 bytes (length field), so:

```c
num_names = _libssh2_ntohu32(data + 5);
if(num_names > (data_len - 5) / 4) {
    /* server claims more names than could possibly fit */
    return _libssh2_error(session, LIBSSH2_ERROR_PROTO,
                          "Invalid num_names in FXP_NAME response");
}
handle->u.dir.names_left = num_names;
```

## Attack Scenario

1. Client connects to malicious SSH server, opens SFTP session
2. Client calls `libssh2_sftp_readdir()` on a directory
3. Server responds with `SSH_FXP_NAME` containing `num_names = 0xFFFFFFFF` but only 1 actual entry
4. Client processes 1 entry, then `names_left = 0xFFFFFFFE` — stuck returning errors forever

SebTardif:
## Summary

`sftp_readdir()` stores the server-supplied `num_names` into `names_left` without checking it against the remaining packet body. A malicious or truncated `SSH_FXP_NAME` can leave `names_left` non-zero while parse failures never free `names_packet`, causing a stuck readdir loop and leak until the handle is closed (#2018).

## Change

1. Reject `num_names` larger than `(data_len - 9) / 8` (minimum 8 bytes per name entry).
2. On entry parse failure, free `names_packet` and reset `names_left` / lengths so the next call issues a new READDIR instead of spinning.

Fixes #2018

## Test plan

- [x] CMake configure + build (`cmake --build`) including tests targets

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -1877,10 +1877,7 @@ end:
              * forever: each later call re-enters this branch and fails again
              * without ever freeing names_packet (see issue #2018). */
             if((ssize_t)filename_len < 0) {
-                if(handle->u.dir.names_packet) {
-                    SSH2_FREE(session, handle->u.dir.names_packet);
-                    handle->u.dir.names_packet = NULL;
-                }
+                SSH2_SAFEFREE(session, handle->u.dir.names_packet);
                 handle->u.dir.names_left = 0;
                 handle->u.dir.names_packet_len = 0;
                 handle->u.dir.next_name = NULL;
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -1873,6 +1873,18 @@ static ssize_t sftp_readdir(LIBSSH2_SFTP_HANDLE *handle, char *buffer,
                 SSH2_FREE(session, handle->u.dir.names_packet);
 
 end:
+            /* Parse failure with names_left still > 0 would otherwise stick
+             * forever: each later call re-enters this branch and fails again
+             * without ever freeing names_packet (see issue #2018). */
+            if((ssize_t)filename_len < 0) {
+                if(handle->u.dir.names_packet) {
+                    SSH2_FREE(session, handle->u.dir.names_packet);
+                    handle->u.dir.names_packet = NULL;
+                }
+                handle->u.dir.names_left = 0;
+                handle->u.dir.names_packet_len = 0;
+                handle->u.dir.next_name = NULL;
+            }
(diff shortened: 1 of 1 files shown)


src/sftp.c
```diff
@@ -1864,6 +1864,18 @@ static ssize_t sftp_readdir(LIBSSH2_SFTP_HANDLE *handle, char *buffer,
                 SSH2_FREE(session, handle->u.dir.names_packet);
 
 end:
+            /* Parse failure with names_left still > 0 would otherwise stick
+             * forever: each later call re-enters this branch and fails again
+             * without ever freeing names_packet (see issue #2018). */
+            if((ssize_t)filename_len < 0) {
+                if(handle->u.dir.names_packet) {
+                    SSH2_FREE(session, handle->u.dir.names_packet);
+                    handle->u.dir.names_packet = NULL;
```
vszakats:
```suggestion
                    SSH2_SAFEFREE(session, handle->u.dir.names_packet);
```
apply checksrc suggestion

vszakats (approved):

vszakats:
Thanks, this is merged now!

PR #2281, opened 2026-07-12
openssl: drop no-op code from `ssh2_sk_pub_keyfilememory()`

vszakats:
Prior to this patch it was loading a PEM-formatted private key via
OpenSSL. When successful, it did nothing, when failed, it fell back
trying to load an OpenSSH-formatted key. This was no-op in itself, but
SK keys don't support the PEM format AFAIU, so the function went to the
OpenSSH codepath when used as intended. In both cases, it leaked the
OpenSSL memory object.

Drop the superfluous intermediate step and map this function directly to
the internal worker `ossl_key_sk_from_openssh_blob()`, also fixing the
memory leak and unnecessary processing.

Also:
- sync `passphrase` type with rest of internal APIs.

Memleak reported repeatedly by GitHub Code Quality

---

"`st` is initialized to -1 at line 4242, and is only assigned in the
`if(!pk)` branch. If `PEM_read_bio_PrivateKey` succeeds (i.e., `pk` is
non-NULL), `st` is never updated and the function always returns -1. The
non-NULL `pk` case should be handled, including freeing `pk` before
returning."

"When `PEM_read_bio_PrivateKey` succeeds (pk != NULL), `st` remains -1 
and the function immediately returns it without processing the key or
freeing `pk`. This causes a memory leak of the `EVP_PKEY` object and
always returns failure for PEM-format SK keys. The function needs to
handle the successful PEM parse case, and `EVP_PKEY_free(pk)` must be
called before returning."

### DIFF
--- a/src/crypto.h
+++ b/src/crypto.h
@@ -315,7 +315,7 @@ int ssh2_sk_pub_keyfilememory(LIBSSH2_SESSION *session,
                               size_t *handle_len,
                               const char *privatekeydata,
                               size_t privatekeydata_len,
-                              const char *passphrase);
+                              const unsigned char *passphrase);
 
 #ifndef ssh2_bn_ctx
 #define ssh2_bn_ctx              int
--- a/src/misc.c
+++ b/src/misc.c
@@ -971,7 +971,7 @@ int ssh2_sk_pub_keyfilememory(LIBSSH2_SESSION *session,
                               size_t *handle_len,
                               const char *privatekeydata,
                               size_t privatekeydata_len,
-                              const char *passphrase)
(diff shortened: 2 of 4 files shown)


vszakats:
Still having a hard time understanding how is this supposed to work
and why the from-blob and from-file codepaths are different for SK
keys.

edit: even more, it's two in-memory functions, both handling sk keys.
See `ossl_key_from_openssh_blob()`. Both calling
`ssh2_openssh_pem_parse_memory()`.

edit 2: maybe that should have been two tries with the two method
names from userauth, to a standard `ossl_key_from_openssh_blob()`.

PR #2289, opened 2026-07-13
userauth: drop busyloop on EAGAIN in `libssh2_userauth_publickey_sk()`

vszakats:
"This busy-wait loop spins on `LIBSSH2_ERROR_EAGAIN` without yielding or
waiting for socket readiness. In non-blocking mode this will monopolize
the CPU without making progress, and in blocking mode it bypasses the
session's blocking infrastructure (e.g. `ssh2_wait_socket`). The
enclosing public API function should instead return
`LIBSSH2_ERROR_EAGAIN` to the caller when the inner call returns it, so
the caller can drive retries as it does for every other `userauth`
function in this file."

Reported by GitHub Code Quality

---

/cc @MichaelBuckley

### DIFF
--- a/src/userauth.c
+++ b/src/userauth.c
@@ -2392,16 +2392,10 @@ int libssh2_userauth_publickey_sk(
         return ssh2_err(session, LIBSSH2_ERROR_FILE,
                         "Invalid data in public and private key.");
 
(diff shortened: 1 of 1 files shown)


vszakats:
What I don't grok is if there is a particular reason this function was
implemented differently than all the other sibling functions.

PR #2319, opened 2026-07-15
agent: use `_tdupenv_s()`

vszakats:
Ref: https://learn.microsoft.com/cpp/c-runtime-library/reference/dupenv-s-wdupenv-s

Follow-up to 49abb5e5688f8ec7894f82564afa65379caaea24 #2313

### DIFF
--- a/src/agent.c
+++ b/src/agent.c
@@ -38,9 +38,9 @@
 #include "userauth.h"
 #include "session.h"
 
-#include <stdlib.h>  /* for getenv(), getenv_s(), _wgetenv_s() */
+#include <stdlib.h>  /* for getenv(), _dupenv_s(), _wdupenv_s() */
 #ifdef SSH2_AGENT_BACKEND_WIN32_OPENSSH
-#include <tchar.h>  /* for _tgetenv_s() */
+#include <tchar.h>  /* for _tdupenv_s() */
 #endif
 
 #define OPENSSH_AUTH_SOCK "SSH_AUTH_SOCK"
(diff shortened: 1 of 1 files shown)


vszakats:
Okay, MS CRT API woes, especially when trying to use any "recommended" ones.

This one seems to require UCRT. mingw-w64 declares it anyway, then
fails at link time:
```
/usr/bin/x86_64-w64-mingw32-ld.bfd: .libs/agent.o: in function `agent_connect_openssh':
/home/runner/work/libssh2/libssh2/bld/src/../../src/agent.c:348:(.text+0x45d): undefined reference to `__imp__dupenv_s'
```
https://github.com/libssh2/libssh2/actions/runs/29415185756?pr=2319

Revisit in the future when VS2015+ and UCRT will have become requirements.

vszakats:
Found it surprising that it builds fine back to VS2010:
https://ci.appveyor.com/project/libssh2org/libssh2/builds/54381542

vszakats:
The baseline Windows horror:
https://stackoverflow.com/questions/61553637/which-lib-do-i-have-to-link-for-wdupenv-s

Of course availability is not documented by Microsoft. According
to comments, this function has been introduced by VS2008, but
did not become part of MSVCRT.

MSYS2 still relies on the latter, but it's now a legacy environment:
https://www.msys2.org/docs/environments/#__tabbed_1_2

It means maybe not so far in the future, libssh2 may deprecate the
legacy (and insecure) MSVCRT MSYS2 envs and revisit this patch.
This also means dropping 32-bit support when using mingw-w64.
(which in turn makes libssh2 require MSVC to test 32-bit, which is
not ideal.)

There are also shaky-looking (IMO) workarounds:
https://stackoverflow.com/questions/61553637/which-lib-do-i-have-to-link-for-wdupenv-s

PR #2327, opened 2026-07-16
tests: use `ssh2_fopen()`

vszakats:
Follow-up to 521146a4f1ddd926117ae07156542da4a517521d #2318

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -1193,7 +1193,7 @@ cleanup:
     return !!*out;
 }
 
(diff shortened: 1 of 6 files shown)


vszakats:
This will need to wait till we link all tests against the static libssh2 lib.
Which in turn will need to wait till we merged tests executables into
a single bundle executable. Which would also eliminate the need for
`librunner.a`.

It's actually a happy accident that `ssh2_snprintf()` isn't affected by
this same issue.

PR #2197, opened 2026-07-03
publickey: fix potential heap OOB reads in `publickey_init()`

vszakats:
Reported-by: Mateusz Gierblinski
Reported-by: Haruki Oyama
Follow-up to ceaa7e139f400e737ad83c2d9a29d29c7f96690c #2147

PR #2338, opened 2026-07-17
publickey: fix out-of-bounds read in `publickey_init()`

abdul-khaliq-khalid:
publickey_init reads the version field with ssh2_ntohu32 after publickey_response_id has already advanced s past the length prefix and the response name, but the length guard ahead of the switch only checks pkeyInit_data_len against 4 for the whole buffer, not against what is left at s. publickey_packet_receive allocates the response buffer at exactly the length the server sent, so a server answering libssh2_publickey_init with the 11 bytes `00 00 00 07 "version"` makes that read run 4 bytes off the end of an 11-byte allocation. the status case right below it, publickey_response_success and libssh2_publickey_list_fetch all already compare against (s - data) + n, so this mirrors them rather than inventing a new shape. checked with asan driving publickey_init against that response: heap-buffer-overflow read at publickey.c:479 before, clean BUFFER_TOO_SMALL reject after. the parse sits behind channel reads so there is no seam for a unit test like test_auth_keyboard_info_request, happy to add one if you would rather have the seam.

### DIFF
--- a/src/publickey.c
+++ b/src/publickey.c
@@ -476,6 +476,12 @@ static LIBSSH2_PUBLICKEY *publickey_init(LIBSSH2_SESSION *session)
 
             case SSH2_PUBLICKEY_RESPONSE_VERSION:
                 /* What we want */
+                if(session->pkeyInit_data_len <
+                   (size_t)(s - session->pkeyInit_data) + 4) {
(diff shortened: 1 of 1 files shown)


vszakats:
Duplicate of #2197

vszakats:
Fix for this issue was merged via #2197.

PR #2342, opened 2026-07-18
transport: fix out-of-bounds write on short EtM packets

abdul-khaliq-khalid:
The EtM branch of transport_fullpacket copies blocksize-1 bytes of the decrypted first block into a heap buffer sized from decrypt_size, which is the server's packet_length field, and the read path only rejects that field when it is below 1. A server that negotiated an -etm@openssh.com MAC with a block cipher (AES-CTR/CBC, blocksize 16) can send a packet whose length is smaller than one cipher block, and since the MAC is verified first with a tag the peer can compute, the copy runs past the too-small allocation. This rejects packets whose encrypted portion is shorter than one block before the copy, matching the transport requirement that it be a whole number of cipher blocks. Confirmed with asan on the extracted arithmetic: length 1 with blocksize 16 and a 32-byte mac gives a 1-byte decrypt_buffer and a 15-byte write; clean reject after.

### DIFF
--- a/src/transport.c
+++ b/src/transport.c
@@ -239,13 +239,20 @@ static int transport_fullpacket(LIBSSH2_SESSION *session,
 
                 first_block[0] = 0;
 
+                /* the encrypted portion (everything but the cleartext length
+                   field and the trailing MAC) must be at least one whole
+                   cipher block; a shorter value would make the block copies
+                   below read and write past their buffers */
+                decrypt_size = p->total_num - mac_len - 4;
+                if(decrypt_size < blocksize)
+                    return LIBSSH2_ERROR_PROTO;
+
                 rc = transport_decrypt(session, p->payload + 4,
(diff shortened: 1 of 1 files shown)


vszakats:
This seems to be slightly modified variant of existing fix
in PR #2198.

It moves the check before `transport_decrypt()` which I
guess is fine, but why not further before `first_block[0] = 0;`?

It also ups the limit by one byte, I wonder if this is correct?

Then it leaks `p→payload` the same way as my patch attempt,
as Copilot reported there.

vszakats:
That's said I suggest to first verify if an existing patch is already in the works,
and try testing, commenting, improving on that, or acknowledge you
checked it and say why we should prefer a new take.

vszakats:
Closing in favor of pre-existing PR #2198, patch 42e33d81577ed4b95d4b4f6f845e5ee8efe5eeb4.

PR #1953, opened 2026-05-29
ossfuzz: add ssh2_sftp_packet_fuzzer and ssh2_knownhost_fuzzer

XananasX7:
## Summary

Adds two new libFuzzer harnesses targeting previously uncovered parsing paths in libssh2.

### New harness: `tests/ossfuzz/ssh2_sftp_packet_fuzzer.cc`

Complements the existing `ssh2_client_fuzzer` by prepending a valid `SSH-2.0` server banner so that `banner_receive()` in `session.c` always completes and the code enters the SSH binary-packet transport layer with fuzz-controlled payload bytes.

**Coverage added:**
- Server banner/version string parsing (`session.c:banner_receive`)
- SSH binary packet transport reader (`transport.c:_libssh2_transport_read`)
- SSH message type dispatch (`packet.c:_libssh2_packet_add`) including:
  - `SSH_MSG_DISCONNECT` string parsing
  - `SSH_MSG_DEBUG` and `SSH_MSG_IGNORE` handlers
  - `SSH_MSG_EXT_INFO` extension negotiation
  - `SSH_MSG_GLOBAL_REQUEST` parsing
  - `SSH_MSG_CHANNEL_*` family (DATA, EOF, CLOSE, REQUEST, OPEN, WINDOW_ADJUST)
  - `SSH_MSG_USERAUTH_BANNER` string extraction
  - `SSH_MSG_KEXINIT` algorithm name-list parser in `kex.c`

### New harness: `tests/ossfuzz/ssh2_knownhost_fuzzer.cc`

Exercises `libssh2_knownhost_readline()` which parses lines from an OpenSSH `known_hosts` file. Uses only the public API; no network access required.

**Coverage added:**
- Plain and comma-separated hostname parsing (`knownhost.c:oldstyle_hostline`)
- Hashed hostname parsing (`knownhost.c:hashed_hostline`, `|1|salt|hash` format)
- Key-type token recognition and base64 public-key blob decoding (`misc.c:_libssh2_base64_decode`)
- HMAC-SHA1 computation for hashed host verification
- `libssh2_knownhost_writeline()` round-trip for successfully parsed entries

### Why this matters

The existing `ssh2_client_fuzzer` exercises the client-side SSH handshake but:
- Does not bypass banner parsing (banner data and binary-packet data are interleaved in the same fuzz corpus, making it hard for the fuzzer to reach post-banner code)
[...]

PR #2363, opened 2026-07-20
knownhost: fix out-of-bounds read in `knownhost_line_hashed()`

abdul-khaliq-khalid:
The salt separator scan in `knownhost_line_hashed()` tested only `*p` and never consulted `hostlen`, so a `|1|` hashed-host line with no closing `|` walked off the end of a non-NUL-terminated span handed in by `libssh2_knownhost_readline()`. With an exact-length, unterminated buffer, `"|1|abc ssh-rsa AAAAB3NzaC1yc2E"` gives a heap-buffer-overflow read at knownhost.c:725. Bounding the scan by `hostlen` keeps it inside the buffer; valid plain and hashed lines still parse unchanged.

Narrowed to just this fix. The zero-`keylen` / `strlen()` fallback over-read that was originally in here is a separate issue, left for #2364 (or a standalone patch) to handle at the root.

### DIFF
--- a/src/knownhost.c
+++ b/src/knownhost.c
@@ -722,10 +722,10 @@ static int knownhost_line_hashed(LIBSSH2_KNOWNHOSTS *hosts,
     hostlen -= 3; /* deduct the marker */
 
     /* this is where the salt starts, find the end of it */
-    for(p = salt; *p && *p != '|'; p++)
+    for(p = salt; (size_t)(p - salt) < hostlen && *p && *p != '|'; p++)
         ;
 
-    if(*p == '|') {
+    if((size_t)(p - salt) < hostlen && *p == '|') {
         const char *hash = NULL;
         size_t saltlen = p - salt;
         if(saltlen >= (sizeof(saltbuf) - 1)) /* weird length */
@@ -862,6 +862,13 @@ static int knownhost_line(LIBSSH2_KNOWNHOSTS *hosts,
         }
         break;
     }
+
+    /* The key is passed on as base64 with its length, and knownhost_add()
+       falls back to strlen() when the length is zero, so an empty key would
+       make it scan past the caller-supplied buffer. */
+    if(!keylen)
+        return ssh2_err(hosts->session, LIBSSH2_ERROR_METHOD_NOT_SUPPORTED,
+                        "Failed to parse known_hosts line (no key)");
 
     /* Figure out host format */
     if(hostlen < 3 || memcmp(host, "|1|", 3))


src/knownhost.c
```diff
@@ -863,6 +863,13 @@ static int knownhost_line(LIBSSH2_KNOWNHOSTS *hosts,
         break;
     }
 
+    /* The key is passed on as base64 with its length, and knownhost_add()
+       falls back to strlen() when the length is zero, so an empty key would
+       make it scan past the caller-supplied buffer. */
+    if(!keylen)
+        return ssh2_err(hosts->session, LIBSSH2_ERROR_METHOD_NOT_SUPPORTED,
+                        "Failed to parse known_hosts line (no key)");
```
vszakats, line 871: Proposing to fix the root cause of this: #2364

vszakats: I suspect this may fix this address sanitizer report seen a few days ago:
```
==52==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7bd9a7be4047 at pc 0x55e309b9a76b bp 0x7ffec35dc810 sp 0x7ffec35dbfd8
READ of size 1 at 0x7bd9a7be4047 thread T0
SCARINESS: 12 (1-byte-read-heap-buffer-overflow)
    #0 0x55e309b9a76a in strlen /src/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:425:5
    #1 0x55e309c686ae in knownhost_add /src/libssh2/src/knownhost.c:143:18
    #2 0x55e309c705ff in knownhost_line_legacy /src/libssh2/src/knownhost.c:689:18
    #3 0x55e309c6d8b6 in knownhost_line /src/libssh2/src/knownhost.c:870:16
    #4 0x55e309c6cbf4 in libssh2_knownhost_readline /src/libssh2/src/knownhost.c:969:10
```
https://github.com/libssh2/libssh2/actions/runs/29666842385/job/88138609820?pr=2350#step:5:729

Since this looks like a separate issue, I'd test and merge this change in
a separate PR, if you don't mind?

(till we decide on #2364, which is breaking and needs a decision.)

vszakats:
Issue tested and reported by address sanitizer:
```
==52==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7c55dfbe19f4 at pc 0x562f17a35b1b bp 0x7ffc3d389470 sp 0x7ffc3d389468
READ of size 1 at 0x7c55dfbe19f4 thread T0
SCARINESS: 12 (1-byte-read-heap-buffer-overflow)
    #0 0x562f17a35b1a in knownhost_line_hashed /src/libssh2/src/knownhost.c:725:19
    #1 0x562f17a3297d in knownhost_line /src/libssh2/src/knownhost.c:882:16
    #2 0x562f17a31be4 in libssh2_knownhost_readline /src/libssh2/src/knownhost.c:976:10
```
https://github.com/libssh2/libssh2/actions/runs/29745783525/job/88363260911?pr=2365#step:5:833

vszakats:
Also confirmed (using PR #1953) these two patches fix the two asan issues:
https://github.com/libssh2/libssh2/actions/runs/29746738006/job/88366489751?pr=2365
(no idea why the fuzz process fails like it does anyway, it seems unrelated.)

src/knownhost.c
vszakats: Merging it as-os, to keep it simple and keep the PR message.

vszakats:
Thanks, merged now.

src/knownhost.c
abdul-khaliq-khalid: Not at all, that split makes sense to me. I've dropped the keylen hunk here, so this PR is now just the `hostlen`-bounded salt scan, and the empty-key path is yours to land wherever it fits best.

Agreed that #2364 is the better place for it. Fixing it at the parse layer only papers over the `strlen()` fallback, and the fallback is the actual hazard since any caller passing an unterminated key with zero keylen hits the same thing. The one thing worth weighing is that the behavior is documented in `libssh2.h`, so returning `LIBSSH2_ERROR_BAD_USE` will break callers that rely on it. If that turns out to be too disruptive, the parse-layer check is still available as the narrow fix.

Rebuilt with asan after the trim: the hashed case is silent and valid plain plus hashed lines still return rc=0. checksrc and spacecheck are clean.

PR #2367, opened 2026-07-20
tidy-up: miscellaneous

vszakats:
- openssl: drop always true `if` condition.
- wincng: fix misleading argument name.
- HACKING-CRYPTO.md: markdown, formatting.
- whitespace.


vszakats:
With GitHub Code Quality gone from libssh2, there will now be much less
patches like this, also much less accidental/random little bugfixes.

(Though these particular ones were not reported by GHCQ.)

PR #2369, opened 2026-07-21
userauth: keep the key method name null-terminated

abdul-khaliq-khalid:
`ssh2_userauth_plain_method()` compared with `strncmp(literal, method, method_len)`, which is a prefix match, but `session->userauth_pblc_method` was allocated at exactly `method_len` bytes and not null-terminated, so a truncated name matched a longer one. A public key blob whose leading type string is the 8 bytes `sk-ecdsa` reached the sk-ecdsa-sha2-nistp256 branch and the in-place `memcpy()` wrote 34 bytes into that 8-byte allocation; `sk-ssh` did the same through the sk-ssh-ed25519 branch. Shorter names also returned a length larger than the buffer (`s` returned 7), which the caller stores back into `userauth_pblc_method_len` and later passes to `ssh2_store_str()`. The blob is caller-supplied via `libssh2_userauth_publickey()`/`_frommemory()`, and comes from the agent identity list under `libssh2_agent_userauth()`.

Per review, the fix is to make the method name null-terminated rather than to fix up the comparisons in place. It is now allocated with one extra byte and terminated everywhere it is produced: the blob parse in `ssh2_userauth_publickey()`, the algo upgrade in `userauth_key_sign_algs()`, `userauth_read_pubkey()` (the separating space becomes the terminator), `libssh2_agent_sign()`, and the backend `method` buffers in openssl/mbedtls/wincng/os400qc3. That makes the comparisons plain `strcmp()`, makes the in-place rewrites fit by construction since the plain names are shorter than the cert names, and drops the four bugprone-not-null-terminated-result suppressions. The now-redundant length argument of `ssh2_userauth_plain_method()` is gone; `userauth_pblc_method_len` is kept for now since unwinding it reaches into `libssh2_agent_sign()`.

Confirmed with ASAN: `heap-buffer-overflow WRITE of size 34` at `ssh2_userauth_plain_method` on the `sk-ecdsa` input before, clean after, and every real cert name still maps to the same plain name and length. Picky `-Werror` build, checksrc-all.pl and spacecheck.pl are green locally.

### DIFF
--- a/src/userauth.c
+++ b/src/userauth.c
@@ -676,6 +676,8 @@ static int userauth_read_pubkey(
        it a wash */
     *method = (char *)pubkey;
     *method_len = sp1 - pubkey - 1;
+    /* keep the method name null-terminated, overwriting the separator */
+    pubkey[*method_len] = '\0';
 
     *pubkeydata = tmp;
     *pubkeydata_len = tmp_len;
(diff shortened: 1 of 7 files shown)


src/userauth.c
```diff
@@ -1135,38 +1135,42 @@ int libssh2_userauth_hostbased_fromfile_ex(LIBSSH2_SESSION *session,
 
 size_t ssh2_userauth_plain_method(char *method, size_t method_len)
 {
-    if(!strncmp("ssh-rsa-cert-v01@openssh.com",
-                method, method_len))
+    /* `method` is not null-terminated, so every name has to be matched on its
+       exact length, otherwise a truncated name matches as a prefix */
+
```
vszakats: I wonder if `method` should instead be made guaranteed to be null-terminated
throughout the code, to prevent many pitfalls and also quite a bit of suppressed
clang-tidy warnings (and LLM warnings for the pattern, as in this function.)
Possibly also allowing to drop tracking its length all over the code. I don't expect
method to ever become a binary blob.

abdul-khaliq-khalid: Makes sense, so I redid it that way instead. `session->userauth_pblc_method` is now allocated with one extra byte and null-terminated everywhere it is produced (the blob parse, the algo upgrade in `userauth_key_sign_algs()`, `userauth_read_pubkey()` where the separating space becomes the terminator, `libssh2_agent_sign()`, and the backend `method` buffers in openssl/mbedtls/wincng/os400qc3). With that the comparisons here are plain `strcmp()`, the in-place rewrites fit by construction since the plain names are shorter than the cert names, and the four bugprone-not-null-terminated-result suppressions are gone.

I dropped the length argument of `ssh2_userauth_plain_method()` since it is redundant now, but left `userauth_pblc_method_len` alone; unwinding that reaches into `libssh2_agent_sign()`'s public signature, so it felt better as its own change. Happy to do that too if you want it here.

Also rebased on master. ASAN build is clean on the truncated names that used to overflow (`sk-ecdsa`, `sk-ssh`) and every real cert name still maps to the same plain name and length; picky -Werror build, checksrc-all.pl and spacecheck.pl are green.

abdul-khaliq-khalid:
Sorry about the CI noise, my last push accidentally carried a local `bld/` build tree into the commit, which is what broke the REUSE check and the configure step everywhere. Amended and force-pushed, the branch is back to just the seven source files.

src/userauth.c
vszakats: Thanks, I also started working on this locally, and some related cleanups.

vszakats:
@abdul-khaliq-khalid I did it manually then compared to your patch,
and it matches almost fully. Applied a few nits, a synced up a few lines: #2381
I believe agent.c may be missing an update here, unless I missed
something (at line ~20).

I did not do the `strncmp` → `strcmp` changes (though Copilot
warned again https://github.com/libssh2/libssh2/pull/2381#discussion_r3624377549).

Once I merged, and you do mind rebasing this patch onto master, we'll
get those too.

abdul-khaliq-khalid:
Rebased onto master and there is nothing left over: 8448596 plus 8bd003b already cover every hunk I had, including the strncmp/memcmp to strcmp switch in ssh2_userauth_plain_method() and userauth_key_sign_algs(). Your version is a bit further along than mine too (the null-byte checks and the sizeof() based truncation read better).

On agent.c: I think it is covered. libssh2_agent_sign() allocates method_len + 1 and terminates at agent.c:1243-1248, and agent_sign() does the same for method_name at agent.c:853-859, so both buffers going into the strcmp are terminated. If you meant something else at ~line 20 of that hunk, point me at it and I will look again.

Rechecked master under ASAN with the same harness I used to find this: the truncated names that used to overflow (sk-ecdsa, sk-ssh, s) now fall through untouched and every cert name still maps to the expected plain name and length, no reports. Happy to close this one as superseded unless you want the branch kept around.

vszakats:
@abdul-khaliq-khalid Thanks a lot for double-checking! My bad for the line ~20 reference, it's the line number from the `.patch`, not agent.c (argh!). I meant within `agent_sign()` as you spotted.

If there is nothing left to fix, I think it's fine to close.

abdul-khaliq-khalid:
No worries, that lines up. Master covers everything I had here, so closing as superseded by #2381. Thanks for picking it up.

PR #2386, opened 2026-07-21
publickey: honor key type length in traces

carrerasdarren-cell:
The publickey add/remove APIs accept a key type with an explicit `name_len`, but their trace messages formatted it as a null-terminated string. A caller using an exact-length buffer could therefore expose adjacent bytes in debug output.

Use the already validated length as the printf precision in all three publickey trace messages. The regression test places a sentinel suffix after the declared key type and verifies that the suffix never reaches the trace handler.

Verified with:
- debug static OpenSSL build under UBSan
- release static/shared OpenSSL build
- `test_simple` in both configurations
- `scripts/checksrc-all.pl`

### DIFF
--- a/src/publickey.c
+++ b/src/publickey.c
@@ -569,8 +569,8 @@ int libssh2_publickey_add_ex(LIBSSH2_PUBLICKEY *pkey,
     if(pkey->add_state == ssh2_NB_state_idle) {
         pkey->add_packet = NULL;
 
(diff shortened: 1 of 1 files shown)


vszakats (changes requested):
The fix LGTM, but I'm taking issue with adding 70 lines of unit test
for continiously testing 3 lines of debug output. IMO this test is
overkill in this case.

vszakats (approved):

vszakats:
Thanks, merged!

PR #2401, opened 2026-07-23
openssl: fix potential OOB read/write with AES-GCM in `ssh2_cipher_crypt()`

### LINKED ISSUE #1491: Crash on decrypt operation using OpenSSL backend
Hi, we noticed some crashes from our userbase after updating to v1.11.1. The stack trace seems consistent in all cases, the crash happens during initial handshake on `memcpy` call:
```
memcpy() at memcpy.asm(436)
_libssh2_cipher_crypt(evp_cipher_ctx_st * * ctx, const evp_cipher_st *(*)() algo, int encrypt, unsigned char * block, unsigned __int64 blocksize, int firstlast) at openssl.c(1132)
decrypt(_LIBSSH2_SESSION * session, unsigned char * source, unsigned char * dest, __int64 len, int firstlast) at transport.c(159)
_libssh2_transport_read(_LIBSSH2_SESSION * session) at transport.c(827)
_libssh2_packet_require(_LIBSSH2_SESSION * session, unsigned char packet_type, unsigned char * * data, unsigned __int64 * data_len, int match_ofs, const unsigned char * match_buf, unsigned __int64 match_len, packet_require_state_t * state) at packet.c(1503)
session_startup(_LIBSSH2_SESSION * session, unsigned __int64 sock) at session.c(830)
libssh2_session_handshake(_LIBSSH2_SESSION * session, unsigned __int64 sock) at session.c(878)
```

We're using OpenSSL 3.2.2. The issue was observed on Windows 10 and 11, 64b.
Reproduction steps are not currently known but the issue does not seem to be super rare, we already have couple of reports.

From the minidump I have at hand, I noticed that the size passed to `memcpy` function is negative. It comes from this piece of code in `openssl.c`:
```
/* length to encrypt */
const int cryptlen = (unsigned int)blocksize - aadlen - authenticationtag;
```
and is calculated as `4 - 0 - 16` and ends up being -12 which is definitely not supposed to happen. Looks like the input variable `blocksize` or `algo` (which affects `authenticationtag`) is not correct.

vszakats:
By applying two bounds checks to non-debug builds.

Reported-by: Vladimir Eli Tokarev
Fixes GHSA-c4f7-cvfc-33j7
Reported-by: Trzik on github
Fixes #1491
Thanks-to: Afonso Januário
Follow-up to 3c953c05d67eb1ebcfd3316f279f12c4b1d600b4 #797

### DIFF
--- a/src/openssl.c
+++ b/src/openssl.c
@@ -946,13 +946,15 @@ int ssh2_cipher_crypt(ssh2_cipher_ctx *ctx, SSH2_CIPHER_T(algo), int encrypt,
     const int aadlen = (is_aesgcm && IS_FIRST(firstlast)) ? 4 : 0;
     /* size of AT, if present */
     const int authenticationtag = IS_LAST(firstlast) ? authlen : 0;
-    /* length to encrypt */
-    const int cryptlen = (unsigned int)blocksize - aadlen - authenticationtag;
+    unsigned int cryptlen; /* length to encrypt */
 
     (void)algo;
 
-    assert(blocksize <= sizeof(buf));
-    assert(cryptlen >= 0);
+    if(blocksize > sizeof(buf) ||
+       blocksize < (size_t)(aadlen + authenticationtag))
+        return 1;
+
+    cryptlen = (unsigned int)blocksize - aadlen - authenticationtag;
 
 #if LIBSSH2_AES_GCM
     /* First block */


vszakats:
Updated the PR message to mention that this patch also fixes #1491.

PR #2408, opened 2026-07-23
scp: drain long C-response basenames past fixed buffer

### LINKED ISSUE #1738: scp_recv2 - cannot download file with a name longer than 246 chars
**Describe the bug**
When testing scp file download via `libssh2_scp_recv2` if the filename in the destination is longer than 246 chars the transfer fails with Error: -28 - Unterminated response from SCP server.

**To Reproduce**
Compile the scp.c example and run:
`examples/.libs/scp <destination> <user> <password> <filename longer than 246 chars>`

**Expected behavior**
The example should print on stdout the contents of the requested file.

 - OS and version: Windows 11 - Tested inside Ubuntu 24.04.3 WSL
 - libssh2 version: Tested on both 1.11.1 and master
 - crypto backend and version: Openssl

**Additional context**
Checking the source code the error is created by this code (lines 664-671 inside src/scp.c).
```C
                      if(session->scpRecv_response_len ==
                          LIBSSH2_SCP_RESPONSE_BUFLEN) {
                          /* You had your chance */
                          _libssh2_error(session, LIBSSH2_ERROR_SCP_PROTOCOL,
                                         "[2] Unterminated response "
                                         "from SCP server");
```

SebTardif:
## Summary

`libssh2_scp_recv2` fails with `LIBSSH2_ERROR_SCP_PROTOCOL`
(\"Unterminated response from SCP server\") when the server SCP `C` line
exceeds the fixed 256-byte response buffer. Mode and size always fit early;
long basenames overflow the buffer.

Parse mode and size as soon as those fields are complete. If the buffer
fills before the trailing newline, drain the unused basename until `\n`
and continue with the SCP ACK. libssh2 does not use the basename.

## Evidence

- Issue: #1738
- Unit: `tests/test_simple` covers short lines, partial name, incomplete
  size, full 256-byte long-name buffer, and malformed mode

Fixes #1738

### DIFF
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -672,7 +672,7 @@ struct _LIBSSH2_PUBLICKEY {
 #define SSH2_SCP_RESPONSE_BUFLEN     256
 
 /* Parse SCP "C" response mode/size; used by scp_recv and unit tests. */
-int ssh2_scp_parse_c_fields(const unsigned char *buf, size_t len,
+int ssh2_scp_parse_c_fields(const char *buf, size_t len,
                             long *mode_out, libssh2_int64_t *size_out);
 
 struct flags {
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -671,6 +671,10 @@ struct _LIBSSH2_PUBLICKEY {
 
 #define SSH2_SCP_RESPONSE_BUFLEN     256
 
+/* Parse SCP "C" response mode/size; used by scp_recv and unit tests. */
+int ssh2_scp_parse_c_fields(const unsigned char *buf, size_t len,
+                            long *mode_out, libssh2_int64_t *size_out);
+
 struct flags {
     int sigpipe;     /* LIBSSH2_FLAG_SIGPIPE */
     int compress;    /* LIBSSH2_FLAG_COMPRESS */
--- a/src/scp.c
+++ b/src/scp.c
@@ -99,8 +99,7 @@ int ssh2_scp_parse_c_fields(const unsigned char *buf, size_t len,
     if(!size_len)
         return (i >= len) ? SCP_C_FIELDS_INCOMPLETE : SCP_C_FIELDS_MALFORMED;
     if(i >= len)
-        /* size digits may still continue */
-        return SCP_C_FIELDS_INCOMPLETE;
+        return SCP_C_FIELDS_INCOMPLETE; /* size digits may still continue */
     if(buf[i] != ' ' && buf[i] != '\r' && buf[i] != '\n')
         return SCP_C_FIELDS_MALFORMED;
 
--- a/src/scp.c
+++ b/src/scp.c
@@ -68,7 +68,7 @@
  *
  * Returns SCP_C_FIELDS_OK, SCP_C_FIELDS_INCOMPLETE, or SCP_C_FIELDS_MALFORMED.
  */
-int ssh2_scp_parse_c_fields(const unsigned char *buf, size_t len,
+int ssh2_scp_parse_c_fields(const char *buf, size_t len,
                             long *mode_out, libssh2_int64_t *size_out)
 {
     size_t i;
@@ -695,9 +695,9 @@ static LIBSSH2_CHANNEL *scp_recv(LIBSSH2_SESSION *session,
                 int prc;
 
                 /* Complete line in the fixed buffer (common case). */
-                prc = ssh2_scp_parse_c_fields(session->scpRecv_response,
-                                              session->scpRecv_response_len,
-                                              &mode, &size);
+                prc = ssh2_scp_parse_c_fields(
+                    (const char *)session->scpRecv_response,
+                    session->scpRecv_response_len, &mode, &size);
                 if(prc != SCP_C_FIELDS_OK) {
                     ssh2_err(session, LIBSSH2_ERROR_SCP_PROTOCOL,
                              "Invalid response from SCP server");
@@ -722,9 +722,9 @@ static LIBSSH2_CHANNEL *scp_recv(LIBSSH2_SESSION *session,
                  * we have and drain the rest of the name until newline
                  * (basename is unused by libssh2).
                  */
-                prc = ssh2_scp_parse_c_fields(session->scpRecv_response,
-                                              session->scpRecv_response_len,
-                                              &mode, &size);
+                prc = ssh2_scp_parse_c_fields(
+                    (const char *)session->scpRecv_response,
+                    session->scpRecv_response_len, &mode, &size);
                 if(prc == SCP_C_FIELDS_OK) {
                     session->scpRecv_mode = mode;
                     session->scpRecv_size = size;
--- a/src/scp.c
+++ b/src/scp.c
@@ -55,7 +55,7 @@
 /* ssh2_scp_parse_c_fields() results */
 #define SCP_C_FIELDS_OK          0
 #define SCP_C_FIELDS_INCOMPLETE  1
-#define SCP_C_FIELDS_MALFORMED  (-1)
+#define SCP_C_FIELDS_MALFORMED   (-1)
 
 /*
  * Parse mode and size from an SCP "C" response line fragment.
--- a/src/scp.c
+++ b/src/scp.c
@@ -76,62 +76,50 @@ int ssh2_scp_parse_c_fields(const unsigned char *buf, size_t len,
     char tmp[32];
     char *end = NULL;
 
-    if(!buf || len < 1 || buf[0] != 'C') {
+    if(!buf || len < 1 || buf[0] != 'C')
         return SCP_C_FIELDS_MALFORMED;
-    }
 
     i = 1;
     mode_start = i;
-    while(i < len && buf[i] >= '0' && buf[i] <= '7') {
+    while(i < len && buf[i] >= '0' && buf[i] <= '7')
         i++;
-    }
     mode_len = i - mode_start;
-    if(!mode_len) {
+    if(!mode_len)
         return (i >= len) ? SCP_C_FIELDS_INCOMPLETE : SCP_C_FIELDS_MALFORMED;
-    }
-    if(i >= len) {
+    if(i >= len)
         return SCP_C_FIELDS_INCOMPLETE;
-    }
-    if(buf[i] != ' ') {
+    if(buf[i] != ' ')
         return SCP_C_FIELDS_MALFORMED;
-    }
     i++; /* skip space after mode */
 
     size_start = i;
-    while(i < len && buf[i] >= '0' && buf[i] <= '9') {
+    while(i < len && buf[i] >= '0' && buf[i] <= '9')
         i++;
-    }
     size_len = i - size_start;
-    if(!size_len) {
+    if(!size_len)
         return (i >= len) ? SCP_C_FIELDS_INCOMPLETE : SCP_C_FIELDS_MALFORMED;
-    }
-    if(i >= len) {
+    if(i >= len)
         /* size digits may still continue */
         return SCP_C_FIELDS_INCOMPLETE;
-    }
-    if(buf[i] != ' ' && buf[i] != '\r' && buf[i] != '\n') {
+    if(buf[i] != ' ' && buf[i] != '\r' && buf[i] != '\n')
         return SCP_C_FIELDS_MALFORMED;
-    }
 
-    if(mode_len >= sizeof(tmp) || size_len >= sizeof(tmp)) {
+    if(mode_len >= sizeof(tmp) || size_len >= sizeof(tmp))
         return SCP_C_FIELDS_MALFORMED;
-    }
 
     memcpy(tmp, buf + mode_start, mode_len);
     tmp[mode_len] = '\0';
     /* !checksrc! disable BANNEDFUNC 1 */
     *mode_out = strtol(tmp, &end, 8);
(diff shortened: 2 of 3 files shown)


src/scp.c
```diff
@@ -49,6 +49,93 @@
...
+                            long *mode_out, libssh2_int64_t *size_out)
+{
+    size_t i;
+    size_t mode_start, mode_len, size_start, size_len;
+    char tmp[32];
+    char *end = NULL;
+
+    if(!buf || len < 1 || buf[0] != 'C') {
+        return SCP_C_FIELDS_MALFORMED;
+    }
```
vszakats:
```suggestion
    if(!buf || len < 1 || buf[0] != 'C')
        return SCP_C_FIELDS_MALFORMED;
```
formatting: please drop `{}` for single instruction blocks.
also below.

src/scp.c
```diff
@@ -577,149 +664,152 @@ static LIBSSH2_CHANNEL *scp_recv(LIBSSH2_SESSION *session,
...
+            if(session->scpRecv_response_len == SSH2_SCP_RESPONSE_BUFLEN) {
+                long mode = 0;
+                libssh2_int64_t size = 0;
+                int prc;
+
+                /*
+                 * Fixed buffer is full without a newline. Mode and size always
+                 * fit early; a long basename overflows the buffer. Parse what
+                 * we have and drain the rest of the name until newline
+                 * (basename is unused by libssh2). See issue #1738.
```
vszakats:
```suggestion
                 * (basename is unused by libssh2). See issue #1738.
```
Let's drop GitHub issue references. Information is there in the
commit, which is searchable. #1738 may not say much in the
longer future.

Also below.

tests/test_simple.c
```diff
@@ -124,6 +124,78 @@ static int test_ssh2_dh_validate(void)
...
+        (const unsigned char *)"C0644 12", 8, &mode, &size);
+    if(prc != 1) {
+        fprintf(stderr, "scp_parse incomplete size: prc=%d (want 1)\n", prc);
+        err++;
+    }
+
+    /*
+     * Reproduce #1738 shape: fixed buffer full of "Cmode size " + long name
+     * without newline. Mode and size must still parse so scp_recv can drain.
+     * Avoid snprintf: old MSVC (AppVeyor VS2010) lacks it.
```
vszakats: Why not use `ssh2_snprintf()`? (it's now fixed up for old MSVC version to be `snprintf()`-compatible)

tests/test_simple.c
```diff
@@ -124,6 +124,78 @@ static int test_ssh2_dh_validate(void)
...
+    size_t i;
+    int prc;
+    int err = 0;
+
+    /* Normal complete line with short name */
+    prc = ssh2_scp_parse_c_fields(
+        (const unsigned char *)"C0644 123 shortname\n", 20, &mode, &size);
+    if(prc || mode != 420L || size != 123) { /* 0644 octal == 420 */
+        fprintf(stderr, "scp_parse short: prc=%d mode=%ld size=%lld\n",
+                prc, mode, (long long)size);
```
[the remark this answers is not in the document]
vszakats: I confirm. There is no `ssh2_snprintf()` in tests (and examples),
and for `fprintf()` this mask is not guaranteed to work. The
fallback practice is to cast to e.g. `unsigned long` in this case,
which can be printed portably.

SebTardif:
Addressed review notes:

- Dropped braces on single-statement `if`/`while` in `ssh2_scp_parse_c_fields`
- Removed GitHub issue numbers from comments (kept in commit/`Fixes`)
- Switched the long-name unit test prefix to `ssh2_snprintf`
- Failure paths in the unit test now use `SSH2_INT64_T_FORMAT` instead of hard-coded `%lld`

Latest tip: `2a0d2eea`

tests/test_simple.c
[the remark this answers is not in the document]
vszakats: Hm, let's see how `SSH2_INT64_T_FORMAT` passes the builds,
it may work after all. (it's an ever confusing topic and I may well
be mistaken with the above comment.)

tests/test_simple.c
```diff
@@ -124,6 +124,77 @@ static int test_ssh2_dh_validate(void)
...
+        fprintf(stderr, "scp_parse incomplete size: prc=%d (want 1)\n", prc);
+        err++;
+    }
+
+    /*
+     * Fixed buffer full of "Cmode size " + long name without newline. Mode and
+     * size must still parse so scp_recv can drain the unused basename.
+     */
+    prefix_len = (size_t)ssh2_snprintf((char *)long_line, sizeof(long_line),
+                                       "C0644 99 ");
```
vszakats: Ah no, sorry, we're in tests, and in tests `ssh2_snprintf()`
isn't supported. It needs relatively complicate build overhaul.

SebTardif: Thanks for the correction. Dropped `ssh2_snprintf` from the unit test and build the long-name fixture with `memcpy` again. Failure logs now use `%lu` with an `unsigned long` cast, as you suggested for portable test `fprintf`.

Also distinguish full-buffer parse failures: `Invalid response` when malformed, `Unterminated` when incomplete.

Tip: `34c9d1d3`

src/scp.c
```diff
@@ -49,6 +49,81 @@
...
+int ssh2_scp_parse_c_fields(const unsigned char *buf, size_t len,
+                            long *mode_out, libssh2_int64_t *size_out)
+{
+    size_t i;
+    size_t mode_start, mode_len, size_start, size_len;
+    char tmp[32];
+    char *end = NULL;
+
+    if(!buf || len < 1 || buf[0] != 'C')
+        return SCP_C_FIELDS_MALFORMED;
```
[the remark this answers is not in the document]
SebTardif: Internal helper only; both call sites (scp_recv and the unit test) always pass non-NULL stack pointers. Adding NULL guards here would be defensive noise rather than a real failure mode.

src/scp.c
```diff
@@ -49,6 +49,81 @@
 /* Max. length of a quoted string after scp_shell_quotearg() processing */
 #define shell_quotedsize(s)  (3 * strlen(s) + 2)
 
+/* Cap on basename bytes drained after the fixed response buffer is full */
+#define SCP_C_NAME_DRAIN_MAX  (64 * 1024)
+
+/* ssh2_scp_parse_c_fields() results */
+#define SCP_C_FIELDS_OK          0
+#define SCP_C_FIELDS_INCOMPLETE  1
+#define SCP_C_FIELDS_MALFORMED  (-1)
```
vszakats:
```suggestion
#define SCP_C_FIELDS_MALFORMED   (-1)
```
alignment nit

SebTardif: Done in the tip commit (extra space so the values line up).

tests/test_simple.c
```diff
@@ -124,6 +125,78 @@ static int test_ssh2_dh_validate(void)
...
+    /*
+     * Fixed buffer full of "Cmode size " + long name without newline. Mode and
+     * size must still parse so scp_recv can drain the unused basename.
+     * Use memcpy for the prefix (tests do not call ssh2_snprintf).
+     */
+    {
+        static const char prefix[] = "C0644 99 ";
+        prefix_len = sizeof(prefix) - 1;
+        memcpy(long_line, prefix, prefix_len);
+    }
```
vszakats, line 172: I'm not sure if it simplifies, but the _standard_ `snprintf()` is available
in tests (it's translated to `_snprintf()` for old MSVC). IIRC it's fine for
basic use, if it helps the code.

tests/test_simple.c
```diff
@@ -124,6 +125,78 @@ static int test_ssh2_dh_validate(void)
...
+    for(i = prefix_len; i < sizeof(long_line); i++)
+        long_line[i] = 'a';
```
vszakats: Can this be a `memset()`?

src/scp.c
```diff
@@ -49,6 +49,81 @@
...
+
+    size_start = i;
+    while(i < len && buf[i] >= '0' && buf[i] <= '9')
+        i++;
+    size_len = i - size_start;
+    if(!size_len)
+        return (i >= len) ? SCP_C_FIELDS_INCOMPLETE : SCP_C_FIELDS_MALFORMED;
+    if(i >= len)
+        /* size digits may still continue */
+        return SCP_C_FIELDS_INCOMPLETE;
```
vszakats:
```suggestion
        return SCP_C_FIELDS_INCOMPLETE; /* size digits may still continue */
```
merge to save a line

tests/test_simple.c
SebTardif: Switched the prefix to standard `snprintf` and the basename fill to `memset` in `3b994829`.

tests/test_simple.c
SebTardif: Done: `memset(long_line + prefix_len, 'a', ...)` in `3b994829`.

src/scp.c
SebTardif: Merged the comment onto the return line in the tip commit.

SebTardif:
AppVeyor VS2010 failed on `test_simple.c` with C4013 (`snprintf` undefined under `/WX`).

`libssh2_setup.h` only maps `snprintf` to `_snprintf` when `LIBSSH2_TESTS` is defined. `test_simple` included `libssh2_priv.h` without that define (unlike `session_fixture.h`).

Fixed in `d9776561` by adding `#define LIBSSH2_TESTS` before the priv header include.

vszakats:
> AppVeyor VS2010 failed on `test_simple.c` with C4013 (`snprintf` undefined under `/WX`).
> 
> `libssh2_setup.h` only maps `snprintf` to `_snprintf` when `LIBSSH2_TESTS` is defined. `test_simple` included `libssh2_priv.h` without that define (unlike `session_fixture.h`).
> 
> Fixed in `d9776561` by adding `#define LIBSSH2_TESTS` before the priv header include.

Sounds reasonable. I wonder if we should set this macro via the builds. I'll have a look.

tests/test_simple.c
```diff
@@ -124,6 +128,73 @@ static int test_ssh2_dh_validate(void)
...
+    if(prc != 1) {
+        fprintf(stderr, "scp_parse incomplete size: prc=%d (want 1)\n", prc);
+        err++;
+    }
+
+    /*
+     * Fixed buffer full of "Cmode size " + long name without newline. Mode and
+     * size must still parse so scp_recv can drain the unused basename.
+     */
+    prefix_len = (size_t)snprintf((char *)long_line, sizeof(long_line),
```
vszakats: Possibly acedemic, but other calls make sure to verify
the return value is non-negative and smaller than the
buffer size, and bail out otherwise.

tests/test_simple.c
```diff
@@ -124,6 +128,73 @@ static int test_ssh2_dh_validate(void)
...
+                                  "C0644 99 ");
+    memset(long_line + prefix_len, 'a',
+           sizeof(long_line) - prefix_len);
```
vszakats:
```suggestion
    memset(long_line + prefix_len, 'a', sizeof(long_line) - prefix_len);
```
unfold line?

src/libssh2_priv.h
```diff
@@ -671,6 +671,10 @@ struct _LIBSSH2_PUBLICKEY {
 
 #define SSH2_SCP_RESPONSE_BUFLEN     256
 
+/* Parse SCP "C" response mode/size; used by scp_recv and unit tests. */
+int ssh2_scp_parse_c_fields(const unsigned char *buf, size_t len,
+                            long *mode_out, libssh2_int64_t *size_out);
```
vszakats:
```suggestion
int ssh2_scp_parse_c_fields(const char *buf, size_t len,
                            long *mode_out, libssh2_int64_t *size_out);
```
would it make the code shorter with fewer casts if using `const char *` buffer here?
(even if the buff may contain non-ASCII, though at a quick look it's an ASCII string.)

tests/test_simple.c
```diff
@@ -31,9 +31,11 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+#define LIBSSH2_TESTS
```
vszakats:
```suggestion
```
#define no longer needed once rebased on 76d30e5d984cdc88ed802aafc12a34132f370ed8 #2417.

tests/test_simple.c
SebTardif: Done: check `prefix_len < 0 || (size_t)prefix_len >= sizeof(long_line)` and bail, same pattern as `session_fixture.c`.

tests/test_simple.c
SebTardif: Done: single-line `memset`.

src/libssh2_priv.h
SebTardif: Switched the helper to `const char *`. Test string literals no longer need casts; `scp_recv` casts the session unsigned buffer at the two call sites.

tests/test_simple.c
SebTardif: Rebased onto current master (includes #2417) and dropped the local `#define LIBSSH2_TESTS`.

SebTardif:
Addressed the open review notes and rebased onto current `master` (includes #2417 / #2418):

- `ssh2_scp_parse_c_fields` now takes `const char *`
- `snprintf` return checked before the long-name fill
- single-line `memset`
- dropped local `LIBSSH2_TESTS` define

Tip: `7d75c1f4`

vszakats:
Thank you, merged now!

PR #2410, opened 2026-07-23
crypt: fix rare unused argument warnings

vszakats:


vszakats:
Also skipping this one.

PR #2412, opened 2026-07-23
mac: fix potential key memleak on rekey

vszakats:
When switching from a non-AEAD mac to an AEAD one.

Reported by Copilot
Bug: https://github.com/libssh2/libssh2/pull/2407#discussion_r3638646061

---

- [ ] consider replacing init/dtor with static functions, current code
  can call them for a different mac's data structure, making the
  current system misleading and confusing: all init functions must
  always be compatible with all dtor functions to avoid issues.
  → The transfer vs. don't transfer key buffer ownership is what
  the init callback decides. Could be replaced by a mac_method
  flag.

### DIFF
--- a/src/mac.c
+++ b/src/mac.c
@@ -74,7 +74,7 @@ static int mac_method_common_init(LIBSSH2_SESSION *session, unsigned char *key,
                                   int *free_key, void **abstract)
 {
     *abstract = key;
-    *free_key = 0;
+    *free_key = 0;  /* mac dtor must free */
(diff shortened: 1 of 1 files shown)


vszakats:
> ## Pull request overview
> This PR addresses a potential MAC key memory leak during rekey on macOS by ensuring the AES-GCM “integrated MAC” stub properly resets and cleans up the MAC abstract pointer across rekeys.

Right, on macOS!

/shakes head

PR #2414, opened 2026-07-23
agent: fix potential infinite loop while receiving data from SSH agent

vszakats:
If the local SSH agent terminates while we are receiving data from it.

Refs:
https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-readfile#pipes
https://linux.die.net/man/2/recv
https://www.man7.org/linux/man-pages/man2/recv.2.html

Reported-by: afldl on github

---

- [x] I assume the patch is broken for `send()`/`WriteFile()`, need fix. [SKIP, can't confirm if it's necessary]
  https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-writefile#pipes

### DIFF
--- a/src/agent.c
+++ b/src/agent.c
@@ -618,8 +618,8 @@ static int agent_connect_unix(LIBSSH2_AGENT *agent)
             rc = (func)(socket,                                           \
                         (char *)(buffer) + finished, (length) - finished, \
                         flags, abstract);                                 \
-            if(rc < 0)                                                    \
-                return rc;                                                \
+            if(rc <= 0)                                                   \
+                return rc ? rc : LIBSSH2_ERROR_SOCKET_DISCONNECT;         \
                                                                           \
             finished += rc;                                               \
         }                                                                 \
--- a/src/agent.c
+++ b/src/agent.c
@@ -426,11 +426,14 @@ cleanup:
     int rc;                                                            \
                                                                        \
     while(*(total) < (length)) {                                       \
-        if(!(agent)->pending_io)                                       \
+        if(!(agent)->pending_io) {                                     \
             ret = (func)((agent)->pipe, (char *)(buffer) + *(total),   \
                          (DWORD)((length) - *(total)),                 \
                          &bytes_transferred,                           \
                          &(agent)->overlapped);                        \
+            if(ret && !bytes_transferred)                              \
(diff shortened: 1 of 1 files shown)


vszakats:
> ## Pull request overview
> Copilot reviewed 1 out of 1 changed files in this pull request and generated no new comments.
> 
> Comments suppressed due to low confidence (1)
> **src/agent.c:443**
> 
> * `bytes_transferred` is added to `*total` even when `ReadFile`/`GetOverlappedResult` returns FALSE. Per WinAPI, the output byte count is only defined on success, so this can read an undefined value on failure paths (e.g., ERROR_IO_PENDING/ERROR_IO_INCOMPLETE). Initializing `bytes_transferred` to 0 at the start of each loop iteration avoids potential UB and keeps the EAGAIN logic correct.
> 
> ```
>                                       &(agent)->overlapped,            \
>                                       &bytes_transferred, FALSE);      \
>                                                                        \
>         *(total) += bytes_transferred;                                 \
>         if(!ret) {                                                     \
> ```

This seems bogus. Per MS documentation, ReadFile() starts by setting this value to zero,
for the other, there are two conditions where it may not be returned, and none of those
apply to this use-case.

"ReadFile sets this value to zero before doing any work or error checking."
https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-readfile

"A pointer to a variable that receives the number of bytes that were actually transferred by a read or write operation. For a [TransactNamedPipe](https://learn.microsoft.com/en-us/windows/desktop/api/namedpipeapi/nf-namedpipeapi-transactnamedpipe) operation, this is the number of bytes that were read from the pipe. For a [DeviceIoControl](https://learn.microsoft.com/en-us/windows/desktop/api/ioapiset/nf-ioapiset-deviceiocontrol) operation, this is the number of bytes of output data returned by the device driver. For a [ConnectNamedPipe](https://learn.microsoft.com/en-us/windows/desktop/api/namedpipeapi/nf-namedpipeapi-connectnamedpipe) or [WaitCommEvent](https://learn.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-waitcommevent) operation, this value is undefined."
https://learn.microsoft.com/windows/win32/api/ioapiset/nf-ioapiset-getoverlappedresult

PR #2429, opened 2026-07-24
hostkey: replace `strncmp()` with `SSH2_IS_LITERAL()` to harden checks

vszakats:
To match RSA header types and ECDSA curves and domain in KEX responses.
As I understand reading the code, the buffers returned by
`ssh2_get_chars()` are not guaranteed to be null-terminated, and may thus
potentially result in an OOB read when using `strncmp()`.

Also helps eliminating magic numbers from these checks.


vszakats:
After looking at this more, there was no potential issue with
the pre-existing code AFAICT. Yet, no regrets because it
(IMO) improves readability/writability, possibly robustness.

PR #1179, opened 2023-08-25
Properly bounds check packet_authagent_open

MichaelBuckley:
An automated fuzzer found an out of bounds read in packet_authagent_open().

And as a bonus commit, use strlen() for strings in packet.c instead of sizeof().

PR #2432, opened 2026-07-25
packet, sftp: replace `strlen()` on literals with `sizeof()`

vszakats:
This reverts an earlier change doing the opposite. `sizeof()` determines
length at build-time, while `strlen()` does the same at runtime. In case
of literals the results are identical, but the build-time option does it
with zero runtime overhead.

Also:
- fix typo in literal used for determining length. (length was correct.)
  Follow-up to bc4e619e76071393e466c29220fc4ef5764c2820 #752
- move literals to the right-side of comparisons, where missing.

Follow-up to 88c5fd9871bb7e0fe16b43a3a147f8bac93cbd82 #2400
Follow-up to af1b50192398c1fc491479960db3f14f8c5ee683 #2224
Follow-up to 88a960a86b9ffbabd17a4bbcf767e26514c4678e #1179

### DIFF
--- a/src/packet.c
+++ b/src/packet.c
@@ -69,7 +69,7 @@ static SSH2_INLINE int packet_queue_listener(
     int rc;
 
     if(listen_state->state == ssh2_NB_state_idle) {
-        size_t offset = strlen("forwarded-tcpip") + 5;
+        size_t offset = sizeof("forwarded-tcpip") - 1 + 5;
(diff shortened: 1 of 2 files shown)


vszakats:
Another "second look" comment: `strlen("literal")` is optimized
at build-time by most popular compilers. But not all, e.g. MSVC
does it only with `/O2`.

PR #2435, opened 2026-07-26
hostkey: implement ECDSA/RSA certificate hostkey verification

### LINKED ISSUE #2434: hostkey: ECDSA/RSA `*-cert-v01@openssh.com` stubs cause KEX_FAILURE (-5) against cert-hostkey-only servers
## Summary

Six certificate hostkey algorithms (`*-cert-v01@openssh.com` for ECDSA and RSA) are advertised in `KEXINIT` and accepted by `libssh2_session_method_pref()`, but their `init` and `sig_verify` callbacks are `NULL`. When a server offers only a certificate hostkey (e.g. Teleport's proxy, which advertises only `ecdsa-sha2-nistp256-cert-v01@openssh.com`), `kex_agree_hostkey()` rejects the method as unusable and the handshake fails with `LIBSSH2_ERROR_KEX_FAILURE (-5)`.

`ssh-ed25519-cert-v01@openssh.com` had the same class of bug and was fixed in #2064. The ECDSA and RSA cert variants are still stubs on `master`.

## Affected versions

libssh2 1.11.1 and `master` (`0c94b90a`). OpenSSL, mbedTLS, libgcrypt, WinCNG, wolfSSL — all backends, since the stubs live in `src/hostkey.c`.

## Description

The cert method structs declare the algorithm name and `initPEM`/`signv`/`dtor` (used for client-side signing), but leave `init` and `sig_verify` `NULL`:

```c
/* src/hostkey.c:832 (master) */
static const struct hostkey_method hostkey_method_ecdsa_ssh_nistp256_cert = {
    "ecdsa-sha2-nistp256-cert-v01@openssh.com",
    SSH2_SHA256_DIG_LEN,
    NULL,                                  /* init */
    hostkey_method_ssh_ecdsa_initPEM,
    NULL,                                  /* sig_verify */
    hostkey_method_ssh_ecdsa_signv,
    NULL, /* encrypt */
    hostkey_method_ssh_ecdsa_dtor,
};
```

Same shape for `nistp384_cert`, `nistp521_cert`, `ssh_rsa_cert`, `rsa_sha2_256_cert`, `rsa_sha2_512_cert` (6 structs total across `src/hostkey.c:379-408` and `832-860`).

Because the struct is present, `libssh2_session_method_pref(LIBSSH2_METHOD_HOSTKEY, ...)` happily accepts the cert name and it shows up in `KEXINIT`. But during agreement, `kex_agree_hostkey()` requires a non-`NULL` `sig_verify` for any kex that signs the hostkey (curve25519, ECDH, DH all set `KEX_METHOD_FLAG_REQ_SIGN_HOSTKEY`):

```c
/* src/kex.c:3244 (and the default-list path at 3269) */
if((kex_flags & KEX_METHOD_FLAG_REQ_SIGN_HOSTKEY) == 0 ||
   method->sig_verify) {        /* NULL → false */
    session->hostkey = method;
    return 0;
}
/* falls through, no match → returns -1 → KEX_FAILURE (-5) */
```

So the cert method is silently skipped as unusable, and if it was the only overlap with the server's offer, the handshake dies with `-5` — no crash, no NULL deref, just a clean failure.

## Why only the server-hostkey direction is affected

For **client-side user authentication** (the client presenting its own cert), libssh2 passes the cert blob through **opaquely** as the `publickey` field of the `SSH_MSG_USERAUTH_REQUEST` — the *server* parses and validates the cert. libssh2 only needs to locate the matching private-key handler, which it does by stripping the `-cert-v01@openssh.com` suffix from the method name (`ssh2_userauth_plain_method`, `src/userauth.c:1145`). The cert blob itself is never parsed field-by-field.

For **server hostkey verification** (the reverse direction), there is no pass-through: the client must itself extract the inner public key from the cert blob to verify the server's signature over the exchange hash. That extraction was never implemented (except for Ed25519 in #2064). So servers that present cert hostkeys (Teleport's proxy, or any `sshd` with `HostCertificate`) hit the unimplemented path.

This is also why the changelog's "RSA-SHA2 certificate-based authentication" refers to userauth, not hostkey verification.

## Reproduce

Reproduced first-hand against libssh2 `master` (`0c94b90a`, `1.11.2_DEV`), OpenSSL backend, with a local `sshd` offering only a cert hostkey.

**Server** — generate an ECDSA host key, sign it with a CA, and constrain `sshd` to the cert algorithm only:

```sh
ssh-keygen -t ecdsa -b 256  -N '' -f ssh_host_ecdsa_key
ssh-keygen -t ecdsa -b 521  -N '' -f ca_ecdsa
ssh-keygen -I host_cert -h -n host.test -s ca_ecdsa ssh_host_ecdsa_key.pub
```
```sshd_config
HostKey        ssh_host_ecdsa_key
HostCertificate ssh_host_ecdsa_key-cert.pub
HostKeyAlgorithms ecdsa-sha2-nistp256-cert-v01@openssh.com
```

`sshd -d` confirms the server advertises only the cert:
```
debug1: list_hostkey_types: ecdsa-sha2-nistp256-cert-v01@openssh.com [preauth]
debug1: kex: host key algorithm: ecdsa-sha2-nistp256-cert-v01@openssh.com [preauth]
```

**Client** (libssh2):
```c
libssh2_session_method_pref(session, LIBSSH2_METHOD_HOSTKEY,
                            "ecdsa-sha2-nistp256-cert-v01@openssh.com");  /* rc == 0, pref accepted */
rc = libssh2_session_handshake(session, sock);
/* rc == -5 (LIBSSH2_ERROR_KEX_FAILURE) */
```

The system `ssh` client completes KEX against the same server (fails later at userauth), so the server config is valid and the `-5` is in libssh2.

**Controls** — same libssh2 build and client binary, only the server hostkey changes:

| Server offers | libssh2 `handshake` |
|---|---|
| `ecdsa-sha2-nistp256-cert-v01` (cert) | `-5` (KEX_FAILURE) |
| `ecdsa-sha2-nistp256` (plain) | `0` ✅ |
| `ssh-ed25519-cert-v01` (cert, fixed by #2064) | `0` ✅ |

The ed25519-cert-succeeds / ecdsa-cert-fails contrast isolates the bug to the ECDSA/RSA cert `init`/`sig_verify` stubs.

Real-world case: Teleport's proxy with `signature_algorithm_suite: balanced-v1` (the default) advertises only `ecdsa-sha2-nistp256-cert-v01@openssh.com`, so any libssh2 client hits this `-5`.

## Root cause

Per `PROTOCOL.certkeys`, an OpenSSH cert blob is:

```
string    "<algo>-cert-v01@openssh.com"
string    nonce
<inner public key fields>            /* curve + Q for ECDSA; e, n for RSA */
uint64   serial
uint32   type
string   key id
string   valid principals
... (extensions, reserved, signature key, signature)
```

The plain `hostkey_method_ssh_ecdsa_init` (`src/hostkey.c:619`) already parses `curve_name` + `EC_point` and calls `ssh2_ecdsa_curve_name_with_octal_new()`. A cert `init` just needs to additionally skip the cert-type string and the nonce, then reuse that same parsing. The existing `sig_verify` operates on the `abstract` (the inner key ctx), which the cert `init` would now populate — so `sig_verify` can be reused verbatim. This is exactly the pattern #2064 used for Ed25519.

CA signature verification (checking the cert is signed by a trusted CA) is **out of scope** here, matching libssh2's existing plain-hostkey behaviour: libssh2 never did CA trust for plain hostkeys either — app-level `known_hosts` / `libssh2_knownhost_*` / pinning via `libssh2_hostkey_hash()` handles trust. The fix should just extract the inner public key and delegate to the existing verify path.

## Proposed fix

For each of the 6 cert variants, add a cert-aware `init` that:
1. `ssh2_match_string(&buf, "<algo>-cert-v01@openssh.com")`
2. read + skip the nonce (`ssh2_get_string`, min 16 bytes)
3. read the inner public key fields (ECDSA: curve name + point; RSA: `e`, `n`)
4. build the key ctx via the existing `ssh2_ecdsa_curve_name_with_octal_new()` / `ssh2_rsa_new()` — same as the plain `init`
5. skip the rest of the cert (no `ssh2_eob` check — certs carry trailing metadata we don't read, same comment as #2064)
6. store the ctx in `*abstract`

Then wire `init` + the **existing** `sig_verify` into each cert struct (replace the two `NULL`s). No `kex.c` change needed — `kex_agree_hostkey` already gates on `sig_verify != NULL`.

The RSA-SHA1 and RSA-SHA2 cert `init` bodies are identical (only the cert-type string + which `sig_verify`/`signv` is wired differ), so a shared static helper is cleanest. Estimated ~250-300 lines of C total; #2064 was ~58 lines for one variant and is a good template.

I have a Teleport cluster to reproduce against and am happy to work on a PR — wanted to file this first to confirm scope is welcome (one PR for all 6, or split by family).

cad0p:
---

## Summary

Implements certificate hostkey verification for the six ECDSA/RSA `*-cert-v01@openssh.com` variants, completing the work started by #2064 (ed25519-cert). Fixes #2434.

The cert method structs advertised the algorithm names in KEXINIT but had `NULL` `init`/`sig_verify` callbacks, so any server offering only a cert hostkey (sshd with `HostCertificate`, Teleport's proxy) failed with `LIBSSH2_ERROR_KEX_FAILURE (-5)`.

## Root cause

`kex_agree_hostkey()` gates on `sig_verify != NULL` for signing kex methods. The cert structs had `NULL` sig_verify, so the negotiated cert method was silently rejected as unusable.

## The fix

Each cert variant gets a cert-aware `init` that parses the cert-type string and nonce, then reuses the existing plain-key parsing to extract the inner public key (ECDSA: curve + EC point; RSA: e, n). The existing `sig_verify` callbacks operate on the abstract (inner key ctx) and are reused verbatim — no `kex.c` change needed.

Per PROTOCOL.certkeys, the RSA/SHA-2 cert types appear only in KEXINIT and the signature; the on-wire cert blob type is always `ssh-rsa-cert-v01@openssh.com`, so the rsa-sha2-{256,512} inits match against that.

CA signature verification is out of scope (matches libssh2's existing plain-hostkey behaviour — app-level `known_hosts`/pinning handles trust).

## Reproduction / verification

Reproduced first-hand against `master` (`0c94b90a`) before fixing: a local sshd offering only `ecdsa-sha2-nistp256-cert-v01@openssh.com` → libssh2 handshake returns `-5`, while the system `ssh` client completes KEX against the same server.

After the fix, all 7 cert variants negotiate successfully:

| Server offers | Before | After |
|---|---|---|
| `ecdsa-sha2-nistp256-cert-v01` | `-5` | `0` |
| `ecdsa-sha2-nistp384-cert-v01` | `-5` | `0` |
| `ecdsa-sha2-nistp521-cert-v01` | `-5` | `0` |
| `ssh-ed25519-cert-v01` (#2064) | `0` | `0` (no regression) |
| `ssh-rsa-cert-v01` | `-5` | `0` |
| `rsa-sha2-256-cert-v01` | `-5` | `0` |
| `rsa-sha2-512-cert-v01` | `-5` | `0` |

## Tests

Adds test infrastructure (the one gap #2064 had):
- Host certificates generated by `gen_keys.sh` and installed in the docker test server (`Dockerfile` + `sshd_config`)
- `FIXTURE_TEST_HOSTKEY` env support in `session_fixture.c` (mirrors the existing `FIXTURE_TEST_CRYPT`/`FIXTURE_TEST_MAC` pattern)
- `test_hostkey_cert.c` + parameterized runs via `test_hostkey_cert_algos.txt` covering ecdsa/ed25519/rsa cert negotiation
- Existing `test_hostkey`/`test_hostkey_hash` pinned to plain `ssh-ed25519` so they don't accidentally negotiate a cert variant now that the server offers both

Full docker test suite passes (48/48), including the 3 new cert tests.

## Scope decision

Filed #2434 first to confirm scope; this PR implements all six variants in one go since they share the same pattern. Happy to split by family (ECDSA / RSA) if preferred.


cad0p:
Closing — I opened this against the wrong repo by mistake. Will reopen targeting my fork so I can use it as a base for downstream work first. Sorry for the noise.

PR #2443, opened 2026-07-28
kex: disable SHA1-based methods by default

vszakats:
Also:
- RELEASE-NOTES: announce removal after next release.
- allow build-time macro to enable in the meantime.
- keep testing in CI with other legacy options.


willco007:
Can it just be removed from the default key exchange list? Seems disabling the entire code block isn't strictly necessary?

PR #2444, opened 2026-07-28
mac: disable SHA1-based MACs by default

vszakats:
Also:
- RELEASE-NOTES: announce removal after next release.
- allow build-time macro to enable in the meantime.
- skip in tests if disabled.
- keep testing in CI with other legacy options.
- workaround oss-fuzzer hang and timeout after 10 minutes by keeping 
  SHA1-based MACs enabled in the fuzzer build.
  Ref: https://github.com/libssh2/libssh2/pull/2444#issuecomment-5110853610

Follow-up to 99f935358d942283194e15c1ded76a247a3d9087 #2443

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -503,7 +503,7 @@ jobs:
         run: |
           cflags=''
           if [[ ( "${MATRIX_CRYPTO}" = 'Libgcrypt' && "${MATRIX_ZLIB}" = 'OFF' ) || "${MATRIX_TEST}" = 'legacy-options' ]]; then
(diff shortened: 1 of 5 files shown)


vszakats:
Nice, the fuzzer is hanging after this change too.
I'm at loss why.

Same effect as seen here: https://github.com/libssh2/libssh2/pull/2199

vszakats:
Work around oss-fuzz hang/timeout by keeping the insecure MACs enabled in that build.

It's suspicious that it's now the second hang caused by touching the elements of algo
lists. Is there maybe an issue in the `kex_agree_*()` functions, or is there perhaps
something wired up in the fuzzer input set that interacts badly with libssh2 changes?

PR #2448, opened 2026-07-29
GHA/cifuzz: pin oss-fuzz actions to hash

vszakats:
---

https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/actions/build_fuzzers/action.yml
https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/actions/run_fuzzers/action.yml
https://github.com/google/oss-fuzz/blob/master/infra/run_fuzzers.Dockerfile
https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/requirements.txt

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -29,13 +29,13 @@ jobs:
     steps:
     - name: 'Build Fuzzers'
       id: build
-      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master  # zizmor: ignore[unpinned-uses]
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@85f0ef392a3a051929176c4ceb834298a152c799
       with:
         oss-fuzz-project-name: 'libssh2'
         dry-run: false
         language: c
     - name: 'Run Fuzzers'
-      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master  # zizmor: ignore[unpinned-uses]
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@85f0ef392a3a051929176c4ceb834298a152c799
       with:
         oss-fuzz-project-name: 'libssh2'
         fuzz-seconds: 600


vszakats:
Well, another zizmor warning to silence:
```
help[stale-action-refs]: commit hash does not point to a Git tag
  --> .github/workflows/cifuzz.yml:32:13
   |
32 |       uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@85f0ef392a3a051929176c4ceb834298a152c799
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this step
   |
   = note: audit confidence → High
   = help: audit documentation → https://docs.zizmor.sh/audits/#stale-action-refs

help[stale-action-refs]: commit hash does not point to a Git tag
  --> .github/workflows/cifuzz.yml:38:13
   |
38 |       uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@85f0ef392a3a051929176c4ceb834298a152c799
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this step
   |
   = note: audit confidence → High
   = help: audit documentation → https://docs.zizmor.sh/audits/#stale-action-refs
```

Why this warning for a target repo not having tags? It's also a fairly well-known repo with this property.

This incentivizes keeping the yolo reference, because shorter, needs
no maintenance (manual bumps!) and needs same amount of zizmor
silencing.

On the good side: pinning to this hash works. The repo is super busy,
but the files actually used in the Action change rarely. (but also silently, so
manually bumping is annoying/error prone.)

vszakats:
Closing without merges, as an experiment with mixed results.

edit: Probably a solution could be if the oss-fuzz repo started tagging after changes involving the actions and their dependencies. Not sure whom to nudge for that to happen.

PR #2450, opened 2026-07-29
packet: fix use-after-free of ignored extended data on eagain

abdul-khaliq-khalid:
the extended-data ignore path in ssh2_packet_add frees data before the ssh2_channel_receive_window_adjust jump point, but data aliases session->packet.payload and the plain free doesn't clear that alias. when the window adjust returns eagain the packAdd_state is left at jump1, so the transport eagain guard that only nulls p->payload on the idle state is skipped and the alias is left dangling: the retry re-reads data[0] from the freed buffer and a session teardown while suspended frees the same buffer again. defer the free to after the eagain check so the buffer stays live across re-entry and gets nulled by the transport layer on the success return. reachable with LIBSSH2_CHANNEL_EXTENDED_DATA_IGNORE set when the server sends extended data and the window-adjust send would block.

### DIFF
--- a/src/packet.c
+++ b/src/packet.c
@@ -950,8 +950,6 @@ ssh2_packet_add_jump_point5:
                 LIBSSH2_CHANNEL_EXTENDED_DATA_IGNORE) &&
                (msg == SSH_MSG_CHANNEL_EXTENDED_DATA)) {
                 /* Pretend we did not receive this */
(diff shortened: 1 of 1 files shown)


vszakats (approved):

vszakats:
Thank you!

PR #710, opened 2022-06-14
Support RSA certificate authentication

msabansal:
This PR is based of https://github.com/libssh2/libssh2/pull/570/files and just the tests were fixed to make sure everything works smoothly.

PR #2474, opened 2026-08-01
tests: regenerate `id_rsa_signed.pub` to match intent, fix `sshd_config` to accept it

vszakats:
Explicitly use `-t ssh-rsa` (signature algorithm) to generate
`id_rsa_signed.pub` to match the purpose of the patch it came with. Also
to differenciate this key (and test) from the subsequently merged
`id_rsa_sha2_512_signed-cert.pub`, which used the exact same keys (`-t
rsa-sha2-512`, the default) before this patch, and made these tests
duplicates.

Also:
- update `sshd_config` to actually allow using an obsolete/insecure
  certificate signature `ssh-rsa`, by adding it to `CASignatureAlgorithms`.
- delete recently added entry from `HostKeyAlgorithms`, it was not
  necessary.
  Follow-up to 5b494b04b1617c9efab569a40c288288d97ba314 #2478

Testing obsolete/insecure algs is not time best spent in CI and for
maintainers, so maybe in a next step it'd useful to disable the RSA
SHA1-based ones by default.

Ref: https://superuser.com/questions/1556852/how-to-check-if-your-ssh-keys-are-in-the-ssh-rsa2-format
Ref: 7392dc95dca4e6d617846b17059ab096f1d9c771 #2479
Follow-up to 4b21e49d9d2db74579b18804ed1f5eeb16578b2f #710

---

Trying to untangle this, some info:
`id_rsa_signed-cert.pub` was `rsa-sha2-512` without the original patch mentioning this, and the `test_auth_pubkey_ok_rsa_signed` tests breaks when switching it to `ssh-rsa`.
`id_rsa_sha2_256_signed-cert.pub` is also sha2, but contrary to its name, it is
`rsa-sha2-512`. (`rsa-sha2-512` is the default in ssh-keygen).
This is visible in the base64-decoded content of these files.


vszakats:
Slightly related observation: it'd be immensly helpful if the test log
included the sshd log output for failed tests. perhaps also successful
ones, with an option.

Another observation: the cert-based keys have the username `libssh2`
hardcoded, which effectively prevents testing them outside of the
Docker env. It'd probably help testing if the hard-wired usernames
could be minimized, in this case perhaps by always generating these
files afresh, with the actual username used. Another option would be
if sshd could somehow be configured to accept a test username/pw
combo (this doesn't seem possible without patching OpenSSH and
building from source.)

PR #2479, opened 2026-08-02
tests: rename a test and keys to match the actual format

vszakats:
The committed key was in reality `rsa-sha2-512`, not `rsa-sha2-256`, as
the filenames suggested.

Also set the type explicitly to `rsa-sha2-512`, to avoid relying on
ssh-keygen's default (`rsa-sha2-512`).

```shell
$ grep -E -o ' .+ ' id_rsa_sha2_512_signed-cert.pub | base64 -d | strings | grep -i rsa
```

Bug: #2474
Follow-up to 3a6ab70dcfeb87a3acbb5af7ea6fc783c3981e04 #1314


vszakats:
Thought: storing keys in repo looks more and more like a bad idea
to me, because of issues like the one fixed in this PR, and others when:
1. someone other than the original commiter needs to reverse-engineer
  the command necessary to generate the key committed as-is.
2. the committed key wasn't in the format it was intented to be,
  only to find out about it later, and requiring a non-trivial to fix.
3. the generator scripts (we now use only as a sort of documentation)
  gets out of sync with the key actually stored in repo.
4. it results in opaque binary blobs in the repo, which is strongly
  not desired since the xz hack attempt.

Dynamically generating the key is possible though it needs to
specify key parameters explicitly to avoid regressing when
`ssh-keygen` changes defaults depending on version. It also
makes it impossible to generate too old or too modern keys
depending on the ssh-keygen version used at generation-time.
E.g. generating `ssh-dss` fails with modern OpenSSH versions.
(though in this case maybe libssh2 should not really deal with
that particular insecure/obsolete key type anymore.)

Another technical issue is that the generator script needs to be
rewritten in Perl to avoid POSIX shell as a dependency for Docker
tests when using CMake.

PR #2500, opened 2026-08-03
tests: add PKCS8-formatted plain and encrypted keys and tests

vszakats:
They are supported via the backends, except OS/400 which has its own
parser, leaving libgcrypt and WinCNG (untested) without support. With
mbedTLS PKCS8 encryption (aes-128-cbc) requires 3.6.0+.

Add RSA, ECDSA and ED25519 keys.

What I find interesting in PKCS8 private keys, is the easy way to 
inspect the ASN.1 content (`openssl asn1parse -dump -in <fn>`). I could
not find similar tooling for PEM and OpenSSH files.

Also:
- tests/openssh_server/authorized_keys: sync up comments with private
  keys and the generator script.

Ref: https://github.com/Mbed-TLS/mbedtls/releases/tag/v3.6.0

---

https://github.com/libssh2/libssh2/pull/2500/files?w=1


vszakats:
mbedTLS fails in CI for encrypted PKCS8 keys, while they seem to work fine locally (using Homebrew mbedtls@3).

vszakats:
After tracing it turns out mbedTLS 3.6.x (3.6.6) supports it fine,
but earlier versions return `MBEDTLS_ERR_PKCS5_FEATURE_UNAVAILABLE`,
from the `mbedtls_pk_parse_key*()` in `ssh2_rsa_new_priv()`.

I guess this is "Add support for using AES-CBC 128, 192, and 256 bit schemes
with PKCS#5 PBES2. Keys encrypted this way can now be parsed by PK parse."
https://github.com/Mbed-TLS/mbedtls/releases/tag/v3.6.0

PR #2509, opened 2026-08-04
ossfuzz: test #1953 on master

vszakats:
Also:
- tidy up code and LLM comments.


vszakats:
the hang again... nice.

what may be the reason of this?

edit: in ssh2_knownhost_fuzzer.

vszakats:
After making the knownhost fuzzer a no-op, it still hangs the same way:
```
2026-08-05 12:09:41,178 - root - WARNING - Failed to download corpus for ssh2_knownhost_fuzzer.
2026-08-05 12:09:41,178 - root - INFO - Starting fuzzing
```

```c
#include "libssh2.h"

extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
{
    return 0;
}
```

PR #2514, opened 2026-08-08
session_fixture: try simplify skipping non-supported algs

vszakats:
Spotted by Copilot
Ref: https://github.com/libssh2/libssh2/pull/2513#discussion_r3740394757


vszakats:
Problematic because:
- each skipped tests need initialization and thus non-zero time (usually 1s per test on Linux).
- maybe it's better to keep manually excluding to avoid surprises or `libssh2_session_method_pref()` bugs.

Not merging.

PR #2517, opened 2026-08-11
session: free keyboard-interactive state

hunterinvariants:
Free keyboard-interactive authentication state when a session is destroyed.

This covers the authentication name, prompt and response arrays, and their individual strings. Avoid calling memcpy() with null pointers for zero-length test input.

Tested with the complete CMake/OpenSSL test suite and the targeted ASan/UBSan test.

### DIFF
--- a/tests/test_auth_keyboard_info_request.c
+++ b/tests/test_auth_keyboard_info_request.c
@@ -253,9 +253,11 @@ static int test_case(int num,
         return 1;
     }
 
-    session->userauth_kybd_data = SSH2_ALLOC(session, data_len);
-    session->userauth_kybd_data_len = data_len;
-    memcpy(session->userauth_kybd_data, data, data_len);
+    if(data_len) {
+        session->userauth_kybd_data = SSH2_ALLOC(session, data_len);
+        session->userauth_kybd_data_len = data_len;
+        memcpy(session->userauth_kybd_data, data, data_len);
+    }
 
     rc = userauth_keyboard_interactive_decode_info_request(session);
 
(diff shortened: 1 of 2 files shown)


tests/test_auth_keyboard_info_request.c
```diff
@@ -255,7 +255,8 @@ static int test_case(int num,
 
     session->userauth_kybd_data = SSH2_ALLOC(session, data_len);
     session->userauth_kybd_data_len = data_len;
-    memcpy(session->userauth_kybd_data, data, data_len);
+    if(data_len)
+        memcpy(session->userauth_kybd_data, data, data_len);
```
vszakats, line 259: Would it make sense to guard all 3 lines for `data_len` (>0), to avoid
allocating zero-length buffer here?

hunterinvariants: Yes, that avoids the zero-length allocation and keeps the initialized session state unchanged. Updated accordingly.

hunterinvariants:
Two unrelated CI jobs failed after the review update:

- Ubuntu 26.04 completed the build, and `test_auth_keyboard_info_request` passed all 18 cases. Two integration tests failed earlier in the connection setup while verifying the server hostkey signature for `mlkem768x25519`, before reaching the public-key authentication exercised by those tests.
- MidnightBSD could not download prerequisite packages and then failed because CMake was unavailable.

Could someone with workflow permissions rerun these two jobs?

vszakats:
Thanks, merged now.

(Those were just flaky tests, we can ignore them for now.)

PR #2521, opened 2026-08-11
openssl: guard zero-length input in ossl_swap_bytes

abdul-khaliq-khalid:
ossl_swap_bytes() sets j = len - 1 and loops while i < j, so a zero length wraps j to SIZE_MAX and the in-place byte reversal runs off the heap. On OpenSSL 3 builds that len can be zero: ossl_ecdsa_openssh_priv_to_pubkey() takes the private scalar from ssh2_get_bignum_bytes(), which returns a zero length for an empty or all-zero mpint in an OpenSSH key blob, and unlike the ssh2_rsa_new()/ssh2_dsa_new() call sites this one has no length guard. Return early when len < 2, the same way wcng_reverse_bytes() does in the WinCNG backend.


vszakats:
Thank you!

PR #2534, opened 2026-08-24
libgcrypt: guard zero-length datum in `ssh2_dsa_sha1_sign()`

abdul-khaliq-khalid:
ssh2_dsa_sha1_sign strips a leading zero from the r and s components returned by gcry_sexp_nth_data before it checks their length, so a zero-length datum makes tmp[0] a one-byte out-of-bounds read and the following size-- wrap to SIZE_MAX. the later size > 20 test does catch the wrapped value before the memcpy, so nothing is written out of bounds, but the stray read is still there. the sibling strippers already avoid this: ssh2_rsa_sha2_sign in this same file guards with `if(size && s[0] == '\0')` and the wincng backend does the same for both halves, so this brings the dsa path in line by testing size first. valid signatures fall in the 1..20 range and are unchanged.

### DIFF
--- a/src/libgcrypt.c
+++ b/src/libgcrypt.c
@@ -532,7 +532,7 @@ int ssh2_dsa_sha1_sign(ssh2_dsa_ctx *dsa,
     if(!tmp)
         goto err;
 
-    if(tmp[0] == '\0') {
+    if(size && tmp[0] == '\0') {
         tmp++;
         size--;
     }
@@ -554,7 +554,7 @@ int ssh2_dsa_sha1_sign(ssh2_dsa_ctx *dsa,
     if(!tmp)
         goto err;
 
-    if(tmp[0] == '\0') {
+    if(size && tmp[0] == '\0') {
         tmp++;
         size--;
     }


willco007:
This seems fine; but could you actually get a zero length reply from `gcry_sexp_nth_data()` that also returned a valid pointer or is this simply theoretical?

vszakats:
> This seems fine; but could you actually get a zero length reply from `gcry_sexp_nth_data()` that also returned a valid pointer or is this simply theoretical?

Question also applies to this pattern in pre-existing `ssh2_rsa_sha2_sign()` (it already has the size check).

vszakats (approved):

vszakats:
According to libgcrypt sources (between the oldest supported version 1.8
and Git master) a non-NULL value is only returned when size is also
non-zero. If a NULL is returned, the returned size is always zero. This may
not have been the case before libgcrypt 1.1.2, in release 1.1.0 from
1999-10-26 and earlier.

Documentation doesn't explicitly say, but I think it's fair to expect it
to continue behaving this way.

With 1.8 as the minimum version, checking both ptr and size is redundant,
so perhaps the dupe check could be deleted, instead of adding extra checks?

Refs:
https://github.com/gpg/libgcrypt/commit/2a6d975764aa898d7ab2023d1642f6743ea8aa0f
https://github.com/gpg/libgcrypt/blob/9fd5ecd9abe3606e96fb85981ad386f6ea8d0d0a/src/sexp.c#L700-L756
https://www.gnupg.org/documentation/manuals/gcrypt-devel/Working-with-S_002dexpressions.html#index-gcry_005fsexp_005fnth_005fdata-226

vszakats:
I think this is fine to make it clear in the code that `size` is non-zero.
Freeing the pointer in that case feels overkill, but if a human or machine
takes issue, it can be tackled for good measure. Not a practical problem
though.

Thanks @abdul-khaliq-khalid, merged now.

PR #2537, opened 2026-08-26
sftp: free response buffer on read protocol-error paths

abdul-khaliq-khalid:
sftp_read takes ownership of the FXP_DATA response buffer that sftp_packet_requirev unlinks from the packet brigade, but three protocol-error branches in the SSH_FXP_DATA arm return without freeing it: an unexpected chunk offset, a declared data length past the end of the packet body, and a chunk larger than the one we requested. A server can trigger any of these on a read and leak the whole response buffer each time it happens. Free data before those returns, the way the neighbouring SSH_FXP_STATUS arm already does.

### DIFF
--- a/src/sftp.c
+++ b/src/sftp.c
@@ -1655,26 +1655,32 @@ static ssize_t sftp_read(LIBSSH2_SFTP_HANDLE *handle,
                 }
 
             case SSH_FXP_DATA:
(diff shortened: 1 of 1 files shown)


vszakats (approved):

vszakats:
Thank you!

PR #2539, opened 2026-08-27
hostkey: implement ECDSA/RSA certificate hostkey verification

### LINKED ISSUE #2434: hostkey: ECDSA/RSA `*-cert-v01@openssh.com` stubs cause KEX_FAILURE (-5) against cert-hostkey-only servers

cad0p:
## Summary

Implement server hostkey verification for the six ECDSA/RSA OpenSSH certificate hostkey variants:

- `ecdsa-sha2-nistp256-cert-v01@openssh.com`
- `ecdsa-sha2-nistp384-cert-v01@openssh.com`
- `ecdsa-sha2-nistp521-cert-v01@openssh.com`
- `ssh-rsa-cert-v01@openssh.com`
- `rsa-sha2-256-cert-v01@openssh.com`
- `rsa-sha2-512-cert-v01@openssh.com`

This fixes [libssh2/libssh2#2434](https://github.com/libssh2/libssh2/issues/2434), where these algorithms were advertised and accepted as preferences but had `NULL` hostkey `init`/`sig_verify` callbacks. A certificate-only hostkey offer therefore caused `LIBSSH2_ERROR_KEX_FAILURE (-5)`.

## Implementation

- ECDSA certificate parsing uses one shared init path for all three curves. It validates the certificate type, nonce, inner curve name, and EC point before constructing the verification context.
- RSA certificate parsing uses one shared init path for all three RSA signature variants.
- Per-algorithm RSA `sig_verify` callbacks remain distinct.
- Per `PROTOCOL.certkeys`, the RSA/SHA-2 certificate blob itself carries the type string `ssh-rsa-cert-v01@openssh.com`; the negotiated SHA-2 algorithm is represented by KEXINIT and the exchange signature.
- Certificate CA trust validation remains out of scope, consistent with existing plain-hostkey handling.

## Tests

- Host CA fixtures (`keys/ca_host_ecdsa`, `keys/ca_host_rsa`) and CA-signed host certificates for the ECDSA P-256 and RSA host keys in the Docker OpenSSH fixture.
- The certificate algorithms are exercised through `test_read_algos.txt` and `FIXTURE_TEST_HOSTKEY`: `ecdsa-sha2-nistp256-cert-v01@openssh.com`, `rsa-sha2-256-cert-v01@openssh.com`, `rsa-sha2-512-cert-v01@openssh.com`, `ssh-rsa-cert-v01@openssh.com`, plus the pre-existing `ssh-ed25519-cert-v01@openssh.com`.
- The ECDSA P-384/P-521 certificate variants share the same codepath and are covered by it; only P-256 is tested directly to keep the fixture matrix lean (per maintainer review feedback).
- Disabled-backend skip entries in `tests/session_fixture.c` for the new certificate algorithms.
- Preserved the existing ed25519 certificate test path.

## Verification

- Fork CI: all checks pass, including the Docker-backed algorithm tests, clang-tidy, Autotools `distcheck`, CodeQL, and fuzzing.
- `checksrc-all.pl`, `spacecheck.pl`, `reuse lint` pass locally.

Fixes #2434
Follow-up to 6036e577206266ca19950f939d3e1186b6d5cd80 #2513
Follow-up to 65a276e8fe77868d212fe814213fca6f438e9c0a #2064


cad0p:
CI note: the 4 failing jobs are all infrastructure flakes, not code issues:

1. **midnightbsd 4.0.4** — package mirror timeout (`midnightbsd.org` fetch of `rhash`/`mpdecimal` timed out) → `cmake: not found` → configure failed (exit 127).
2. **linux (clang, amd64, OpenSSL, AM, OFF)** — Docker container startup race: `test_read-ssh-ed25519-cert-v01@openssh.com` failed with `-43 Failed getting banner` (container not yet listening when the test connected). All certificate tests introduced by this PR passed.
3. **linux (clang-tidy, amd64, OpenSSL, CM, ON)** — `ghcr.io/libssh2/ci_tests_openssh_server:349ab27d7c152c81fd58: not found` — the CI image tag is missing from GHCR, so every Docker test falls back to building a fresh container; `test_read-aes256-ctr` hit the startup race. clang-tidy itself reported zero warnings.
4. **linux (gcc, amd64, OpenSSL-3-no-deprecated, CM, ON, ubuntu-26.04)** — same missing GHCR image tag + container startup race.

All new certificate tests passed (`ecdsa-sha2-nistp256-cert-v01@openssh.com`, `rsa-sha2-256-cert-v01@openssh.com`, `rsa-sha2-512-cert-v01@openssh.com`, `ssh-rsa-cert-v01@openssh.com`), and CodeQL, CIFuzz, and all 6 AppVeyor jobs passed. A retry of the failed jobs should be enough.

tests/openssh_server/ssh_host_ed25519_key-cert.pub
```diff
@@ -1 +1 @@
-ssh-ed25519-cert-v01@openssh.com AAAAIHNzaC1lZDI1NTE5LWNlcnQtdjAxQG9wZW5zc2guY29tAAAAING+WjIhgorwyJHfxYyRjA/I7C1wCeRfO6iZ/ea3wJIXAAAAIO7Dhx/ox7Xoi/sg2GrR7j2x5NxvTGKnMH3wifd6UPCxAAAAAAAAAAAAAAACAAAADGhvc3RfZWQyNTUxOQAAAAsAAAAHbGlic3NoMgAAAAAAAAAA//////////8AAAAAAAAAAAAAAAAAAAAzAAAAC3NzaC1lZDI1NTE5AAAAIFA8foTc5+IFCqtcMNCLoh7hrmovve3Y7pRopB8REY3QAAAAUwAAAAtzc2gtZWQyNTUxOQAAAEAxbW++zhv7ENBj74RL/91rq5COmlxsM8MjpQj9LUjUytm/hR4ZZj8Now/PGl1bO62zCTXa6BBC6xkcvaZp9D0H openssh_server/ssh_host_ed25519_key.pub
+ssh-ed25519-cert-v01@openssh.com AAAAIHNzaC1lZDI1NTE5LWNlcnQtdjAxQG9wZW5zc2guY29tAAAAIEaUi28wVfr0XwzJRguDDvQ0tnGRtyODnFSWK2gmUDytAAAAIO7Dhx/ox7Xoi/sg2GrR7j2x5NxvTGKnMH3wifd6UPCxAAAAAAAAAAAAAAACAAAADGhvc3RfZWQyNTUxOQAAAAsAAAAHbGlic3NoMgAAAAAAAAAA//////////8AAAAAAAAAAAAAAAAAAAAzAAAAC3NzaC1lZDI1NTE5AAAAIFA8foTc5+IFCqtcMNCLoh7hrmovve3Y7pRopB8REY3QAAAAUwAAAAtzc2gtZWQyNTUxOQAAAEAveWZ3z3+0rkJJRSGLVQYUOt/SUGV7dxFik/AGKVv8gbrLFzLsvHlHg007EMLrHqPxDuDMuvkxssrK3t3RTjQJ openssh_server/ssh_host_ed25519_key.pub
```
vszakats: Is there a reason to change the pre-existing (also unrelated to this PR) ed25519 key?
If not, I suggest keeping the original.

cad0p: Good catch — that was an unintended regeneration from our keys-generate run. Restored the original committed file in c9bbe29f; the PR now leaves the pre-existing ed25519 fixture untouched.

tests/openssh_server/Dockerfile
```diff
@@ -43,7 +49,12 @@ RUN sed -i 's/KbdInteractiveAuthentication no/KbdInteractiveAuthentication yes/'
 COPY sshd_config /tmp/etc/ssh/sshd_config
 RUN cat /tmp/etc/ssh/sshd_config >> /etc/ssh/sshd_config
 RUN echo 'TrustedUserCAKeys /etc/ssh/ca_user_keys.pub' >> /etc/ssh/sshd_config
+RUN echo 'HostKey /etc/ssh/ssh_host_rsa_key' >> /etc/ssh/sshd_config
+RUN echo 'HostKey /etc/ssh/ssh_host_ecdsa_key' >> /etc/ssh/sshd_config
+RUN echo 'HostKey /etc/ssh/ssh_host_ed25519_key' >> /etc/ssh/sshd_config
```
vszakats: According to `man sshd_config`, the values set here are the defaults, thus redundant,
is there a strong reason to add them via this PR in particular?

tests/session_fixture.c
```diff
@@ -210,9 +220,9 @@ LIBSSH2_SESSION *start_session_fixture(int *skipped, int *err)
     if(!hostkey)
         hostkey =
             "ssh-ed25519,"
-            "ecdsa-sha2-nistp521,"
-            "ecdsa-sha2-nistp384,"
             "ecdsa-sha2-nistp256,"
+            "ecdsa-sha2-nistp384,"
+            "ecdsa-sha2-nistp521,"
```
vszakats, line 223: What is the reason for reordering the list here? Is it necessary for this PR?

src/hostkey.c
```diff
@@ -357,13 +357,82 @@ static const struct hostkey_method hostkey_method_ssh_rsa_sha2_512 = {
...
+ * signature, so all three cert variants share this blob layout and init.
+ * The existing per-variant sig_verify operates on the abstract (the inner
+ * RSA key ctx) and is reused verbatim.
+ */
+static int hostkey_method_ssh_rsa_init_cert(
+    LIBSSH2_SESSION *session,
+    const unsigned char *hostkey_data,
+    size_t hostkey_data_len,
+    void **abstract)
+{
```
vszakats:
```suggestion
static int hostkey_method_ssh_rsa_init_cert(LIBSSH2_SESSION *session,
                                            const unsigned char *hostkey_data,
                                            size_t hostkey_data_len,
                                            void **abstract)
{
```
formatting nit

tests/keys-generate.sh
```diff
@@ -37,6 +39,8 @@ pr="${1:-libssh2}"
 
 # host certificates
 ssh-keygen                 -h -I 'host_ed25519' -n "${pr}" -s 'keys/ca_host_ed25519'         'openssh_server/ssh_host_ed25519_key.pub'
+ssh-keygen                 -h -I 'host_ecdsa'       -n "${pr}" -s 'keys/ca_host_ecdsa'           'openssh_server/ssh_host_ecdsa_key.pub'
+ssh-keygen                 -h -I 'host_rsa'         -n "${pr}" -s 'keys/ca_host_rsa'             'openssh_server/ssh_host_rsa_key.pub'
```
vszakats:
```suggestion
ssh-keygen                 -h -I 'host_rsa'     -n "${pr}" -s 'keys/ca_host_rsa'             'openssh_server/ssh_host_rsa_key.pub'
ssh-keygen                 -h -I 'host_ecdsa'   -n "${pr}" -s 'keys/ca_host_ecdsa'           'openssh_server/ssh_host_ecdsa_key.pub'
ssh-keygen                 -h -I 'host_ed25519' -n "${pr}" -s 'keys/ca_host_ed25519'         'openssh_server/ssh_host_ed25519_key.pub'
```
let's align vertically, and adjust order with rest of key groups.

tests/keys-generate.sh
```diff
@@ -25,6 +25,8 @@ rm -f keys/id_*
 
 # host CAs
 ssh-keygen -t ed25519         -N ''          -m RFC4716 -C 'ca_host_ed25519'              -f 'keys/ca_host_ed25519'
+ssh-keygen -t ecdsa   -b  256 -N ''          -m RFC4716 -C 'ca_host_ecdsa'                -f 'keys/ca_host_ecdsa'
+ssh-keygen -t rsa     -b 2048 -N ''          -m RFC4716 -C 'ca_host_rsa'                  -f 'keys/ca_host_rsa'
```
vszakats, line 27:
```suggestion
ssh-keygen -t rsa     -b 2048 -N ''          -m RFC4716 -C 'ca_host_rsa'                  -f 'keys/ca_host_rsa'
ssh-keygen -t ecdsa   -b  256 -N ''          -m RFC4716 -C 'ca_host_ecdsa'                -f 'keys/ca_host_ecdsa'
ssh-keygen -t ed25519         -N ''          -m RFC4716 -C 'ca_host_ed25519'              -f 'keys/ca_host_ed25519'
```
let's sync order with rest of key groups here also.

tests/openssh_server/Dockerfile
```diff
@@ -30,6 +30,12 @@ RUN install -m 0600 /tmp/etc/ssh/ssh_host_ed25519_key /etc/ssh/ssh_host_ed25519_
 COPY ssh_host_ed25519_key-cert.pub /tmp/etc/ssh/ssh_host_ed25519_key-cert.pub
 RUN install -m 0644 /tmp/etc/ssh/ssh_host_ed25519_key-cert.pub /etc/ssh/ssh_host_ed25519_key-cert.pub
 
+COPY ssh_host_ecdsa_key-cert.pub /tmp/etc/ssh/ssh_host_ecdsa_key-cert.pub
+RUN install -m 0644 /tmp/etc/ssh/ssh_host_ecdsa_key-cert.pub /etc/ssh/ssh_host_ecdsa_key-cert.pub
+
+COPY ssh_host_rsa_key-cert.pub /tmp/etc/ssh/ssh_host_rsa_key-cert.pub
+RUN install -m 0644 /tmp/etc/ssh/ssh_host_rsa_key-cert.pub /etc/ssh/ssh_host_rsa_key-cert.pub
```
vszakats: Let's also reorder these commands to have RSA first, then ECDSA, and then the pre-existing ED25519, to sync with existing order.

tests/openssh_server/Dockerfile
```diff
@@ -43,7 +49,12 @@ RUN sed -i 's/KbdInteractiveAuthentication no/KbdInteractiveAuthentication yes/'
 COPY sshd_config /tmp/etc/ssh/sshd_config
 RUN cat /tmp/etc/ssh/sshd_config >> /etc/ssh/sshd_config
 RUN echo 'TrustedUserCAKeys /etc/ssh/ca_user_keys.pub' >> /etc/ssh/sshd_config
+RUN echo 'HostKey /etc/ssh/ssh_host_rsa_key' >> /etc/ssh/sshd_config
+RUN echo 'HostKey /etc/ssh/ssh_host_ecdsa_key' >> /etc/ssh/sshd_config
+RUN echo 'HostKey /etc/ssh/ssh_host_ed25519_key' >> /etc/ssh/sshd_config
 RUN echo 'HostCertificate /etc/ssh/ssh_host_ed25519_key-cert.pub' >> /etc/ssh/sshd_config
+RUN echo 'HostCertificate /etc/ssh/ssh_host_ecdsa_key-cert.pub' >> /etc/ssh/sshd_config
+RUN echo 'HostCertificate /etc/ssh/ssh_host_rsa_key-cert.pub' >> /etc/ssh/sshd_config
```
vszakats, line 52:
```suggestion
RUN echo 'HostCertificate /etc/ssh/ssh_host_rsa_key-cert.pub' >> /etc/ssh/sshd_config
RUN echo 'HostCertificate /etc/ssh/ssh_host_ecdsa_key-cert.pub' >> /etc/ssh/sshd_config
RUN echo 'HostCertificate /etc/ssh/ssh_host_ed25519_key-cert.pub' >> /etc/ssh/sshd_config
```
and here too

tests/openssh_server/Dockerfile
cad0p: Yes, sshd only fills its default key set when no HostKey is configured: https://github.com/openssh/openssh-portable/blob/master/servconf.c#L311 . Since this PR adds new host keys/certificates to the fixture, once any HostKey is present the implicit defaults (which include the keys we depend on) are replaced. We hit this empirically [in CI](https://github.com/cad0p/libssh2/actions/runs/32396037104): adding the p384/p521 HostKey lines made all other hostkey tests fail until we listed every key explicitly.

tests/session_fixture.c
cad0p: The reorder is needed to keep `test_hostkey` and `test_hostkey_hash` passing once the fixture offers more than one ECDSA key. `kex_agree_hostkey()` walks the client's pref list in order and picks the first key type the server also offers. With the original order (`ecdsa-sha2-nistp521` first) and the p521 host key now present in the fixture, negotiation picked p521, and both tests failed with `Unexpected type of hostkey: 5: ecdsa-sha2-nistp521` — they only assert against the P-256 key (expected blob + `LIBSSH2_HOSTKEY_TYPE_ECDSA_256` + p256 digests). Evidence: [run 32395097557](https://github.com/cad0p/libssh2/actions/runs/32395097557) at `08ca70de`. The p384/p521 entries remain in the list, just after p256.

Note: after the p384/p521 fixture trim (per your review), the server only offers p256 ECDSA again, so the reorder is currently dormant, but keeping p256 first is still the safer default and matches what these tests assert.

tests/openssh_server/Dockerfile
cad0p: Done

tests/openssh_server/Dockerfile
vszakats: I still don't grasp it: that CI run had 5 of the HostKeys, two of them non-defaults:
https://github.com/cad0p/libssh2/blob/0bf9b9abbda5a62e0e1ba4a312c59e6b3cd07e41/tests/openssh_server/Dockerfile#L64-L68
Those two are no longer present in the PR, leaving 3 that are defaults. Why list
them in this scenario?

tests/session_fixture.c
vszakats: If not necessary for the current state of the PR, only for future/theoretical,
I tend to prefer not to touch it as part of this patch. If there ever comes up
a strong reason to test with multiple ECDSA keys, it will be a better time to
do this IMO.

cad0p: Ack, restoring the upstream order (p521, p384, p256)

tests/openssh_server/Dockerfile
cad0p: You're right, sorry about the confusion, removing the other two keys only the default ones remained. I removed the HostKey lines and verified CI green: the defaults are honoured.

vszakats (dismissed):

tests/session_fixture.c
```diff
@@ -99,6 +99,16 @@ static char const *skip_mac[] = {
...
 /* List of HOSTKEY algorithms for which tests are skipped */
 static char const *skip_hostkey[] = {
+#if !LIBSSH2_ECDSA
+    "ecdsa-sha2-nistp256-cert-v01@openssh.com",
+#endif
+#if !LIBSSH2_RSA_SHA2
+    "rsa-sha2-256-cert-v01@openssh.com",
+    "rsa-sha2-512-cert-v01@openssh.com",
+#endif
+#if !LIBSSH2_RSA_SHA1
```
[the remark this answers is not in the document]
vszakats: ~~Small nit, but valid.~~

tests/session_fixture.c
```diff
@@ -99,6 +99,16 @@ static char const *skip_mac[] = {
 
 /* List of HOSTKEY algorithms for which tests are skipped */
 static char const *skip_hostkey[] = {
+#if !LIBSSH2_ECDSA
+    "ecdsa-sha2-nistp256-cert-v01@openssh.com",
+#endif
+#if !LIBSSH2_RSA_SHA2
```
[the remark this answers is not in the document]
vszakats: ~~This one is valid too.~~

src/hostkey.c
```diff
@@ -811,12 +879,105 @@ static const struct hostkey_method hostkey_method_ecdsa_ssh_nistp521 = {
...
+                       "ecdsa-sha2-nistp256-cert-v01@openssh.com"))
+        curve = SSH2_EC_CURVE_NISTP256;
+    else if(SSH2_IS_LITERAL(type_str, type_len,
+                            "ecdsa-sha2-nistp384-cert-v01@openssh.com"))
+        curve = SSH2_EC_CURVE_NISTP384;
+    else if(SSH2_IS_LITERAL(type_str, type_len,
+                            "ecdsa-sha2-nistp521-cert-v01@openssh.com"))
+        curve = SSH2_EC_CURVE_NISTP521;
+    else
+        return -1;
```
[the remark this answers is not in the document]
vszakats: Perhaps valid? I'm personally fine without the regression test case suggested by the LLM.

tests/session_fixture.c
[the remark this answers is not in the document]
vszakats: Not needed after merging 2042863db5a87830781c9ed12d9fec7e9e13444a #2565.

tests/session_fixture.c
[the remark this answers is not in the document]
vszakats: Not needed after merging 2042863db5a87830781c9ed12d9fec7e9e13444a #2565.

src/hostkey.c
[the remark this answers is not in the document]
cad0p: Reproduced with a PoC driving the actual `hostkey_method_ssh_ecdsa_init_cert` / `ssh2_ecdsa_sig_verify` code paths:

- A peer negotiating `ecdsa-sha2-nistp256-cert-v01@openssh.com` can present an **unmodified, validly-signed nistp384 certificate blob**: `init_cert` maps the curve from the blob's own type string, builds a p384 context, and `ssh2_ecdsa_verify` then verifies successfully.
- Fixed by rejecting the blob unless its cert type string equals the negotiated method name (`session->hostkey->name`, which `kex_agree_hostkey()` sets before `init()` runs). With the guard, the forged case is rejected and all genuine cases still verify.
- Two side notes: (a) for RSA, per-method `sig_verify` already pins the hash, so no check is needed there; (b) the pre-existing plain-ECDSA `hostkey_method_ssh_ecdsa_init` has the same blob-driven curve mapping upstream; should be addressed separately.

Pushed the 4-line guard in c755ad9d62e517be90e1dd191a99f8266d29da29 (no regression test per the discussion above).

src/hostkey.c
```diff
@@ -805,12 +873,109 @@ static const struct hostkey_method hostkey_method_ecdsa_ssh_nistp521 = {
...
+    buf.data = SSH2_UNCONST(hostkey_data);
+    buf.dataptr = buf.data;
+    buf.len = hostkey_data_len;
+
+    if(ssh2_get_chars(&buf, &type_str, &type_len))
+        return -1;
+
+    /* cert type string must match the negotiated hostkey method */
+    if(strcmp(type_str, session->hostkey->name))
+        return -1;
```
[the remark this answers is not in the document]
cad0p: Valid: demonstrated with an ASAN harness: the `strcmp()` verdict diverged on field-identical blobs differing only in the bytes *after* the type field, and ASAN flagged a 41-byte read from the 40-byte field (`heap-buffer-overflow` on an exact-size allocation when the field ends the buffer). In the shipped code it stayed in-bounds only thanks to the earlier 145-byte minimum-length check. Replaced with an explicit length + bytes comparison in 0c6e8ff512d211b6e019d71e1413e9d3460628d6, matching the `ssh2_match_string()` pattern used by the other cert inits: the forged-blob PoC still rejects, all genuine cases still verify.

vszakats (approved):

vszakats:
Thank you @cad0p, merged now!

PR #2541, opened 2026-08-27
openssl: bound EC point length in `ssh2_ecdsa_create_key()`

abdul-khaliq-khalid:
The OpenSSL 3 path of ssh2_ecdsa_create_key copies the generated public point into the fixed octal_value[EC_MAX_POINT_LEN] stack buffer, handing the length reported by EVP_PKEY_get_octet_string_param straight back as the destination capacity without checking it against the buffer. Every other point extraction in this file bounds the length against EC_MAX_POINT_LEN first, including the legacy path a few lines below in the same function. Add the same check right after the length is queried so both branches guard the stack write the same way.


vszakats (approved):

vszakats:
Thanks!

PR #2557, opened 2026-08-28
crypto: try dropping unused RSA/DSA private key generation codepaths

vszakats:
Internal crypto API `ssh2_rsa_new()` and `ssh2_dsa_new()` were able to
generate both public and private keys. However, the private key part was
not actually used by any external callers (from `hostkey.c`). It was 
used in backends lacking a native private key loader: libgcrypt for both
RSA and DSA, and WinCNG for DSA [1].

Also add `_pub` suffic to RSA/DSA APIs to reflect their actual purpose.

[1] bcrypt seems to offer support for DSA, but not implemented in
WinCNG, it also makes no sense to try implementing it now that DSA is
dead and WinCNG is untestable in CI.

### DIFF
--- a/src/HACKING-CRYPTO.md
+++ b/src/HACKING-CRYPTO.md
@@ -436,26 +436,14 @@ If defined as 0, the rest of this section can be omitted.
 `ssh2_rsa_ctx`: Type of an RSA computation context. Generally a struct.
 
 ```c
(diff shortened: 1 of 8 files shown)


vszakats:
Turns out there is not much to save here. OpenSSL is also using these internally. Closing without merge.

PR #2572, opened 2026-09-01
GHA/codeql: build C++ (fuzzer) code

vszakats:


vszakats:
~~Built, but not picked up by CodeQL unfortunately.~~

Had to wait a bit, and yes, they were picked up.

PR #2578, opened 2026-09-03
publickey: preserve list state across EAGAIN

ceodaradigu:
`libssh2_publickey_list_fetch()` preserves its packet-receive state when a
nonblocking channel returns `LIBSSH2_ERROR_EAGAIN`, but the accumulated list,
key count and capacity were local to a single invocation.

Keep those aggregate values in the opaque publickey handle while a list fetch
is in progress. Restore them on retry, transfer ownership to the caller on
success, and release them on an error or subsystem shutdown.

Validation included a static CMake build with GCC, the nonblocking regression
harness under ASan/UBSan/LeakSanitizer, a poisoned-allocation retry and cleanup
matrix, the configured standalone upstream test, `scripts/checksrc-all.pl`,
and `git diff --check`.

OpenAI Codex assisted with the analysis, patch preparation and test harness. I
manually reviewed the source diff, the description and the test results.

Fixes GHSA-qwvp-xm4w-32m7

### DIFF
--- a/src/libssh2_priv.h
+++ b/src/libssh2_priv.h
@@ -675,6 +675,9 @@ struct _LIBSSH2_PUBLICKEY {
     unsigned char listFetch_buffer[12];
     unsigned char *listFetch_data;
     size_t listFetch_data_len;
+    libssh2_publickey_list *listFetch_list;
+    unsigned long listFetch_keys;
+    unsigned long listFetch_max_keys;
 };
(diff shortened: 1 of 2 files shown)


vszakats:
Needs a rebase on master after merging #2579.

vszakats:
Thanks @ceodaradigu, merged now.

PR #2583, opened 2026-09-04
wincng: normalize fixed-width DH values before mpint encoding

### LINKED ISSUE #804: WinCNG test failures

Lv2dev:
## Summary

- normalize fixed-width WinCNG bignums before SSH `mpint` serialization
- apply the normalization to DH public values and DH/ECDH shared secrets
- zero the removed tail bytes and add common bignum conversion tests

## Problem

WinCNG returns DH public values and raw DH/ECDH shared secrets in fixed-width
buffers. The WinCNG bignum retained that buffer width as its numeric length.
When a value started with `0x00`, libssh2 included the redundant byte in the
exchange hash even though RFC 4251 requires a canonical `mpint` representation.

The peer hashes the normalized integer, so the exchange hashes differ and a
valid server host-key signature is rejected. This appeared intermittently as
`LIBSSH2_ERROR_KEY_EXCHANGE_FAILURE` because it depends on the generated value.

The helper introduced here moves significant bytes to the start of the existing
allocation, explicitly clears the removed tail, and updates only the logical
length. It does not add an allocation or a new failure path.

## Validation

- `scripts/checksrc-all.pl`
- MSVC Release build with the default WinCNG configuration
- MSVC Release build with `ENABLE_ECDSA_WINCNG=ON`
- `test_simple` in both configurations
- WSL Clang/OpenSSL build with warnings as errors and project clang-tidy checks;
  `test_simple` passes
- WSL Clang/libgcrypt and Clang/mbedTLS builds with warnings as errors;
  `test_simple` passes on each backend
- earlier Windows connection investigation: 22,000 successful handshakes across default
  group exchange, group14, group16, and independent-process cases
  These connection runs were not repeated for the latest base update.

Reported-by: Dan Fandrich
Fixes #804
Ref: #2590 (tests)


src/wincng.c
```diff
@@ -317,37 +317,33 @@ size_t ssh2_bn_bits(const ssh2_bn *bn)
     return bits;
 }
 
-int ssh2_bn_from_bin(ssh2_bn *bn, const unsigned char *bin, size_t len)
+void ssh2_wcng_bn_normalize(ssh2_bn *bn)
```
vszakats:
```suggestion
static void wcng_bn_normalize(ssh2_bn *bn)
```
Let's keep this static and in the wcng_ namespace because it isn't used outside wincng.

tests/test_simple.c
```diff
@@ -294,6 +294,38 @@ static int test_ssh2_dh_validate(void)
     return err > 0;
 }
 
+#ifdef LIBSSH2_WINCNG
+static int test_ssh2_wcng_bn_normalize(void)
```
vszakats: Could this test a public bn function that relies on the new internal normalization function?
Could it then be enabled and passing for all backends?

vszakats (changes requested):
Can you say how did you stress test connections locally?

Lv2dev:
Thanks for the review. I pushed `34575bad` to address both inline comments:

- The helper is now file-local as `static void wcng_bn_normalize()` and is no longer declared in `wincng.h`.
- The regression now goes through the common bignum API (`ssh2_bn_init_from_bin()`, `ssh2_bn_from_bin()`, `ssh2_bn_bytes()`, and `ssh2_bn_to_bin()`) and is enabled for every backend. It verifies that `{ 0, 0, 0x80, 0x42 }` round-trips with the canonical two-byte magnitude `{ 0x80, 0x42 }`.

Local checks after that change:

- `checksrc-all.pl`: pass
- MSVC Release, WinCNG: build + `test_simple` pass
- MSVC Release, WinCNG with `ENABLE_ECDSA_WINCNG=ON`: build + `test_simple` pass
- WSL GCC Release, OpenSSL: build + `test_simple` pass

For the connection stress testing, I used a local Docker OpenSSH server reached over loopback from a small Windows handshake driver linked to the MSVC WinCNG build. Every attempt created a fresh TCP socket and `LIBSSH2_SESSION`, set the requested KEX preference where applicable, ran `libssh2_session_handshake()`, recorded the result, and then discarded the session. Authentication and file transfer were not part of these measurements.

The unpatched WinCNG control failed intermittently in both serial and one-session-per-process runs, so concurrency was not required. Normalizing only the shared secret still left 41 failures in 10,000 handshakes; instrumentation on those remaining failures showed a redundant leading zero in the exported client public value. With both the public value and shared secret normalized, the runs completed as follows:

- default/GEX: 10,000/10,000
- forced group14: 4,000/4,000
- forced group16: 4,000/4,000
- fresh process per attempt: 4,000/4,000
- total: 22,000/22,000 successful handshakes

The same local server path was also clean with the Windows OpenSSH client (2,000/2,000), and the WSL/OpenSSL control completed its tested KEX variants without failures.

The new upstream CI run is in progress. Please let me know if you would prefer the stress harness to be reduced into a repository test in a follow-up.

vszakats:
Thanks for the fixes! Just curious, are you real person or am I talking
to an LLM?

I think this is the moment when libssh2 would need to have WinCNG
tests in CI (again) to verify if this fixes this long-running problem. I have
an idea for this, which would also hit this specific issue, but so far had
no capacity to implement it.

If you have suggestions for a WnCNG CI test, I'd be useful to see it, yes.
Note, I may have strong opinions on an proposal like this.

Refs:
https://github.com/libssh2/libssh2/issues/804
https://github.com/libssh2/libssh2/pull/1023 (to be reverted if this fix works)
https://github.com/libssh2/libssh2/discussions/2049
https://github.com/curl/curl/pull/16781

@mback2k: Your thoughts on this fix proposal?

Lv2dev:
I'm a real person and the owner of this account. I'm a native Korean speaker and not fluent in English, so I use an LLM to translate my comments into English.

The investigation and patch were AI-assisted. The builds, stress runs, hashes, and CI results I reported are outputs from commands actually executed in my environment, not generated estimates. I make sure I understand the root cause and the purpose of each change, verify the technical claims with reproducible tests, and take responsibility for the submission.

I'd be happy to help restore WinCNG runtime coverage in CI. Since you already have an idea and may have strong preferences about the design, could you outline the topology and constraints you have in mind, such as the runner and SSH server arrangement, minimum KEX coverage, acceptable runtime, and whether retries should remain only as a temporary comparison? I'll review those constraints and the references you shared before preparing a focused follow-up proposal or writing code.

vszakats:
> I'd be happy to help restore WinCNG runtime coverage in CI. Since you already have an idea and may have strong preferences about the design, could you outline the topology and constraints you have in mind, such as the runner and SSH server arrangement, minimum KEX coverage, acceptable runtime, and whether retries should remain only as a temporary comparison? I'll review those constraints and the references you shared before preparing a focused follow-up proposal or writing code.

Short history: We traditionally had WinCNG tests running in AppVeyor CI,
using Docker machines running in GHA and communicating via a tunnel
(with a help of an external server kindly provided by Marc Hörsken). Sadly,
this setup became unstable over time and doesn't work (reliably) anymore,
due to factors outside our control.

Preferably WinCNG tests should reuse the existing test infrastructure
and Docker to make them mirror the tests we do on Linux. It'd be nice to 
avoid the complexities of the previous Windows setup and run Docker on
the same CI machine where the tests are running. Ideally we shall not use
3rd-party GitHub Actions, Powershell code and generally keep complexity
and maintenance overhead at minimum. Probably the best place to run this
would be GHA (as opposed to AppVeyor CI), due to parallelism, and better
expected performance (we lose the option to run tests on old Windows,
and old MSVC versions, which I think is OK.)

[ A long shot, and this isn't the goal within this PR, but I'd much prefer
_not_ using Docker for tests if at all possible. It'd allow running them
locally on any platform, and also allow them to run on macOS (and other
non-Linux platforms) in CI. It'd also allow to dynamically build test keys.
sshd insisting on a valid local user/pw is an issue here I think. ]

As for my idea on running tests: In curl we do run libssh2 tests via curl's
test harness (`runtests`). It runs without Docker and also supports
Windows, and other platforms. It also reproduced this particular WinCNG
issue as flaky fails (for this reason, we had to disable these tests in curl CI).
Therefore my idea is to build curl (e.g. the latest stable, or perhaps the
latest master) in libssh2 CI using the latest libssh2 sources, and use
`runtests` limited to SCP/SFTP tests. These tests can be run against
MSYS2 sshd, the Windows built-in OpenSSH server, or a specific
MS/Windows build installed manually, curl CI has logic for these already.

src/wincng.c
```diff
@@ -317,37 +317,33 @@ size_t ssh2_bn_bits(const ssh2_bn *bn)
...
+static void wcng_bn_normalize(ssh2_bn *bn)
 {
-    unsigned char *bignum;
-    size_t offset, length, bits;
+    size_t offset = 0;
 
+    if(!bn || !bn->bignum || !bn->length)
+        return;
+
+    while(offset + 1 < bn->length && !bn->bignum[offset])
```
mback2k: I think the conditions here should be switched around as the bignum being zero is the primary condition.

mback2k (commented):
This PR itself looks good to me and may finally fix the random issues we had, but of course this will require longer testing as the removed part in `ssh2_bn_from_bin` had the same purpose which is now replaced and generalized.

mback2k (changes requested):
Please implement the requested changes.

mback2k:
I also agree with @vszakats regarding the long-term wishes for a better Windows testsuite, although I would really like to fix the AppVeyor CI, I currently have no time or clue how to do that. WSL could also be an option on GHA to run the Windows tests against a proper Linux-based OpenSSH server as the Windows-version is not completely compatible.

Lv2dev:
Thanks, @vszakats and @mback2k. I've pushed `af506998` with the code and test updates:

- Reordered the normalization loop conditions as requested.
- Normalized both DH public values and shared secrets on the RSA-based fallback paths.
- Expanded the common bignum cases and added native/fallback output tests, including logical lengths and cleared buffer tails. The helper remains file-local, and the common bignum test remains enabled across backends.

The native DH and P-256 ECDH shared-secret tests use fixed vectors. The native DH public-value check compares a fresh key with its CNG export, so detecting omitted normalization there depends on a leading-zero sample. The test itself does not retry or require such a sample. In local mutation checks, six omissions were detected on the first run; the native public-value omission was detected on the 33rd sample.

All 92 existing CI checks for this commit passed. The WinCNG SSH runtime job we discussed has not been added yet.

For that job, I propose building a pinned curl revision against this PR's WinCNG libssh2 on a GHA Windows x64 runner and running curl's SCP/SFTP tests. I'd first check the WSL/Linux OpenSSH option on the same runner against curl's harness and your constraints on setup complexity, third-party Actions, and PowerShell.

Would you prefer this CI work in this PR, or as a follow-up PR?

vszakats:
> For that job, I propose building a pinned curl revision against this PR's WinCNG libssh2 on a GHA Windows x64 runner and running curl's SCP/SFTP tests. I'd first check the WSL/Linux OpenSSH option on the same runner against curl's harness and your constraints on setup complexity, third-party Actions, and PowerShell.

Just to clarify, the curl runtests method does not require or use WSL.

(edit: it _might_ use it, but it seems to be a deep rabbit hole, which
has not been tried before, and with complexities. Since we have zero
tests now, the MSYS2 and the Microsoft port of OpenSSH seems to
be fine enough to know if WinCNG works at all.)

> Would you prefer this CI work in this PR, or as a follow-up PR?

I think it'd be easier to handle in a separate PR, then rebase this one,
once we settled on it.

tests/test_simple.c
```diff
@@ -383,6 +732,14 @@ int main(int argc, char *argv[])
     rc = test_ssh2_base64_decode(session);
     rc |= test_knownhost_ipv6(session);
     rc |= test_ssh2_dh_validate();
+    rc |= test_ssh2_bn_from_bin();
+#ifdef LIBSSH2_WINCNG
+    rc |= test_ssh2_dh_fallback_normalization();
+    rc |= test_ssh2_dh_native_normalization();
+#if LIBSSH2_ECDSA
+    rc |= test_ssh2_ecdh_normalization(session);
+#endif
```
vszakats: Can we avoid the backend-specific (esp. Windows-specific) code in tests?
Either test the primitives for all backends, or drop things from there that are
WinCNG-specific.

vszakats:
> I also agree with @vszakats regarding the long-term wishes for a better Windows testsuite, although I would really like to fix the AppVeyor CI, I currently have no time or clue how to do that. WSL could also be an option on GHA to run the Windows tests against a proper Linux-based OpenSSH server as the Windows-version is not completely compatible.

I think the major issue with AppVeyor CI is that it makes the PR push cycles
take a long time. Even without tests, it used to take 45-60 minutes up until
a few month ago. With cuts and optimizations I could reduce this to 10 minutes
(half of this is waiting to for the machines to spin up), which made iteration
practical again, and ~matches the time a GHA cycle takes. Running tests there
would make it run slow again and working on libssh2 super time-consuming,
and painful. (Most users, myself included, also cannot cancel a job, so one
sometimes need to wait many hours till all the queued up tests complete;
when working on multiple PRs in parallel, and/or pushing to master at the same
time.)

GHA works much better in this regard, by running jobs in parallel, spinning
up machines faster, and there is an option to cancel jobs as a last resort.

To emulate pre-10 Windows versions (only offered by AppVeyor CI) we can
introduce an env to disable `hasAlgDHwithKDF` and test that codepath on
GHA.

vszakats:
Testing the curl runtests idea here: https://github.com/libssh2/libssh2/pull/2590

I suspect this was the WinCNG flaky fail replicated, as spotted in the 2nd run:
```
curl: (2) Failure establishing ssh session: -8, Unable to exchange encryption keys
```
https://github.com/libssh2/libssh2/actions/runs/33995660051/job/101385680240#step:12:2364
More:
https://github.com/libssh2/libssh2/actions/runs/33996376155/job/101387568938#step:12:1836
https://github.com/libssh2/libssh2/actions/runs/33996672153/job/101388338752

vszakats:
With this PR applied as of e95a1eeb2511db996e8035fd9e7c250341b99472,
there seems to be a intermittent hang:
https://github.com/libssh2/libssh2/actions/runs/33996868038/job/101388864311
(There is chance this just the baseline flakiness/hangs seen with runtests on Windows, or a pre-existing WinCNG-backend-specific issue)

Runs fine on the next run:
https://github.com/libssh2/libssh2/actions/runs/33997352596

Lv2dev:
Thanks for testing this in #2590. I've pushed `50e26317` to remove the WinCNG-specific tests and buffer checks from `test_simple.c`, keeping the common bignum tests.

tests/test_simple.c
```diff
@@ -294,6 +294,47 @@ static int test_ssh2_dh_validate(void)
...
+
+    for(i = 0; i < SSH2_ARRAYSIZE(tests); i++) {
+        unsigned char actual[5] = { 0 };
+        size_t length = tests[i].length - tests[i].leading;
+        ssh2_bn *bn = ssh2_bn_init_from_bin();
+
+#ifndef LIBSSH2_LIBGCRYPT
+        if(!bn)
+            return 1;
+#endif
```
vszakats: For non-libgcrypt, shouldn't this fail the conditions below?
If so, perhaps we could drop this backend-specific exception?

tests/test_simple.c
```diff
@@ -294,6 +294,47 @@ static int test_ssh2_dh_validate(void)
...
+        if(ssh2_bn_from_bin(bn, tests[i].input, tests[i].length) ||
+           (size_t)ssh2_bn_bytes(bn) != length ||
```
vszakats: Made this PR to avoid the cast above: #2592

vszakats:
```suggestion
           ssh2_bn_bytes(bn) != length ||
```
cast should no longer be necessary after b766778c220229e57f503dc0ffb9d77d3ffd091a.

tests/test_simple.c
Lv2dev: Thanks for the suggestion. I checked whether the later conditions could cover this and had a question about allocation failures.

[WinCNG's import function rejects NULL](https://github.com/libssh2/libssh2/blob/50e263176be3528df361b34234bfbc83d261c1ac/src/wincng.c#L337-L340). [OpenSSL permits a NULL destination](https://docs.openssl.org/3.5/man3/BN_bn2bin/#description), but [our wrapper](https://github.com/libssh2/libssh2/blob/50e263176be3528df361b34234bfbc83d261c1ac/src/openssl.h#L344-L348) does not assign the returned pointer to `bn`. [mbedTLS requires an initialized destination MPI](https://mbed-tls.readthedocs.io/projects/api/en/v3.6.0/api/file/bignum_8h/). [libgcrypt's initial NULL is intentional](https://github.com/libssh2/libssh2/blob/50e263176be3528df361b34234bfbc83d261c1ac/src/libgcrypt.h#L118-L123), and its scan assigns the new object.

If I'm reading these contracts correctly, keeping the allocation check with a brief comment seems safer. Am I missing anything here? I haven't exercised allocation failures; this is based on the source and API contracts.

Thanks for landing #2592. The `size_t` cast can now be removed after updating this branch's base.

tests/test_simple.c
Lv2dev: I've merged the latest master and removed the `size_t` cast in `fb1261ee`. All 92 CI checks passed. Thanks for aligning the return types.

tests/test_simple.c
vszakats: IMO we shouldn't care about internal differences as a long as the test works,
and as long as the APIs work consistently within a single backend.

(Normally we'd prefer syncing API behavior between backends, but in this
particular case this isn't easily possibel, and the value is only used by the
backend itself, libssh2 code only passes it around. I think that's fine.)

vszakats: Made #2594 to avoid this issue by integrating the special initializer into `ssh2_bn_from_bin()`.

tests/test_simple.c
```diff
@@ -294,6 +294,47 @@ static int test_ssh2_dh_validate(void)
...
+    for(i = 0; i < SSH2_ARRAYSIZE(tests); i++) {
+        unsigned char actual[5] = { 0 };
+        size_t length = tests[i].length - tests[i].leading;
+        ssh2_bn *bn = ssh2_bn_init_from_bin();
+
+#ifndef LIBSSH2_LIBGCRYPT
+        if(!bn)
+            return 1;
+#endif
+        if(ssh2_bn_from_bin(bn, tests[i].input, tests[i].length) ||
```
vszakats:
```suggestion
        ssh2_bn *bn = NULL;
        if(ssh2_bn_from_bin(&bn, tests[i].input, tests[i].length) ||
```
Maybe this, to sync with master, aligning platforms.

src/wincng.c
```diff
@@ -318,37 +318,33 @@ size_t ssh2_bn_bits(const ssh2_bn *bn)
...
+        offset++;
 
+    if(offset > 0) {
+        memmove(bn->bignum, bn->bignum + offset, bn->length - offset);
+        ssh2_explicit_zero(bn->bignum + bn->length - offset, offset);
+        bn->length -= offset;
+    }
+}
+
+int ssh2_bn_from_bin(ssh2_bn *bn, const unsigned char *bin, size_t len)
```
vszakats:
```suggestion
int ssh2_bn_from_bin(ssh2_bn **bn, const unsigned char *bin, size_t len)
```
Needs a rebase on master for this and related changes.

Lv2dev: I've rebased onto the latest master and updated the WinCNG implementation and tests for the new `ssh2_bn_from_bin()` API in `7eaca5dd`. The special initializer and libgcrypt-specific exception are removed.

All 92 CI checks passed. Thanks for simplifying the initialization API.

vszakats (approved):

vszakats:
@mback2k Your change request has been addressed, is that correct?

vszakats:
Thanks @Lv2dev, merged now!

PR #2590, opened 2026-09-05
GHA: test WinCNG SCP/SFTP via curl runtests

vszakats:
The WinCNG backend has been missing runtime tests for a while. To
restore some level of testing, this PR make use of curl's regression
test suite (`runtests`) to run SCP/SFTP-specific tests with the
master/PR libssh2 revision at hand. Some tests need to be disabled when 
using the built-in Windows OpenSSH server.

The current CI jobs reproduce the long-running WinCNG flaky failure
reported in #804 reliability.

Note, if these CI jobs make libssh2 CI a pain to use, they will be
removed. This has a reasonably fair chance to happen, based on the
experience with curl CI with Windows.

Also:
- test both ECDH and non-ECDH builds.
- test build on x64 and ARM64.
- test with both Windows built-in OpenSSH port and MSYS2's sshd.
- enable debug logging in libssh2 while running the tests.
- for extra stress, run each of the 56-57 tests 6 times.
  This may be lowered in a future commit to avoid flakiness and reduce
  CI load and job times.

Caveats: 
- as seen in curl CI, runtests may result in hangs due to broken pipes 
  and error 2304 for undiscovered reasons.
  Discussion: https://github.com/curl/curl/discussions/14854
- MSBuild has an undisablable "smart" feature that sometimes marks a 
  successful run a failure.
  Refs:
  https://github.com/curl/curl/commit/a477789c4344d8231265d739f37787d026086333
  https://github.com/curl/curl/pull/18301
- these tests are different in scope than libssh2's own.
- tests sometime fail flaky with:
  ```
  Failure Event: -43 - Failed getting banner
  ```
  Ref: https://github.com/libssh2/libssh2/actions/runs/34134974899/job/101783681252?pr=2590
- running on modern runner machines, the Win10-introduced DH algo is 
  always enabled.

Ref: https://github.com/libssh2/libssh2/discussions/2049
Ref: #2583

---

TODO:
- [x] perhaps switch to MSYS2 toolchain to avoid the MSBuild issue. (Or keep MSVC and apply the complex workaround used for curl) [MAYBE IN A FUTURE COMMIT]
- [x] perhaps run tests with MS OpenSSH port.
- [x] perhaps run on ARM64.
- [x] finalize workflow name and id.


.github/workflows/ci.yml
```diff
@@ -1312,3 +1312,159 @@ jobs:
...
+            -DCMAKE_GENERATOR_PLATFORM=${archgen} \
+            -DCMAKE_VS_GLOBALS=TrackFileAccess=false \
+            -DCMAKE_UNITY_BUILD=ON -DCURL_DROP_UNUSED=ON -DCURL_WERROR=ON \
+            -DLIBSSH2_INCLUDE_DIR="$HOME"/libssh2-install/include \
+            -DLIBSSH2_LIBRARY="$HOME"/libssh2-install/lib/libssh2.lib
+          cmake --build bld --config Debug --parallel 5
+          /usr/bin/find . \( -name '*.exe' -o -name '*.dll' -o -name '*.lib' -o -name '*.pdb' \) -print0 | grep -z curl | xargs -0 file --
+          /usr/bin/find . \( -name '*.exe' -o -name '*.dll' -o -name '*.lib' -o -name '*.pdb' \) -print0 | grep -z curl | xargs -0 stat -c '%10s bytes: %n' --
+          PATH="$PWD/bld/lib/Debug:$PATH"
+          src/bld/Debug/curl.exe --disable --version
```
vszakats:
```suggestion
          bld/src/Debug/curl.exe --disable --version
```

Lv2dev:
While looking into the test 621 hang, I found busy loops in [SCP error cleanup](https://github.com/libssh2/libssh2/blob/07f8eb0dbb380d2e6aec5ed05c0f0760de5f071d/src/scp.c#L822). Both `scp_recv()` and `scp_send()` keep calling `libssh2_channel_free()` when it returns `LIBSSH2_ERROR_EAGAIN`, even in non-blocking mode. This can keep the call from returning to curl.

I reproduced this with synthetic I/O callbacks, though I haven't confirmed whether it explains the original CI hang.

I have a local patch using the existing `error_closing` state to resume cleanup across calls and preserve the original error. The bounded regression test fails on the original code on Windows and passes with the patch on Windows/WinCNG and Linux/OpenSSL.

Would you like me to open a separate PR for the SCP fix?

vszakats:
> While looking into the test 621 hang, I found busy loops in [SCP error cleanup](https://github.com/libssh2/libssh2/blob/07f8eb0dbb380d2e6aec5ed05c0f0760de5f071d/src/scp.c#L822). Both `scp_recv()` and `scp_send()` keep calling `libssh2_channel_free()` when it returns `LIBSSH2_ERROR_EAGAIN`, even in non-blocking mode. This can keep the call from returning to curl.
> 
> I reproduced this with synthetic I/O callbacks, though I haven't confirmed whether it explains the original CI hang.
> 
> I have a local patch using the existing `error_closing` state to resume cleanup across calls and preserve the original error. The bounded regression test fails on the original code on Windows and passes with the patch on Windows/WinCNG and Linux/OpenSSL.
> 
> Would you like me to open a separate PR for the SCP fix?

Certainly!

vszakats:
@Lv2dev I run it with #2591, and it run fine. But, the hang only happened once,
so this doesn't necessary mean a proof.

PR #2591, opened 2026-09-07
scp: resume non-blocking error cleanup across calls

Lv2dev:
SCP receive and send error cleanup currently keeps retrying `libssh2_channel_free()` when it returns `LIBSSH2_ERROR_EAGAIN`, including in non-blocking mode.

Use the existing `ssh2_NB_state_error_closing` state to return control to the caller and resume cleanup on the next call. Preserve the original error code, message, and ownership flags until cleanup finishes. Release a saved heap-owned message if the session is freed before cleanup completes.

Add a bounded, network-free standalone regression covering receive/send cleanup across calls, restoration of the original error, and saved-message lifetime through completion or session teardown. It uses the common SCP/channel code and does not require a particular crypto backend.

Local validation on upstream `bb8e0957` plus this patch:

- Windows/MSVC/WinCNG: Release and Debug builds with warnings as errors; both standalone tests pass.
- Linux/Clang/OpenSSL under WSL: project clang-tidy checks and a build with warnings as errors; both standalone tests pass.
- Linux/Clang/libgcrypt and mbedTLS: builds with warnings as errors; both standalone tests pass on each backend.
- Linux AddressSanitizer, UndefinedBehaviorSanitizer, and leak detection: both standalone tests pass.
- Upstream `checksrc`: passes, including the new test.

Follow-up to https://github.com/libssh2/libssh2/pull/2590#issuecomment-5565371252. I have not established whether this explains the original test 621 CI hang. The curl runtime suite is provided by the CI workflow from master; it was not run locally.

### DIFF
--- a/tests/test_scp_nonblock.c
+++ b/tests/test_scp_nonblock.c
@@ -243,8 +243,8 @@ static int test_error_cleanup_eagain(int send_case)
                 send_case ? "send" : "recv", (void *)result, rc,
                 session->err_msg ? session->err_msg : "(null)", api_calls,
                 eagain_returns, ctx.recv_calls, ctx.send_calls,
-                (unsigned int)(send_case ? session->scpSend_state :
-                                           session->scpRecv_state),
+                send_case ? (unsigned int)session->scpSend_state :
+                            (unsigned int)session->scpRecv_state,
                 (void *)(send_case ? session->scpSend_channel :
                                     session->scpRecv_channel));
         failed = 1;
--- a/tests/test_scp_nonblock.c
+++ b/tests/test_scp_nonblock.c
@@ -1,0 +1,337 @@
+/* Copyright (C) The libssh2 project and its contributors.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "libssh2_priv.h"
+
+#include <errno.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+struct io_context {
+    LIBSSH2_CHANNEL *channel;
+    void *tracked_message;
+    unsigned int recv_calls;
+    unsigned int send_calls;
+    unsigned int close_after_call;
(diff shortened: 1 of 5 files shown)


Lv2dev:
The remaining [OpenSSL CI failure](https://github.com/libssh2/libssh2/actions/runs/34092474462/job/101648710308) appears to match the [previously reported ML-KEM/Ed25519 handshake flake](https://github.com/libssh2/libssh2/pull/1644#issuecomment-4761871304). The build and the new SCP regression pass; one existing read test fails during hostkey signature verification, before authentication.

Could you rerun that job on `a75804b6`? The other 91 checks passed.

tests/test_scp_nonblock.c
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    rc = test_error_cleanup_eagain(0);
+    rc |= test_error_cleanup_eagain(1);
+    rc |= test_saved_error(0);
+    rc |= test_saved_error(1);
+
+    libssh2_exit();
+    return rc;
+}
```
vszakats, line 272: Can this test be compressed into less amount of code?

vszakats:
> The remaining [OpenSSL CI failure](https://github.com/libssh2/libssh2/actions/runs/34092474462/job/101648710308) appears to match the [previously reported ML-KEM/Ed25519 handshake flake](https://github.com/libssh2/libssh2/pull/1644#issuecomment-4761871304). The build and the new SCP regression pass; one existing read test fails during hostkey signature verification, before authentication.
> 
> Could you rerun that job on `a75804b6`? The other 91 checks passed.

If you rebase on master, the test will run within this PR.

tests/test_scp_nonblock.c
Lv2dev: I've rebased onto the latest master and simplified the test from 337 to 272 lines in `14ac6580`, keeping the four cases and their checks. All 94 CI checks passed. Thanks for the review.

PR #2628, opened 2026-09-14
[TEST] GHA: docker tests

vszakats:
Also:
- Dockerfile: silence progress in `apt` commands. For shorter logs.

---

- to avoid using external Actions (for simple tasks like this).
- [ ] perhaps look into how this'd look with `podman`.
- [ ] test `--push`.
- [ ] always run in PRs, but without `--push`?
  or rather, have it in a test job with no secrets, write right and login.
- [ ] maybe merge Dockerfile instructions.
- [ ] maybe test on Windows. Also to see if the double-copy workaround is still necessary.


vszakats:
Tried buliding a docker image in GHA with Docker to test if Dockerfile chmod workarounds from 2016 are still relevant today. Absolute total disaster to get it moving at all. Not much motivation to repeat it in AppVeyor.

